PR #27, opened 2020-04-07
periodic bc

tangqi:
Add two boundary conditions, for the cases I tested (advection and Poisson), PeriodicBC3 works the best.

Let me know if you catch any bugs, or you want me to make some changes on format. Thanks.

### DIFF
--- a/deepxde/boundary_conditions.py
+++ b/deepxde/boundary_conditions.py
@@ -101,6 +101,48 @@ class PeriodicBC(BC):
         outputs = tf.reshape(outputs, [-1, 2])
         return outputs[:, 0:1] - outputs[:, 1:]
 
+class PeriodicBC2(BC):
+    """Periodic boundary conditions (enforce u' and u) along component_x.
+    """
+    def __init__(self, geom, component_x, on_boundary, component=0):
+        super(PeriodicBC2, self).__init__(geom, on_boundary, component)
+        self.component_x = component_x
+
+    def collocation_points(self, X):
+        X1 = self.filter(X)
+        X2 = np.array([self.geom.periodic_point(x, self.component_x) for x in X1])
+        return np.vstack((X1, X2))
+
+    def error(self, X, inputs, outputs, beg, end):
+        mid=beg+int((end-beg)/2)
+        dydx = tf.gradients(outputs[:, self.component : self.component + 1], inputs)[0]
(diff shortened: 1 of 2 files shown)


lululxvi:
Some suggestions:

1. Remove PeriodicBC2;
2. Merge PeriodicBC3 into the current PeriodicBC, and add a new argument derivative_order to indicate the order of derivative, i.e.,
```
def __init__(self, geom, component_x, on_boundary, component=0, derivative_order=0):
```
3. Python file should be formatted by [black](https://github.com/psf/black)

tangqi:
OK, I am fixing it based on your suggestion. I am testing it in a few cases, but I found an issue.

Here is your implementation:
```python
    def error(self, X, inputs, outputs, beg, end):
        outputs = outputs[beg:end, self.component : self.component + 1]
        outputs = tf.reshape(outputs, [-1, 2])
        return outputs[:, 0:1] - outputs[:, 1:]
```
Here is what I implement (no derivate)
```python
    def error(self, X, inputs, outputs, beg, end):
        mid=beg+int((end-beg)/2)
        yleft = outputs[beg:mid, self.component : self.component + 1]
        yright = outputs[mid:end, self.component : self.component + 1]
        return yleft-yright
```

In 1d they are equivalent. But in 2d, they are not. Maybe I miss something, but I thought in your implementation, the left and right solutions are stack on top of each other through `vstack`. Is the first one a bug? Or did I misunderstand something?

tangqi:
BTW, for the periodic Poisson case, the second implementation gives me the correct answer, but the first one gives a wrong answer.

deepxde/boundary_conditions.py
```diff
@@ -87,19 +87,32 @@ class PeriodicBC(BC):
...
     """
 
-    def __init__(self, geom, component_x, on_boundary, component=0):
+    def __init__(self, geom, component_x, on_boundary, derivative_order=0, component=0):
         super(PeriodicBC, self).__init__(geom, on_boundary, component)
         self.component_x = component_x
+        self.derivative_order = derivative_order
+        if derivative_order > 1:
+            raise NotImplementedError(
+                "PeriodicBC only support derivative_order 0 or 1."
```
lululxvi, line 96: supports

lululxvi:
Nice. Your implementation is correct. See more comments.

tangqi:
Thanks for the suggestion. Fixed.

Issue #93, opened 2020-07-25
How to use the trained NN as a function that can be used in the function pde defined in the main function?


lacrymose:
@lululxvi How to use the trained NN as a function that can be used in the function pde defined in the main function?
 Many thanks in advance!

lululxvi:
Maybe not. But I think your problem can be solved by check the "Q: How can I use a trained model for new predictions?" at FAQ.

lacrymose:
@lululxvi If we want obtain the gradient of the solution, in addition to solution itself, how to accomplish this? Can we directly use tf.gradients(model.predict(x),x) ? If not, what do we do with this case?

lululxvi:
You can use the following to obtain the gradient at `X`:
```python
dy = model.predict(X, operator=lambda x, y: tf.gradients(y, x)[0])
```

lacrymose:
Many thanks! 
Can we use multiple networks to solve a pde system by using DeepXDE? For example, if a system of two pdes is given, is it possible to use two networks in DeepXDE to solve this system?

lacrymose:
these two pdes are tightly coupled. let's say the coefficients of one pde regarding one unkown is a function of the unkown.

lululxvi:
Yes, you can directly choose the output layer of the network to be 2 neurons. See the example ode_system.py

lacrymose:
Can we use two indivisual networks instead of one network with two outputs to achieve this? I find it hard to solve the system using one network with two outputs due to the difficulty of balancing the losses from pdes, bcs and ics. So I want to use two indivisual networks, each with one output, to solve the system. But I have no idea of how to use DeepXDE to get this done.

lululxvi:
Currently DeepXDE only uses one network, and you have to modify the code "maps/fnn.py" to use two networks, which is also easy to do. If you want to do this, we can discuss the details.

Using one network or two networks are essentially the same. You just need to use a large network. Of course, you need to tune the weights of each loss term. (I suspect that even if you use two networks, you still need to tune.) For example, you can check the paper "Systems biology informed deep learning for inferring parameters and hidden dynamics" (https://www.biorxiv.org/content/10.1101/865063v2) for the ODE system, where there are 7 very different outputs.

lacrymose:
I certainly want to learn how to modify the code "maps/fnn.py" to use two network. Tell me how to achieve this, please. Many thanks in advance.
[...]

PR #96, opened 2020-08-01
Feed-forward neural networks that support sub-networks

smao-astro:
The code was modified from @lululxvi 's deepxde/maps/fnn.py, which support users to define independent hidden layers for each unknown variables (output)

The API is the same as deepxde/maps/fnn.py, except that the `layer_size`accept nested list. The general structure of the list define the architecture of the neural network (how the layers are connected). If `layer_size[i]` is int, it represent one layer shared by all the outputs; if `layer_size[i]` is list, it represent `len(layer_size[i])` sub-layers, each of which exclusively used by one output. Note that `len(layer_size[i])` should equal to the number of outputs.  Every number specify the number of neurons of that layer.

example:
```python
        layer_size = [3] + \
                     [config.num_neurons] * config.num_common_hidden_layers + \
                     [[config.num_neurons] * 3] * config.num_independent_hidden_layers + \
                     [3]
        net = dde.maps.PFNN(layer_size, config.activation, config.initializer)
```

Might be useful for #93

### DIFF
--- a/deepxde/maps/__init__.py
+++ b/deepxde/maps/__init__.py
@@ -6,4 +6,5 @@ from .opnn import OpNN
 from .pfnn import PFNN
 from .resnet import ResNet
 
+
 __all__ = ["FNN", "MfNN", "OpNN", "PFNN", "ResNet"]
--- a/deepxde/maps/__init__.py
+++ b/deepxde/maps/__init__.py
@@ -3,7 +3,7 @@ from __future__ import absolute_import
 from .fnn import FNN
 from .mfnn import MfNN
 from .opnn import OpNN
+from .pfnn import PFNN
 from .resnet import ResNet
 
-
-__all__ = ["FNN", "MfNN", "OpNN", "ResNet"]
+__all__ = ["FNN", "MfNN", "OpNN", "PFNN", "ResNet"]
--- a/deepxde/maps/pfnn.py
+++ b/deepxde/maps/pfnn.py
@@ -27,10 +27,10 @@ class PFNN(FNN):
 
         Args:
             layer_size: Accept nested list. The general structure of the list define the architecture of the neural
-                network (how the layers are connected). If layer_size[i] is int, it represent one layer shared by all the
-                outputs; if layer_size[i] is list, it represent len(layer_size[i]) sub-layers, each of which exclusively
-                used by one output. Note that len(layer_size[i]) should equal to the number of outputs. Every number
-                specify the number of neurons of that layer.
+                network (how the layers are connected). If layer_size[i] is int, it represent one layer shared by all
+                the outputs; if layer_size[i] is list, it represent len(layer_size[i]) sub-layers, each of which
+                exclusively used by one output. Note that len(layer_size[i]) should equal to the number of outputs.
+                Every number specify the number of neurons of that layer.
             activation:
             kernel_initializer:
             regularization:
@@ -74,7 +74,8 @@ class PFNN(FNN):
             if type(self.layer_size[i_layer + 1]) is list:
                 if type(y) is list:
                     # e.g. [8, 8, 8] -> [16, 16, 16]
-                    assert len(self.layer_size[i_layer + 1]) == len(self.layer_size[i_layer])
+                    if not (len(self.layer_size[i_layer + 1]) == len(self.layer_size[i_layer])):
+                        raise ValueError("number of sub-layers should be the same when feed-forwarding")
                     y = [
                         layer_map(y[i_net], self.layer_size[i_layer + 1][i_net], self)
                         for i_net in range(len(self.layer_size[i_layer + 1]))
--- a/deepxde/maps/pfnn.py
+++ b/deepxde/maps/pfnn.py
@@ -86,7 +86,11 @@ class PFNN(FNN):
                 y = layer_map(y, self.layer_size[i_layer + 1], self)
         # output layers
         if isinstance(y, (list, tuple)):
-            # e.g. [3, 3, 3] -> [3]
+            # e.g. [3, 3, 3] -> 3
+            if len(self.layer_size[-2]) != self.layer_size[-1]:
+                raise ValueError(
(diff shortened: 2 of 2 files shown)


lululxvi:
Thank you for the code.
- Because the code has lots of common parts with the class `FNN`. Could you use `FNN` as the base class, instead of `Map`? So the only function left is the `build`;
- Just to confirm, the network has to be something like `[3, 100, 100, [100, 100, 100], 3]`, but not `[3, 100, [100, 100, 100], 100, 3]` right? i.e., the parallel part has to be in the end? You can add your first comment in the docstring of `PFNN`.
- Reformat the code using [black](https://github.com/psf/black)

Thanks.

deepxde/maps/pfnn.py
```diff
@@ -0,0 +1,108 @@
...
+                _y = net.dense(_y, layer_size, activation=net.activation)
+            elif net.batch_normalization == "before":
+                _y = net.dense_batchnorm_v1(_y, layer_size)
+            elif net.batch_normalization == "after":
+                _y = net.dense_batchnorm_v2(_y, layer_size)
+            else:
+                raise ValueError("batch_normalization")
+            if net.dropout_rate > 0:
+                _y = tf.layers.dropout(_y, rate=net.dropout_rate, training=net.dropout)
+
```
lululxvi, line 62: Remove this empty line.

deepxde/maps/pfnn.py
```diff
@@ -0,0 +1,108 @@
...
+        print("Building feed-forward neural network...")
+        self.x = tf.placeholder(config.real(tf), [None, self.layer_size[0]])
+
+        y = self.x
+        if self._input_transform is not None:
+            y = self._input_transform(y)
+        # hidden layers
+        for i_layer in range(len(self.layer_size) - 2):
+            if isinstance(self.layer_size[i_layer + 1], list):
```
lululxvi, line 73: Use `isinstance(..., (list, tuple))` to support both list and tuple.

deepxde/maps/pfnn.py
```diff
@@ -0,0 +1,108 @@
...
+                if isinstance(y, list):
```
lululxvi, line 74: Similar here.

deepxde/maps/pfnn.py
```diff
@@ -0,0 +1,108 @@
...
+                    # e.g. [8, 8, 8] -> [16, 16, 16]
+                    if not (
+                        len(self.layer_size[i_layer + 1])
+                        == len(self.layer_size[i_layer])
+                    ):
+                        raise ValueError(
+                            "number of sub-layers should be the same when feed-forwarding"
+                        )
```
lululxvi, line 82: number --> Number

deepxde/maps/pfnn.py
```diff
@@ -0,0 +1,108 @@
...
+                    # e.g. 64 -> [8, 8, 8]
+                    y = [
+                        layer_map(y, self.layer_size[i_layer + 1][i_net], self)
+                        for i_net in range(len(self.layer_size[i_layer + 1]))
+                    ]
+            else:
+                # e.g. 64 -> 64
+                y = layer_map(y, self.layer_size[i_layer + 1], self)
+        # output layers
+        if isinstance(y, list):
```
lululxvi, line 97: Similar here.

deepxde/maps/pfnn.py
```diff
@@ -0,0 +1,100 @@
...
+    Args:
+    layer_size: A nested list to define the architecture of the neural network (how the layers are connected).
+        If `layer_size[i]` is int, it represent one layer shared by all the outputs;
+        if `layer_size[i]` is list, it represent `len(layer_size[i])` sub-layers, each of which exclusively used by one output.
+        Note that `len(layer_size[i])` should equal to the number of outputs.
+        Every number specify the number of neurons of that layer.
+
+    Feed-forward neural networks that support independent "branches" or sub-network inside the network.
+    """
+
```
lululxvi, line 23: - One more indent for line from 15 to 19.
- Move line 21 between line 12 and line 14 with an empty line before and after.

deepxde/maps/__init__.py
```diff
@@ -3,7 +3,7 @@
 from .fnn import FNN
 from .mfnn import MfNN
 from .opnn import OpNN
+from .pfnn import PFNN
 from .resnet import ResNet
 
-
-__all__ = ["FNN", "MfNN", "OpNN", "ResNet"]
+__all__ = ["FNN", "MfNN", "OpNN", "PFNN", "ResNet"]
```
lululxvi, line 9: One extra empty line between `import` and `__all__`

deepxde/maps/pfnn.py
```diff
@@ -0,0 +1,101 @@
...
+    Feed-forward neural networks that support independent "branches" or sub-network inside the network.
+
+    Args:
+        layer_size: A nested list to define the architecture of the neural network (how the layers are connected).
+            If `layer_size[i]` is int, it represent one layer shared by all the outputs;
+            if `layer_size[i]` is list, it represent `len(layer_size[i])` sub-layers, each of which exclusively used by one output.
+            Note that `len(layer_size[i])` should equal to the number of outputs.
+            Every number specify the number of neurons of that layer.
+
```
lululxvi, line 22: Remove this empty line.

deepxde/maps/pfnn.py
```diff
@@ -0,0 +1,101 @@
...
+            y = self._input_transform(y)
+        # hidden layers
+        for i_layer in range(len(self.layer_size) - 2):
+            if isinstance(self.layer_size[i_layer + 1], (list, tuple)):
+                if isinstance(y, (list, tuple)):
+                    # e.g. [8, 8, 8] -> [16, 16, 16]
+                    if not (
+                        len(self.layer_size[i_layer + 1])
+                        == len(self.layer_size[i_layer])
+                    ):
```
lululxvi, line 72: `if len(self.layer_size[i_layer + 1]) != len(self.layer_size[i_layer]):`

PR #170, opened 2020-12-01
deepxde.callbacks.Timer: end training when time used up

smao-astro:


deepxde/callbacks.py
```diff
@@ -211,6 +211,37 @@ def get_monitor_value(self):
...
+    Args:
+        available_time (float): Total time (in minutes) available for the script.
+        buffer_time (float): Time (in minutes) reserved for parts other than training.
+    """
+
+    def __init__(self, available_time, buffer_time=2):
+        super(Timer, self).__init__()
+
+        self.threshold = (available_time - buffer_time) * 60  # convert to seconds
+        self.t_start = time.time()
```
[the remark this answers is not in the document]
smao-astro: Hi Lu, I have modified the code according to your suggestions.

deepxde/callbacks.py
```diff
@@ -211,6 +211,41 @@ def get_monitor_value(self):
         return sum(self.model.train_state.loss_train)
 
 
+class Timer(Callback):
+    """Stop training when training time reach the threshold.
+    This Timer starts after the first call of `on_train_begin`.
+
```
lululxvi, line 217: reach --> reaches

deepxde/callbacks.py
```diff
@@ -211,6 +211,41 @@ def get_monitor_value(self):
...
+        self.threshold = available_time * 60  # convert to seconds
+        self.t_start = None
+        self.time_used = None
+        self.stopped_epoch = None
+
+    def on_train_begin(self):
+        if self.t_start is None:
+            self.t_start = time.time()
+
```
lululxvi, line 233: What about restart the timer in each train, so that the timer measures every single train?

```
def on_train_begin(self):
    self.t_start = time.time()
```

deepxde/callbacks.py
```diff
@@ -211,6 +211,41 @@ def get_monitor_value(self):
...
+            self.model.stop_training = True
+
+    def on_train_end(self):
+        if self.time_used is not None:
+            print(
+                "\nStop training as time used up. time used: {:.1f} mins, epoch trained: {}".format(
+                    self.time_used / 60, self.stopped_epoch
+                )
+            )
+
```
lululxvi, line 247: You can just put the print in the `if` of `on_epoch_end()`, so that the logic is simpler, and you don't need the class variables `self.time_used` and `self.stopped_epoch`

deepxde/callbacks.py
smao-astro: Hi Lu,

I understand your suggestion, while I am wondering what is the most general usage case for this class, originally we implemented it to work with job manager such as slurm (in such case if the time used for the whole script larger than the time requested when submitting the script to the server, the job while be terminated and line after the training part in the script won't be executed), so we originally do want the `Timer` monitor the whole training process and judge whether to end training in advanced based on the **total** training time.

smao-astro: We prefer to keep the timer measures the total training time, so that it can works better with cases such as combining multiple optimizers, etc.

PR #187, opened 2020-12-25
Changing `Pointset` to `PointSetBC` for training using exact solution.

smao-astro:
## Aim
This modification is aim to improve the execution speed of problems that need training data (ground truth, exact solution, numerical solution, `y_train`).

## Background
For the original implementation, take `examples/diffusion_1d_inverse.py` for example, for lines https://github.com/lululxvi/deepxde/blob/3d1edd0529486409fec2c2dadf8775b6684d9bd0/examples/diffusion_1d_inverse.py#L34-L38
`deepxde.boundary_conditions.PointSet` computes the loss by comparing outputs with ground truth, when doing so, since the ground true typically only known on some `x`, the code match each `x` used for training with a list of points where ground truth is known, this maybe slow when number of points is large.

## Modification
The modification improve the code by directly compute loss using ordered (one-to-one match) pairs of outputs and ground truth.

P.S. I am not familiar with `deepxde.data.fpde.FPDE` and how to use git version control with jupyter notebook, so I leave them for the author to update.

### DIFF
--- a/deepxde/boundary_conditions.py
+++ b/deepxde/boundary_conditions.py
@@ -136,23 +136,22 @@ class OperatorBC(BC):
         return self.func(inputs, outputs, X)[beg:end]
 
 
-class PointSet(object):
-    """A set of points.
+class PointSetBC(object):
+    """Boundary Condition for a set of points.
+    Compare the output (that associates with `points`) with `values` (target data).
     """
 
-    def __init__(self, points):
+    def __init__(self, points, values, component):
         self.points = np.array(points)
+        if not isinstance(values, numbers.Number) and values.shape[1] != 1:
+            raise RuntimeError(
+                "PointSetBC should output 1D values. Use argument 'component' for different components."
+            )
+        self.values = np.array(values)
+        self.component = component
 
-    def inside(self, x):
-        return np.any(np.all(np.isclose(x, self.points), axis=1))
+    def collocation_points(self, X):
+        return self.points
 
-    def values_to_func(self, values):
-        zero = np.zeros(len(values[0]))
-
-        def func(x):
-            if not self.inside(x):
-                return zero
-            idx = np.argwhere(np.all(np.isclose(x, self.points), axis=1))[0, 0]
-            return values[idx]
-
-        return lambda X: np.array(list(map(func, X)))
+    def error(self, X, inputs, outputs, beg, end):
+        return outputs[beg:end, self.component : self.component + 1] - self.values
--- a/deepxde/boundary_conditions.py
+++ b/deepxde/boundary_conditions.py
@@ -159,7 +159,7 @@ class PointSet(object):
 
 
 class PointSetBC(object):
-    """Boundary Condition for a set of points.
+    """Dirichlet boundary condition for a set of points.
     Compare the output (that associates with `points`) with `values` (target data).
 
     Args:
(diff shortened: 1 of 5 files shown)


lululxvi:
Could you still keep `PointSet` and add the new `PointSetBC`? Because `PointSet` can be used for other cases.

deepxde/boundary_conditions.py
```diff
@@ -156,3 +156,29 @@ def func(x):
             return values[idx]
 
         return lambda X: np.array(list(map(func, X)))
+
+
+class PointSetBC(object):
+    """Boundary Condition for a set of points.
```
lululxvi, line 162: Dirichlet boundary condition for a set of points.

deepxde/boundary_conditions.py
```diff
@@ -156,3 +156,29 @@ def func(x):
...
+    Compare the output (that associates with `points`) with `values` (target data).
+
+    Args:
+        points: An array of points where the corresponding target values are known and used for training.
+        values: An array of values that gives the exact solution of the problem.
+        component: The output component satisfying this BC.
+    """
+
+    def __init__(self, points, values, component):
```
lululxvi, line 171: component=0

deepxde/boundary_conditions.py
```diff
@@ -156,3 +156,29 @@ def func(x):
...
+        self.points = np.array(points)
+        if not isinstance(values, numbers.Number) and values.shape[1] != 1:
+            raise RuntimeError(
+                "PointSetBC should output 1D values. Use argument 'component' for different components."
+            )
+        self.values = np.array(values)
```
lululxvi, line 177: In the `if` statement, `values.shape[1]` means that `values` must be np array. No need to use `np.array(values)`

examples/Lorenz_inverse.py
```diff
@@ -46,25 +46,16 @@ def boundary(_, on_initial):
...
+    observe_y2 = dde.bc.PointSetBC(observe_t, ob_y[:, 2:3], component=2)
 
     data = dde.data.PDE(
         geom,
         Lorenz_system,
         [ic1, ic2, ic3, observe_y0, observe_y1, observe_y2],
         num_domain=400,
         num_boundary=2,
-        anchors=observe_t,
     )
```
lululxvi, line 59: Previously `observe_t` is also used for PDE loss. Does it affect the accuracy when removing `observe_t`?

examples/Lorenz_inverse.py
```diff
@@ -46,25 +46,16 @@ def boundary(_, on_initial):
...
-    observe_y0 = dde.DirichletBC(
-        geom, ptset.values_to_func(ob_y[:, 0:1]), inside, component=0
-    )
-    observe_y1 = dde.DirichletBC(
-        geom, ptset.values_to_func(ob_y[:, 1:2]), inside, component=1
-    )
-    observe_y2 = dde.DirichletBC(
-        geom, ptset.values_to_func(ob_y[:, 2:3]), inside, component=2
-    )
+    observe_y0 = dde.bc.PointSetBC(observe_t, ob_y[:, 0:1], component=0)
```
lululxvi, line 49: Add `PointSetBC` to `__init__.py`, so that we can use `dde.PointSetBC`

PR #216, opened 2021-02-07
Added elliptic inverse field

jeremyyu8:


examples/elliptic_inverse_field.py
```diff
@@ -0,0 +1,66 @@
...
+from __future__ import print_function
+
+import deepxde as dde
+import matplotlib.pyplot as plt
+from deepxde.backend import tf
+import numpy as np
+
+
+# generate num equally-spaced points from -1 to 1
+def gen_traindata(num):
```
lululxvi, line 12: Use numpy.linspace

examples/elliptic_inverse_field.py
```diff
@@ -0,0 +1,76 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import deepxde as dde
+import matplotlib.pyplot as plt
+from deepxde.backend import tf
+import numpy as np
+
```
lululxvi, line 9: import deepxde as dde
import numpy as np

examples/elliptic_inverse_field.py
```diff
@@ -0,0 +1,76 @@
...
+    # view results
+    x = geom.uniform_points(500)
+    yhat = model.predict(x)
+    uhat, qhat = yhat[:, 0:1], yhat[:, 1:2]
+
+    utrue = np.sin(np.pi * x)
+    print("l2 relative error for u: " + str(dde.metrics.l2_relative_error(utrue, uhat)))
+    plot1 = plt.figure(3)
+    plt.plot(x, uhat, label="uhat")
```
lululxvi, line 61: Just use `plt.figure()`. Similar in line 67

Issue #175, opened 2020-12-07
Implementing SGD


smao-astro:
Hi Lu,

I am interested in how to implement stochastic gradient descent in DeepXDE for training forward PDE problem.

The word SGD here, refer to:
1. At the beginning of one epoch, generate `N_total` number of points over the domain and on BC. Randomly shuffle them. Divide into `m` batches.
2. For every step,  get a new batch of points, evaluate loss, update weight.
3. Iterate for all the batch.

Currently, as far as I know, DeepXDE does not implement SGD, probably because most of cases gradient descent is enough, but for complex problem, SGD could help.

I read part of the code in the repository, and #39 , and find https://github.com/lululxvi/deepxde/issues/39#issuecomment-626407771 have a discussion, seems that modifying `deepxde.data.pde.PDE.train_next_batch` and `deepxde.model.Model._train_sgd` could achieve the goal, and I am wondering that do you interested in implementing it, or does that on your to-do list.

I also would like to discuss a potential issue here: based on my experience and doc for `deepxde.data.pde.PDE.resample_train_points`: https://github.com/lululxvi/deepxde/blob/e7e5276f63b24ba3e49614fa5cb3e4b47daf6285/deepxde/data/pde.py#L155-L157
I am worried that every update of `deepxde.data.pde.PDE.train_x` would require call of `deepxde.model.Model.compile` which could cause the training very slow, do you think this is true? Could you give some suggestion on how to avoid this? Also I am guessing if this is a issue, it might caused by https://github.com/lululxvi/deepxde/blob/e7e5276f63b24ba3e49614fa5cb3e4b47daf6285/deepxde/data/pde.py#L114
Please confirm me about that and give some suggestion, thanks!

lululxvi:
- Yes, we can follow [#39 (comment)](https://github.com/lululxvi/deepxde/issues/39#issuecomment-626407771) to implement SGD by modifying `deepxde.data.pde.PDE.train_next_batch`. No need to modify `deepxde.model.Model._train_sgd`. Also note that you need to keep the points for each BC and PDE the same in each mini-batch, e.g., always 100 BC points and 1000 PDE points, so you need to shuffle and divide each group of the points separately.
- You can also refer the mini-batch implemented in `OpDataSet`:
https://github.com/lululxvi/deepxde/blob/e7e5276f63b24ba3e49614fa5cb3e4b47daf6285/deepxde/data/op_dataset.py#L54
[...]

PR #225, opened 2021-02-18
mini-implementation of resample

smao-astro:
See discussion at #175.


smao-astro:
Notice that this modification could break class that depend on `deepxde.data.pde.PDE` such as `deepxde.data.ide.IDE` and `deepxde.data.fpde.FPDE` and other user defined class if them depends on `pde.train_x_all`, unfortunately we currently do not have time to fix.

lululxvi:
Because it breaks other parts of DeepXDE, could you add a new class instead of modifying `PDE`?

lululxvi:
Should we close this?

smao-astro:
Yes, closed.

PR #229, opened 2021-02-19
Vectorize

smao-astro:
vectorize the following methods:
- on_boundary
- boundary_normal
- periodic_point
- inside

### DIFF
--- a/deepxde/geometry/csg.py
+++ b/deepxde/geometry/csg.py
@@ -29,19 +29,22 @@ class CSGUnion(geometry.Geometry):
         self.geom2 = geom2
 
     def inside(self, x):
-        return self.geom1.inside(x) or self.geom2.inside(x)
+        return np.logical_or(self.geom1.inside(x), self.geom2.inside(x))
 
     def on_boundary(self, x):
-        return (self.geom1.on_boundary(x) and not self.geom2.inside(x)) or (
-            self.geom2.on_boundary(x) and not self.geom1.inside(x)
+        return np.logical_or(
+            np.logical_and(self.geom1.on_boundary(x), ~self.geom2.inside(x)),
+            np.logical_and(self.geom2.on_boundary(x), ~self.geom1.inside(x)),
         )
 
     def boundary_normal(self, x):
-        if self.geom1.on_boundary(x) and not self.geom2.inside(x):
-            return self.geom1.boundary_normal(x)
-        if self.geom2.on_boundary(x) and not self.geom1.inside(x):
-            return self.geom2.boundary_normal(x)
-        return np.zeros(self.dim)
+        return np.logical_and(
+            self.geom1.on_boundary(x), ~self.geom2.inside(x)
+        ) * self.geom1.boundary_normal(x) + np.logical_and(
+            self.geom2.on_boundary(x), ~self.geom1.inside(x)
+        ) * self.geom2.boundary_normal(
+            x
+        )
 
     def random_points(self, n, random="pseudo"):
         x = []
@@ -69,14 +72,19 @@ class CSGUnion(geometry.Geometry):
         return np.random.permutation(x)[:n]
 
     def periodic_point(self, x, component):
-        if self.geom1.on_boundary(x) and not self.geom2.inside(x):
-            y = self.geom1.periodic_point(x, component)
-            if self.on_boundary(y):
-                return y
-        if self.geom2.on_boundary(x) and not self.geom1.inside(x):
-            y = self.geom2.periodic_point(x, component)
-            if self.on_boundary(y):
-                return y
+        x = np.copy(x)
+        on_boundary_geom1 = np.logical_and(
+            self.geom1.on_boundary(x), ~self.geom2.inside(x)
+        )
+        x[on_boundary_geom1] = self.geom1.periodic_point(x, component)[
+            on_boundary_geom1
+        ]
+        on_boundary_geom2 = np.logical_and(
+            self.geom2.on_boundary(x), ~self.geom1.inside(x)
+        )
+        x[on_boundary_geom2] = self.geom2.periodic_point(x, component)[
+            on_boundary_geom2
+        ]
         return x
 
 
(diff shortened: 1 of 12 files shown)


deepxde/geometry/geometry_nd.py
```diff
@@ -91,28 +95,31 @@ def __init__(self, center, radius):
...
         ad = np.dot(xc, dirn)
-        return -ad + (ad ** 2 - np.dot(xc, xc) + self._r2) ** 0.5
+        return -ad + (ad ** 2 - np.sum(xc * xc, axis=-1) + self._r2) ** 0.5
 
     def distance2boundary(self, x, dirn):
         return self.distance2boundary_unitdirn(x, dirn / np.linalg.norm(dirn))
 
     def mindist2boundary(self, x):
-        return np.amin(self.radius - np.linalg.norm(x - self.center, axis=1))
+        return np.amin(self.radius - np.linalg.norm(x - self.center, axis=-1))
```
lululxvi, line 116: This is the same.

lululxvi:
Lots of changes in this pull request:
- bc
    - on_boundary
- geom
    - inside
    - on_boundary
    - on_initial
    - boundary_normal
    - periodic_point
    - distance2boundary
    - distance2boundary_unitdirn
    - is_left
- PointSet
    - inside
    - values_to_func
- Polygon
    - random_points

Most changes are API changes. I think it is better to have a few pull requests step by step to make the modification more clear.

0. If the array is 2d, then `axis=1` and `axis=-1` are the same.
1. Here are 2 changes that does not change the logic: `PointSet.values_to_func` and `Polygon.random_points`. The first pull request can only vectorize these two.
2. Because `bc.on_boundary` is implemented by users, it is better to have backward compatibility. Assume `bc.on_boundary` is not vectorized, then vectorize other codes as more as possible.
3. Still assume `bc.on_boundary` is not vectorized, and then we write a wrapper to make `bc.on_boundary` vectorized in `BC.__init__()`, similar to `np.vectorize`. Then we can vectorize all codes.
4. Support both vectorized or not-vectorized `bc.on_boundary`. If not, then apply the wrapper in Step 3.

smao-astro:
> Because bc.on_boundary is implemented by users, it is better to have backward compatibility. Assume bc.on_boundary is not vectorized, then vectorize other codes as more as possible.
> Still assume bc.on_boundary is not vectorized, and then we write a wrapper to make bc.on_boundary vectorized in BC.__init__(), similar to np.vectorize. Then we can vectorize all codes.

Explored a bit and still not clear how to do this, if we assume the user-defined `bc.on_boundary` apply on single points, and try to do a wrapper to vectorize, my first impression is that the program won't benefit from the acceleration provided by `numpy`'s c API (not an expert on how numpy works, correct me if I am wrong), which is one of the key points of this pull request.

lululxvi:
In BC, change the following
```python
self.on_boundary = on_boundary
```
to
```
self.on_boundary = lambda x, on: np.array([on_boundary(x[i], on[i]) for i in range(len(x))])
```
In this way, `self.on_boundary` is vectorized, and we can vectorize our other codes. Overall, there are benefits.

lululxvi:
Why `on_boundary` and `inside` should be N*1 array? Should we `keepdims`?

smao-astro:
> Why `on_boundary` and `inside` should be N*1 array? Should we just a list or a 1D array?

Hi Lu, that is my implicit convention: all the array should be in shape of (N_sample, N_dimensions), so that all the operation and arrays work together consistently, maybe I should state so explicitly.

lululxvi:
Yes, for all the points, it should be (N_sample, N_dimensions). But `on_boundary` and `inside` are bool arrays, a flatten array seems more intuitive? Because it doesn't have any extra dimenstion. Also, consider the following code
```python
x = np.array([[1], [2], [3]])
x[np.array([True, False, Ture])]  # It is array([[1], [3]])
x[np.array([[True], [False], [Ture]])]  # It is array([1, 3])
```
So `np.array([True, False, Ture])` seems better. What do you think?

lululxvi:
By the way, could you make a separate pull request first to vectorize `Polygon.random_points`?

smao-astro:
Sounds good, implemented.

deepxde/geometry/csg.py
```diff
@@ -29,19 +29,22 @@ def __init__(self, geom1, geom2):
...
-        if self.geom2.on_boundary(x) and not self.geom1.inside(x):
-            return self.geom2.boundary_normal(x)
-        return np.zeros(self.dim)
+        return np.logical_and(
+            self.geom1.on_boundary(x), ~self.geom2.inside(x)
+        ) * self.geom1.boundary_normal(x) + np.logical_and(
+            self.geom2.on_boundary(x), ~self.geom1.inside(x)
+        ) * self.geom2.boundary_normal(
+            x
+        )
```
lululxvi, line 47: Seems a bug. 
```python
        return np.logical_and(
            self.geom1.on_boundary(x), ~self.geom2.inside(x)
        )[:, None] * self.geom1.boundary_normal(x) + np.logical_and(
            self.geom2.on_boundary(x), ~self.geom1.inside(x)
        )[:, None] * self.geom2.boundary_normal(
            x
        )
```

deepxde/geometry/geometry_2d.py
```diff
@@ -382,39 +385,53 @@ def wn_PnPoly(P, V):
...
+                    np.hstack(
+                        [
+                            V[i, 1] > P[:, 1:2],
+                            V[i + 1, 1] <= P[:, 1:2],
+                            is_left(V[i], V[i + 1], P) < 0,
+                        ]
+                    ),
+                    axis=-1,
+                )
+                wn[tmp] -= 1
```
lululxvi, line 414: It is better to add the comments back.

deepxde/geometry/timedomain.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, t0, t1):
         self.t1 = t1
 
     def on_initial(self, t):
-        return np.isclose(t[0], self.t0)
+        return np.isclose(t, self.t0)
```
lululxvi, line 20: Use 1d array.

deepxde/initial_condition.py
```diff
@@ -16,8 +16,7 @@ def __init__(self, geom, func, on_initial, component=0):
         self.component = component
 
     def filter(self, X):
-        X = np.array([x for x in X if self.on_initial(x, self.geom.on_initial(x))])
-        return X if len(X) > 0 else np.empty((0, self.geom.dim))
+        return X[self.on_initial(X, self.geom.on_initial(X)).flatten()]
```
lululxvi, line 19: - Similar to `on_boundary`, in line 15
```python
self.on_initial = lambda ...  # return a 1d array
```
- Then no `flatten` here

lululxvi:
There are still a few bugs, see above. Also, `inside` is vectorized, and then the following should be modified:
- `CSGUnion.random_points()`
- `CSGUnion.random_boundary_points`
- `CSGDifference.random_points()`
- `CSGDifference.random_boundary_points()`
- `CSGIntersection.random_points()`
- `CSGIntersection.random_boundary_points()`

PR #236, opened 2021-02-25
Plot and print features

priyakasimbeg:
- added arguments for passing down use_bias and kernel_constraint from FNN to tf.layers.Dense.
- added __str__ in model to easily get string representation of model info like weights and biases (useful for saving to file).
- addes save options for loss plots and plot from best state

### DIFF
--- a/deepxde/maps/fnn.py
+++ b/deepxde/maps/fnn.py
@@ -25,6 +25,8 @@ class FNN(Map):
         regularization=None,
         dropout_rate=0,
         batch_normalization=None,
+        kernel_constraint=None,
+        use_bias=True,
     ):
         super(FNN, self).__init__()
         self.layer_size = layer_size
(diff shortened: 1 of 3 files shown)


deepxde/maps/fnn.py
```diff
@@ -56,7 +60,7 @@ def build(self):
             y = self._input_transform(y)
         for i in range(len(self.layer_size) - 2):
             if self.batch_normalization is None:
-                y = self.dense(y, self.layer_size[i + 1], activation=self.activation)
+                y = self.dense(y, self.layer_size[i + 1], activation=self.activation, use_bias=self.use_bias)
```
lululxvi, line 63: Should also modify line 72

deepxde/model.py
```diff
@@ -43,6 +43,14 @@ def __init__(self, data, net):
         self.stop_training = False
         self.callbacks = None
 
+    def __str__(self):
```
lululxvi, line 46: This is not very necessary. There is a `state_dict()` function to return a dict.

PR #260, opened 2021-04-09
support resample: output of pde.bc_points() is fixed

smao-astro:
This pull request have the same goal as #225 , what's more, it try to keep backward compatibility, for more detail and background, refer to #175 .

### DIFF
--- a/deepxde/callbacks.py
+++ b/deepxde/callbacks.py
@@ -409,3 +409,29 @@ class MovieDumper(Callback):
                 save_animation(
                     fname_movie, xdata, self.spectrum, logy=True, y_reference=np.abs(A)
                 )
+
+
+class Resampler(Callback):
+    """Resample the training data every given period."""
+
+    def __init__(self, period=1):
+        super(Resampler, self).__init__()
+        self.period = period
+
+        self.num_bcs_initial = None
+        self.epochs_since_last_resample = 0
+
+    def on_train_begin(self):
+        self.num_bcs_initial = self.model.data.num_bcs
+
+    def on_epoch_end(self):
+        self.epochs_since_last_resample += 1
+        if self.epochs_since_last_resample < self.period:
+            return
+        self.epochs_since_last_resample = 0
+        self.model.data.resample_train_points()
+        # consistency check
+        if not np.array_equal(self.num_bcs_initial, self.model.data.num_bcs):
+            print(f"Warning: num_bcs changed!")
+            print("Initial value of self.num_bcs:", self.num_bcs_initial)
+            print("self.model.data.num_bcs:", self.model.data.num_bcs)
--- a/deepxde/callbacks.py
+++ b/deepxde/callbacks.py
@@ -411,11 +411,11 @@ class MovieDumper(Callback):
                 )
 
 
-class Resampler(Callback):
+class PDEResidualResampler(Callback):
     """Resample the training data every given period."""
 
-    def __init__(self, period=1):
-        super(Resampler, self).__init__()
+    def __init__(self, period=100):
+        super(PDEResidualResampler, self).__init__()
         self.period = period
 
         self.num_bcs_initial = None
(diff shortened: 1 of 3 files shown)


deepxde/data/pde.py
```diff
@@ -76,6 +76,7 @@ def __init__(
 
         self.train_x_all = None
         self.train_x, self.train_y = None, None
+        self.train_x_bc = None
```
lululxvi, line 79: Add the `train_x_bc` in the docstring

deepxde/callbacks.py
```diff
@@ -409,3 +409,29 @@ def on_train_end(self):
                 save_animation(
                     fname_movie, xdata, self.spectrum, logy=True, y_reference=np.abs(A)
                 )
+
+
+class Resampler(Callback):
+    """Resample the training data every given period."""
+
+    def __init__(self, period=1):
```
lululxvi, line 417: Is period=1 a good default? Because resampling takes some time, maybe period=100 or 1000 as default?

deepxde/callbacks.py
```diff
@@ -409,3 +409,29 @@ def on_train_end(self):
                 save_animation(
                     fname_movie, xdata, self.spectrum, logy=True, y_reference=np.abs(A)
                 )
+
+
+class Resampler(Callback):
```
lululxvi, line 414: How about `PDEResidualResampler` to be more specific?

deepxde/callbacks.py
```diff
@@ -409,3 +409,29 @@ def on_train_end(self):
...
+
+    def on_epoch_end(self):
+        self.epochs_since_last_resample += 1
+        if self.epochs_since_last_resample < self.period:
+            return
+        self.epochs_since_last_resample = 0
+        self.model.data.resample_train_points()
+        # consistency check
+        if not np.array_equal(self.num_bcs_initial, self.model.data.num_bcs):
+            print(f"Warning: num_bcs changed!")
```
lululxvi, line 435: Do you need `f`?

deepxde/callbacks.py
```diff
@@ -409,3 +409,29 @@ def on_train_end(self):
...
+        self.num_bcs_initial = self.model.data.num_bcs
+
+    def on_epoch_end(self):
+        self.epochs_since_last_resample += 1
+        if self.epochs_since_last_resample < self.period:
+            return
+        self.epochs_since_last_resample = 0
+        self.model.data.resample_train_points()
+        # consistency check
+        if not np.array_equal(self.num_bcs_initial, self.model.data.num_bcs):
```
lululxvi, line 434: if it is not consistent, then we have to re-compile, otherwise the loss is not correct, so we should raise an error here, instead of warining.

smao-astro:
Hi Lu,

another question: how about I change
```python
        self.train_x_all = self.train_points()
        self.train_x = self.bc_points()
        if self.pde is not None:
            self.train_x = np.vstack((self.train_x, self.train_x_all))
        self.train_y = self.soln(self.train_x) if self.soln else None
        if self.auxiliary_var_fn is not None:
            self.train_aux_vars = self.auxiliary_var_fn(self.train_x)
        return self.train_x, self.train_y, self.train_aux_vars
```
to
```python
        self.train_x_all = self.train_points()
        self.bc_points()
        if self.pde is not None:
            self.train_x = np.vstack((self.train_x_bc, self.train_x_all))
        self.train_y = self.soln(self.train_x) if self.soln else None
        if self.auxiliary_var_fn is not None:
            self.train_aux_vars = self.auxiliary_var_fn(self.train_x)
        return self.train_x, self.train_y, self.train_aux_vars
```
guess this is more intuitive?

lululxvi:
The only difference I can see is `self.train_x = self.bc_points()`. But here you cannot remove `self.train_x`, otherwise it is undefined.

smao-astro:
Alright, I see why we should keep it.

PR #268, opened 2021-04-18
Create Burgers 1D

yadhukartha:


lululxvi:
- The filename should be at deepxde/docs/demos/burgers.rst 
- Also need to modify  deepxde/docs/demos/forward.rst

PR #269, opened 2021-04-18
burgers1d

yadhukartha:


lululxvi:
The filename is still not correct. Use the format as the Poisson one.

PR #280, opened 2021-04-26
Add arg to save plots/data to a specified directory

ChrisDeGrendele:


deepxde/postprocessing.py
```diff
@@ -5,7 +5,7 @@
 import matplotlib.pyplot as plt
 import numpy as np
 from mpl_toolkits.mplot3d import Axes3D
-
+import os
```
lululxvi, line 8:
```python
from __future__ import print_function

import os

```

deepxde/postprocessing.py
```diff
@@ -15,14 +15,24 @@ def saveplot(losshistory, train_state, issave=True, isplot=True,
              loss_hist_plot_fname="loss_hist.png",
              save_plot=False,
              show_plot=True,
+             output_dir=os.getcwd()
              ):
+    output_dir += "/"
+    if not os.path.exists(output_dir):
+        print("Warning, output directory doesn't exist. Creating it:")
+        print(output_dir)
```
lululxvi, line 23: print(f"Warning: Directory {output_dir} doesn't exist. Create it.")

deepxde/postprocessing.py
```diff
@@ -15,14 +15,24 @@ def saveplot(losshistory, train_state, issave=True, isplot=True,
...
+        os.mkdir(output_dir)
+
     if issave:
-        save_loss_history(losshistory, loss_fname)
-        save_best_state(train_state, train_fname, test_fname)
+        save_loss_history(losshistory, output_dir+loss_fname)
```
lululxvi, line 27: Use `os.path.join(output_dir, loss_fname)` to support WIndows.

ChrisDeGrendele: Oh good catch, thanks!

lululxvi:
Great. Could you format the code using black https://github.com/psf/black?

PR #306, opened 2021-06-10
Update boundary_conditions.py

AlphaCHO:
Add default return to PointSet function.

### DIFF
--- a/deepxde/boundary_conditions.py
+++ b/deepxde/boundary_conditions.py
@@ -150,12 +150,12 @@ class PointSet(object):
             axis=-1,
         )
 
-    def values_to_func(self, values):
+    def values_to_func(self, values, default_return=0):
         def func(x):
             return np.matmul(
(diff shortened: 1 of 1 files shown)


deepxde/boundary_conditions.py
```diff
@@ -150,12 +150,12 @@ def inside(self, x):
             axis=-1,
         )
 
-    def values_to_func(self, values):
+    def values_to_func(self, values, default_return=0):
```
lululxvi, line 153: use `default_value` instead of `default_return`

deepxde/boundary_conditions.py
```diff
@@ -150,12 +150,12 @@ def inside(self, x):
...
         def func(x):
             return np.matmul(
                 np.all(np.isclose(x[:, np.newaxis, :], self.points), axis=-1),
                 values,
-            )
+            )+default_return*(1-int(np.isclose(x[:, np.newaxis, :], self.points).any()))
```
lululxvi, line 158: format the code using black https://github.com/psf/black

PR #313, opened 2021-06-17
Add algorithm to solve a constraint problem

mBarreau:
The main enhancement is the possibility to use primal-dual algorithms during training (model.py) that results in the following bug changes:
- the weights are now placeholders to be modified during the training
- the back compatibility  is ensured (using loss_constraints=None in the Model.compile implies no modification of the previous code)
- the so-called algorithms are implemented during the sgd training (Model._train_sgd)
- the feed dictionary is updated before each use in the whole class to give values to the weights

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ DeepXDE supports
 - multi-physics, i.e., coupled PDEs;
 - 5 types of boundary conditions (BCs): Dirichlet, Neumann, Robin, periodic, and a general BC; BCs can be defined on an arbitrary domain or on a point set;
 - time-dependent PDEs are solved as easily as time-independent ones by only adding initial conditions;
+- primal-dual algorithms to enforce constraints;
 - residual-based adaptive refinement (RAR);
 - uncertainty quantification using dropout;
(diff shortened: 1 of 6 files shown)


lululxvi:
Thanks for the pull request. Could you write a pseudocode to explain the algorithm?

mBarreau:
Hi,

Thanks. Actually, the submission of the article related to that has been
delayed... I hope this will be done by the end of July.
Should we just wait?

Best,
Matth

lululxvi:
You may write a simplified version of the algorithm to explain the code? Of course, it is up to you whether to release the algorithm or not. I can see that the code mainly modifies the weights and feed_dict. How about you first send a pull request to define the feed_dict:
Replace
```python
            self.sess.run(
                self.train_op,
                feed_dict=self.net.feed_dict(...),
            )
```
to
```python
feed_dict = self.net.feed_dict(...)
self.sess.run(self.train_op, feed_dict=feed_dict)
```
This step is just step 0 without any real modification. Basically, my suggestion is that you can release the code step by step, and release the core part with an algorithm explanation until you feel conformable. How do you think?

PR #328, opened 2021-07-05
Add a example

adsmendesdaniel:
Hello, Lu Lu! 

I'm starting a pull request to add a simple new example: Heat transmission on a uniform 1-dimensional rod (heat equation).

This is a common problem in many textbooks. I've been working with PINN's and similar methods/algorithms for some time and I already have tested some of these common problems. If it is of your interest, I can implement some of then with your library (when I have some time), and contribute with your project.

Thank you.

Daniel Mendes.


lululxvi:
Thank you for the example. Could you add more explanation of the example, something like https://deepxde.readthedocs.io/en/latest/demos/poisson.1d.dirichlet.html? Basically, some math equations to introduction what is the problem setup, and some explanation of your code. You can directly add them to your jyputer notebook.

adsmendesdaniel:
As you suggested, I updated my notebook with a explanation (some equations and text) and commented some lines of the code. 

Basically, I'm just solving a uni-dimensional heat equation with Dirichlet boundary conditions.

In the future, if I have some time, I can add a separate file like your Poisson equation example.

lululxvi:
In the end of the file, "L2 relative error: 1.2298260684876254", so the error is more than 100%?

adsmendesdaniel:
I found the error. A function was defined as _heat_eq_exact_solution(t, x)_, the correct is _heat_eq_exact_solution(x, t)_. This was giving the wrong coordinates to my exact solution points.

PR #339, opened 2021-07-19
Tensorflow supports inverse problems

ZongrenZou:


deepxde/callbacks.py
```diff
@@ -9,6 +9,7 @@
 
 from . import gradients as grad
 from .utils import list_to_str, save_animation
+from .backend import backend_name
```
lululxvi, line 12:
```python
from . import gradients as grad
from .backend import backend_name
from .utils import list_to_str, save_animation
```
`b` is before `u`. ordered by alphabeta

deepxde/callbacks.py
```diff
@@ -255,7 +256,10 @@ class VariableValue(Callback):
 
     def __init__(self, var_list, period=1, filename=None, precision=2):
         super(VariableValue, self).__init__()
-        self.var_list = var_list
+        if isinstance(var_list, list):
+            self.var_list = var_list
+        else:
+            self.var_list = [var_list]
```
lululxvi, line 262: self.var_list = var_list if isinstance(var_list, list) else [var_list]

PR #342, opened 2021-07-20
added comments to pde() and func() to clarify usage

blutjens:
I found the in-/outputs of pde() and func() hard to understand. Here is some comments trying to clarify their usage.


lululxvi:
Great. We have a documentation for this example. Do you want to add your comments into the documentation, instead of the comment of the code?

lululxvi:
I close this because the code has changed.

PR #372, opened 2021-09-10
Update Poisson_Lshape.py

yadhukartha:
L BFGS learning rate missing


lululxvi:
L-BFGS doesn't need learning rate.

PR #416, opened 2021-11-04
Update forward.rst

Handi-Zhang:


lululxvi:
You need to combine the two PR into one.

PR #441, opened 2021-12-04
Update model.py

yadhukartha:
added auxiliary variables


deepxde/model.py
```diff
@@ -532,9 +532,11 @@ def predict(self, x, operator=None, callbacks=None):
             if backend_name == "tensorflow.compat.v1":
```
lululxvi, line 532: Remove the TODO above

PR #445, opened 2021-12-09
Remove redundant code from pde.py

NouamaneTazi:
We don't need to update train points in `add_anchor` since we do it in `train_next_batch`, just like what we did in `__init__`

### DIFF
--- a/deepxde/data/pde.py
+++ b/deepxde/data/pde.py
@@ -193,15 +193,7 @@ class PDE(Data):
             self.anchors = anchors
         else:
             self.anchors = np.vstack((anchors, self.anchors))
-        self.train_x_all = np.vstack((anchors, self.train_x_all))
-        self.train_x = self.bc_points()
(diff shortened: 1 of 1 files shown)


lululxvi:
Good catch. But they are not exactly the same. If we use `train_next_batch`, then the training points `self.train_x_all` will be re-generated and possibly changed. But `add_anchor` shouldn't change it.

NouamaneTazi:
To be honest train_next_batch is still ambiguous to me. For my use case, I trained a model, and just wanted to change the anchors and bcs (PointSets) to refit on new data, but I don't see how it's possible with the current implementation. 
Do I need to create a new model for each training dataset I have if I'm working with a `pde`?

lululxvi:
What part of `train_next_batch` is ambiguous? The correct code doesn't have the function to change the anchors and bc. But it can be implemented easily. You can add new functions to the `PDE`.

PR #456, opened 2021-12-20
Update bionet.py

ShuaiMeng0601:
Add the Cartesian product form in the bionet code


deepxde/nn/tensorflow_compat_v1/bionet.py
```diff
@@ -112,3 +112,110 @@ def _dense(
             kernel_regularizer=regularizer,
             trainable=trainable,
         )
+class BiONetCartesianProd(NN):
```
lululxvi, line 115: subclass of BiONet.
Modify _branch_net and _trunk_net

PR #459, opened 2021-12-21
Update pde.py

AlphaCHO:
Add replace_anchors function at line 206

### DIFF
--- a/deepxde/data/pde.py
+++ b/deepxde/data/pde.py
@@ -206,7 +206,7 @@ class PDE(Data):
     def replace_with_anchors(self, anchors):
         """Replace the current PDE training points with anchors. The BC points will not be changed."""
         self.anchors = anchors.astype(config.real(np))
-        self.train_x_all = anchors
+        self.train_x_all = self.anchors
         self.train_x = self.bc_points()
         if self.pde is not None:
             self.train_x = np.vstack((self.train_x, self.train_x_all))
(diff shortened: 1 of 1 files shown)


deepxde/data/pde.py
```diff
@@ -203,6 +203,19 @@ def add_anchors(self, anchors):
                 config.real(np)
             )
 
+    def replace_anchors(self, anchors):
```
lululxvi, line 206: def replace_with_anchors
    """Replace the current PDE training points with anchors. The BC points will not be changed."""

deepxde/data/pde.py
```diff
@@ -203,6 +203,19 @@ def add_anchors(self, anchors):
...
+        """Replace new points for training PDE losses. The BC points will not be replaced."""
+        self.anchors = anchors
```
lululxvi, line 208: anchors = anchors.astype(config.real(np))

deepxde/data/pde.py
```diff
@@ -203,6 +203,19 @@ def add_anchors(self, anchors):
                 config.real(np)
             )
 
+    def replace_with_anchors(self, anchors):
+        """Replace the current PDE training points with anchors. The BC points will not be changed."""
+        self.anchors = anchors.astype(config.real(np))
+        self.train_x_all = anchors
```
lululxvi, line 209: self.train_x_all = self.anchors

PR #474, opened 2022-01-06
Update optimizers.py

engsbk:
Added the option of using AdaBelief optimizer in TFv1 Backend. Also Added the TF add-ons import line for this purpose.

### DIFF
--- a/deepxde/optimizers/tensorflow_compat_v1/optimizers.py
+++ b/deepxde/optimizers/tensorflow_compat_v1/optimizers.py
@@ -3,6 +3,8 @@ __all__ = ["get", "is_external_optimizer"]
 from .scipy_optimizer import ScipyOptimizerInterface
 from ..config import LBFGS_options
 from ...backend import tf
+import tensorflow_addons as tfa
+
 
 
 def is_external_optimizer(optimizer):
@@ -60,4 +62,5 @@ def _get_optimizer(name, lr):
         "adadelta": tf.train.AdadeltaOptimizer(),
         "rmsprop": tf.train.RMSPropOptimizer(lr),
         "adam": tf.train.AdamOptimizer(lr),
+        "adabelief":tfa.optimizers.AdaBelief(learning_rate = lr),
     }[name]


lululxvi:
Does this code work?

engsbk:
> Does this code work?

Yes, but with a warning:

```
WARNING:tensorflow:5 out of the last 67 calls to <function Model._compile_tensorflow.<locals>.outputs_losses at 0x000001CE9E1B8C10> triggered tf.function retracing. Tracing is expensive and the excessive number of tracings could be due to (1) creating @tf.function repeatedly in a loop, (2) passing tensors with different shapes, (3) passing Python objects instead of tensors. For (1), please define your @tf.function outside of the loop. For (2), @tf.function has experimental_relax_shapes=True option that relaxes argument shapes that can avoid unnecessary retracing. For (3), please refer to https://www.tensorflow.org/guide/function#controlling_retracing and https://www.tensorflow.org/api_docs/python/tf/function for  more details.
```

lululxvi:
From the warning, it seems that it is tensorflow v2 backend, not tensorflow v1

engsbk:
> From the warning, it seems that it is tensorflow v2 backend, not tensorflow v1

I apologize. I forgot to switch the backend.
On tensorflow.compat.v1, I get an error:

```

Traceback (most recent call last):
  File "Wave1D.py", line 96, in <module>
    model.compile("adabelief", lr =1e-4)
  File "/Users/shaikhahalkhadhr/anaconda3/lib/python3.7/site-packages/deepxde/utils/internal.py", line 22, in wrapper
    result = f(*args, **kwargs)
  File "/Users/shaikhahalkhadhr/anaconda3/lib/python3.7/site-packages/deepxde/model.py", line 107, in compile
    self._compile_tensorflow_compat_v1(lr, loss_fn, decay, loss_weights)
  File "/Users/shaikhahalkhadhr/anaconda3/lib/python3.7/site-packages/deepxde/model.py", line 144, in _compile_tensorflow_compat_v1
    total_loss, self.opt_name, learning_rate=lr, decay=decay
  File "/Users/shaikhahalkhadhr/anaconda3/lib/python3.7/site-packages/deepxde/optimizers/tensorflow_compat_v1/optimizers.py", line 38, in get
    train_op = _get_optimizer(optimizer, lr).minimize(loss, global_step=global_step)
TypeError: minimize() got an unexpected keyword argument 'global_step'
```

From what I read the `minimize()` of AdaBelief does not have a `global_step`. 
Is it possible to work around that.

reference: https://www.tensorflow.org/addons/api_docs/python/tfa/optimizers/AdaBelief?hl=es-ar#minimize

lululxvi:
As far as I know, tensorflow-addons only works for tf2. You may double check.

engsbk:
> As far as I know, tensorflow-addons only works for tf2. You may double check.

Yes I think it's a trial to replace various functions that tf.contrib used to offer.
In that case, this optimizer can only be added to TFv2 backend, correct?

lululxvi:
Yes. I think so.

PR #484, opened 2022-01-14
added Helmholtz_Dirichlet_2d example

pescap:
I added a simple Helmholtz 2d case with zero boundary conditions. This example is with soft and hard constraints, for comparison purposes. 

The problem is inspired by [this](https://github.com/FEniCS/dolfinx/blob/main/python/demo/helmholtz2D/demo_helmholtz_2d.py) dolfinx tutorial for the Neumann bcs counterpart.

### DIFF
--- a/docs/demos/helmholtz.2d.dirichlet.rst
+++ b/docs/demos/helmholtz.2d.dirichlet.rst
@@ -72,7 +72,7 @@ Next, we express the PDE residual of the Helmholtz equation:
       return -dy_xx - dy_yy - k0 ** 2 * y - f
 
 
-The first argument to ``pde`` is the network input, i.e., the :math:`x`-coordinate. The second argument is the network output, i.e., the solution :math:`u(x)`, but here we use ``y`` as the name of the variable.
+The first argument to ``pde`` is the network input, i.e., the :math:`x`-coordinate and :math:`y`-coordinate. The second argument is the network output, i.e., the solution :math:`u(x)`, but here we use ``y`` as the name of the variable.
 
 Next, we introduce the exact solution and the Dirichlet boundary condition. 
 If boundary conditions are enforced in a hard fashion, we apply the following transformation to the neural network:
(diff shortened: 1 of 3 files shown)


lululxvi:
in the last line, add
```
dde.saveplot(losshistory, train_state, issave=True, isplot=True)
```
to show the results

lululxvi:
Do you have a link to introduce the PD/BCs of this problem? If so, it would be great to add it in the file as a comment.

Or would you like to write a introduction to your code, like https://deepxde.readthedocs.io/en/latest/demos/forward.html?

lululxvi:
Great. You also need to modify https://github.com/lululxvi/deepxde/blob/master/docs/demos/forward.rst

docs/demos/helmholtz.2d.dirichlet.rst
```diff
@@ -0,0 +1,173 @@
...
+
+  def pde(x, y):
+      dy_xx = dde.grad.hessian(y, x, i=0, j=0)
+      dy_yy = dde.grad.hessian(y, x, i=1, j=1)
+
+      f = k0 ** 2 * sin(k0 * x[:, 0:1]) * sin(k0 * x[:, 1:2])
+      return -dy_xx - dy_yy - k0 ** 2 * y - f
+
+
+The first argument to ``pde`` is the network input, i.e., the :math:`x`-coordinate. The second argument is the network output, i.e., the solution :math:`u(x)`, but here we use ``y`` as the name of the variable.
```
lululxvi, line 75: the :math:`x`-coordinate and :math:`y`-coordinate

pescap: If you prefer, I could use notation :math:`x = (x_1,x_2)` for spatial coordinates to avoid confusion between `y` as coordinate and as the name of the variable.

lululxvi: It is OK. Thanks.

PR #491, opened 2022-01-19
stop training if nan in loss_train or loss_test

pescap:
Hi, it happens sometimes (e.g. when using single precision GPUs) that the loss returns nan values during training.
I added these lines to check for each "display_every" whether there are nans in loss_train or loss_test, in order stop training the model directly.
Are you ok with this procedure? 
Paul

### DIFF
--- a/deepxde/model.py
+++ b/deepxde/model.py
@@ -384,6 +384,9 @@ class Model(object):
             self.train_state.epoch += 1
             self.train_state.step += 1
             if self.train_state.step % display_every == 0 or i + 1 == epochs:
+                if (np.isnan(self.train_state.loss_train).any()) or (np.isnan(self.train_state.loss_test).any()):
+                    self.stop_training = True
+                    break
                 self._test()
 
             self.callbacks.on_batch_end()
--- a/deepxde/model.py
+++ b/deepxde/model.py
@@ -384,12 +384,6 @@ class Model(object):
             self.train_state.epoch += 1
             self.train_state.step += 1
             if self.train_state.step % display_every == 0 or i + 1 == epochs:
-                if (
-                    np.isnan(self.train_state.loss_train).any()
-                    or np.isnan(self.train_state.loss_test).any()
-                ):
-                    self.stop_training = True
-                    break
                 self._test()
 
             self.callbacks.on_batch_end()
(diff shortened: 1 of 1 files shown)


deepxde/model.py
```diff
@@ -384,6 +384,9 @@ def _train_sgd(self, epochs, display_every):
             self.train_state.epoch += 1
             self.train_state.step += 1
             if self.train_state.step % display_every == 0 or i + 1 == epochs:
+                if (np.isnan(self.train_state.loss_train).any()) or (np.isnan(self.train_state.loss_test).any()):
+                    self.stop_training = True
+                    break
```
lululxvi, line 389: No need to call `break` here, because when `stop_training` is True, it will break in line 395

deepxde/model.py
```diff
@@ -384,6 +384,9 @@ def _train_sgd(self, epochs, display_every):
             self.train_state.epoch += 1
             self.train_state.step += 1
             if self.train_state.step % display_every == 0 or i + 1 == epochs:
+                if (np.isnan(self.train_state.loss_train).any()) or (np.isnan(self.train_state.loss_test).any()):
```
lululxvi, line 387: Do we need the outside `()`?

lululxvi:
Great. Also format your code using black https://github.com/psf/black

deepxde/model.py
pescap: Hi Lu, I added this break intentionally  to prevent running `self._test()` again once a nan appears.

Output without break: 
`Step      Train loss    Test loss     Test metric
0         [5.47e+03]    [6.44e+03]    [1.00e+00]
1000      [5.47e+03]    [6.44e+03]    [1.00e+00]
2000      [nan]         [nan]         [nan]
3000      [nan]         [nan]         [nan]

Best model at step 0:
  train loss: 5.47e+03
  test loss: 6.44e+03
  test metric: [1.00e+00]`

Output with break: 

`Step      Train loss    Test loss     Test metric
0         [5.47e+03]    [6.44e+03]    [1.00e+00]
1000      [5.47e+03]    [6.44e+03]    [1.00e+00]
2000      [nan]         [nan]         [nan]

Best model at step 0:
  train loss: 5.47e+03
  test loss: 6.44e+03
  test metric: [1.00e+00]`
  
  Paul

deepxde/model.py
pescap: I will remove them.

deepxde/model.py
```diff
@@ -384,6 +384,12 @@ def _train_sgd(self, epochs, display_every):
             self.train_state.epoch += 1
             self.train_state.step += 1
             if self.train_state.step % display_every == 0 or i + 1 == epochs:
+                if (
+                    np.isnan(self.train_state.loss_train).any()
+                    or np.isnan(self.train_state.loss_test).any()
+                ):
+                    self.stop_training = True
```
lululxvi, line 391: How about putting these lines in `_test()`?

PR #495, opened 2022-01-20
Set Seed

pescap:
Hi,
I Added a set_seed function for reproductibility purposes. 
The function was tested with the three backends, in both CPU and GPU modes.
The seed leads to the same training results!
Paul


lululxvi:
@ZongrenZou You may like this.

deepxde/config.py
```diff
@@ -1,7 +1,9 @@
-from .backend import tf
+import os
+import random
+import numpy as np
+from .backend import tf, torch, backend_name
 from .real import Real
```
lululxvi, line 5: Reorder to:
```
import os
import random

import numpy as np

from .backend import tf, torch, backend_name
from .real import Real
```

deepxde/config.py
```diff
@@ -1,7 +1,9 @@
...
 
-
```
lululxvi, line 4: Add back this empty line.

deepxde/config.py
```diff
@@ -27,3 +29,23 @@ def set_default_float(value):
         print("Set the default float type to float64")
         real.set_float64()
     tf.keras.backend.set_floatx(value)
+
+
+def set_seed(seed):
+    """Set the global random seed.
+
```
lululxvi, line 36: No empty spaces in Line 36.

deepxde/config.py
```diff
@@ -27,3 +29,23 @@ def set_default_float(value):
...
+def set_seed(seed):
+    """Set the global random seed.
+    
+    For reproductibility purposes, one has to set the random, numpy and backend seeds.
+    """
+    random.seed(seed)  #  by python Set random seeds
+    np.random.seed(seed)  #  by numpy Set random seeds
+
+    if backend_name == "pytorch":
+        print("Torch")
```
lululxvi, line 43: It seems there is no need to print here.

deepxde/config.py
```diff
@@ -27,3 +29,23 @@ def set_default_float(value):
...
+        torch.manual_seed(seed)
+    elif backend_name == "tensorflow":
+        tf.random.set_seed(seed)  # tf cpu fix seed
+        os.environ["TF_DETERMINISTIC_OPS"] = "1"
+    elif backend_name == "tensorflow.compat.v1":
+        tf.set_random_seed(seed)  # tf cpu fix seed
+        os.environ["TF_DETERMINISTIC_OPS"] = "1"
+    print("OUT")
```
lululxvi, line 51: Why do we print "OUT"?

deepxde/config.py
lululxvi: There is still empty spaces in line 39

deepxde/config.py
```diff
@@ -27,3 +32,21 @@ def set_default_float(value):
         print("Set the default float type to float64")
         real.set_float64()
     tf.keras.backend.set_floatx(value)
+
+
+def set_seed(seed):
```
lululxvi, line 37: Change function name to `set_random_seed` to be more clear.

PR #499, opened 2022-01-25
Add a function to convert `dat` files to `CSV` format

Saransh-cpp:
I frequently use the data that `DeepXDE` outputs by converting it into `CSV` format, as it makes it compatible with libraries like `csv` and `pandas`. I think others might also find this useful (if they are already doing this with their own code) -
- Added a function to convert `dat` files into `CSV` files.
- Ran `black` across the codebase.
- Added some minor documentation

Usage -
```py
dde.dat_to_csv("../dat_data/heat_1D_loss.dat", "./heat_1D_loss_csv_file.csv", ["x", "t", "u_true", "u_pred"])
```


lululxvi:
Great. Why there are so many formating changes in other files?

Saransh-cpp:
I ran `black` for the entire `DeepXDE` folder. Should I revert them back?

lululxvi:
Yes, that is better, to make this PR more clear.

deepxde/postprocessing.py
```diff
@@ -1,4 +1,5 @@
 import os
+import csv
```
lululxvi, line 2:
```python
import csv
import os
```

deepxde/__init__.py
```diff
@@ -22,7 +22,10 @@
     IC,
 )
 from .model import Model
-from .postprocessing import saveplot
+from .postprocessing import (
```
lululxvi, line 25: It seems to me that `dat_to_csv` is not important enough to be directly under dde namespace

lululxvi:
Last idea: should `dat_to_csv` in `postprocessing` or [`utils`](https://github.com/lululxvi/deepxde/blob/master/deepxde/utils/external.py)?

Saransh-cpp:
> Last idea: should `dat_to_csv` in `postprocessing` or [`utils`](https://github.com/lululxvi/deepxde/blob/master/deepxde/utils/external.py)?

Yes, moving it `utils` sounds better!

PR #507, opened 2022-02-02
first commit for jax backend

ZongrenZou:


deepxde/backend/jax/tensor.py
```diff
@@ -0,0 +1,116 @@
+"""jax backend implementation"""
+import jax
+import jax.numpy as jnp
+import numpy as np
+
```
lululxvi, line 5: Two empty lines between import and lib

deepxde/backend/jax/tensor.py
```diff
@@ -0,0 +1,116 @@
...
+def shape(input_array):
+    return list(input_array.shape)
+
+
+def ndim(input_array):
+    return jnp.ndim(input_array)
+
+
+def Variable(initial_value, dtype=None):
+    ## TODO: add variable class to jax
```
lululxvi, line 37: Only one #

deepxde/backend/jax/tensor.py
```diff
@@ -0,0 +1,116 @@
...
+        "Variable functionality to be implemented for jax backend"
+    )
+
+
+def as_tensor(data, dtype=None):
+    if isinstance(data, jnp.ndarray):
+        if dtype is None or data.dtype == dtype:
+            return data
+        return jnp.asarray(data, dtype=dtype)
+    return jnp.asarray(data, dtype=dtype)
```
lululxvi, line 48: These two lines are exactly the same, and can be merged into one. Does jax has something like `data.type(dtype=dtype)` in torch or `tf.cast(data, dtype)`?

deepxde/backend/jax/tensor.py
```diff
@@ -0,0 +1,116 @@
...
+def square(x):
+    return jnp.square(x)
+
+
+def tanh(x):
+    return jax.nn.tanh(x)
+
+
+def mean(input_tensor, dim, keepdims=False):
+    return jnp.mean(input_tensor, dim, keepdims=keepdims)
```
lululxvi, line 96: return jnp.mean(input_tensor, axis=dim, keepdims=keepdims)

deepxde/backend/jax/tensor.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def reduce_mean(input_tensor):
+    return jnp.mean(input_tensor)
+
+
+def sum(input_tensor, dim, keepdims=False):
+    return jnp.sum(input_tensor, dim, keepdims=keepdims)
```
lululxvi, line 104: return jnp.sum(input_tensor, axis=dim, keepdims=keepdims)

deepxde/backend/jax/tensor.py
ZongrenZou: How about data.astype(dtype)?

lululxvi: Should the first return use astype, or the second return?

PR #514, opened 2022-02-08
Added Explanation for Inverse Poisson Equation with Unknown Forcing Field

bfan05:


lululxvi:
Redo this one, as Burgers case has been merged.

PR #521, opened 2022-02-12
Add PyTorch for LV Equation

eddieqiao23:


examples/Lotka_Volterra.py
```diff
@@ -4,6 +4,8 @@
 from scipy import integrate
 
 from deepxde.backend import tf
```
lululxvi, line 6:
```
# Import tf if using backend tensorflow.compat.v1 or tensorflow
```

lululxvi:
Add comment: # Backend tensorflow.compat.v1 or tensorflow
see https://github.com/lululxvi/deepxde/blob/master/examples/Laplace_disk.py

lululxvi:
You need to add also in other locations

PR #526, opened 2022-02-15
Added Allen-Cahn Equation Code

eddieqiao23:
L2 relative error: 0.015209036238765374


examples/pinn_forward/Allen_Cahn.py
```diff
@@ -0,0 +1,59 @@
+"""Backend supported: tensorflow.compat.v1, tensorflow, pytorch"""
+import numpy as np
+import deepxde as dde
```
lululxvi, line 3:
```
import deepxde as dde
import numpy as np
```

lululxvi:
What is the L2 relative error you obtained?

eddieqiao23:
> What is the L2 relative error you obtained?

L2 relative error: 0.015209036238765374

PR #538, opened 2022-02-23
Float64 for deeponet

minzhu-penn:
Update deeponet.py
Update scipy_optimizer.py

### DIFF
--- a/deepxde/nn/tensorflow_compat_v1/deeponet.py
+++ b/deepxde/nn/tensorflow_compat_v1/deeponet.py
@@ -343,7 +343,7 @@ class DeepONetCartesianProd(NN):
             )
         self.y = tf.einsum("bi,ni->bn", y_func, y_loc)
         # Add bias
-        b = tf.Variable(tf.zeros(1))
+        b = tf.Variable(tf.zeros(1, dtype=config.real(tf)))
         self.y += b
 
         if self._output_transform is not None:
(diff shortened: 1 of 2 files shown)


deepxde/nn/tensorflow_compat_v1/deeponet.py
```diff
@@ -167,7 +167,7 @@ def build(self):
         # Add bias
         if self.use_bias:
             b = tf.Variable(tf.zeros(1))
-            self.y += b
+            self.y += tf.cast(b, tf.float64)
```
lululxvi, line 170:
```
self.y += tf.cast(b, config.real(tf))
```

lululxvi, line 170:
```
b = tf.Variable(tf.zeros(1), dtype=config.real(tf))
self.y += b
```

deepxde/optimizers/tensorflow_compat_v1/scipy_optimizer.py
```diff
@@ -247,7 +247,7 @@ def _pack(cls, tensors):
         elif len(tensors) == 1:
             return tf.reshape(tensors[0], [-1])
         else:
-            flattened = [tf.reshape(tensor, [-1]) for tensor in tensors]
+            flattened = [tf.reshape(tf.cast(tensor, tf.float64), [-1]) for tensor in tensors]
```
lululxvi, line 250: I don't think you need to change this

PR #586, opened 2022-03-29
Added num_trainable_parameters resolves #498

### LINKED ISSUE #498: Print the number of trainable parameters
Hi, 
Do you consider adding a Keras-inspired model.summary() for DeepXDE?
Is there a simple way to access the number of trainable parameters for a model?

Paul

pescap:


lululxvi:
Counting the number of parameters from the size of layers is OK, but might not be a convenience way, e.g., https://github.com/lululxvi/deepxde/blob/master/deepxde/nn/tensorflow_compat_v1/msffn.py

It may be easier to use the code like
```
print("# Parameters:", np.sum([np.prod(v.get_shape().as_list()) for v in tf.trainable_variables()]))
```

- For tf 2, it has `net.trainable_variables` https://www.tensorflow.org/api_docs/python/tf/keras/layers/Layer#attributes
For PyTorch, it has https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.parameters

pescap:
> Counting the number of parameters from the size of layers is OK, but might not be a convenience way, e.g., https://github.com/lululxvi/deepxde/blob/master/deepxde/nn/tensorflow_compat_v1/msffn.py
> 
> It may be easier to use the code like
> 
> ```
> print("# Parameters:", np.sum([np.prod(v.get_shape().as_list()) for v in tf.trainable_variables()]))
> ```
> 
> * For tf 2, it has `net.trainable_variables` https://www.tensorflow.org/api_docs/python/tf/Module#attributes

`net.trainable_variables`gives me `[]`

> * For PyTorch, it has https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.parameters
> 
Worked with:
```pytorch_total_params = sum(p.numel() for p in net.parameters())```

> We can compute this in nn.py, instead of for each NN. Or, we can have a function in utils.py.

**Comment by lululxvi** [comment]
It is strange that `net.trainable_variables` gives you `[]`. It should work, as we have used it here
https://github.com/lululxvi/deepxde/blob/707b0984312d0ad5f28522b2aa863fed923b424a/deepxde/model.py#L190

**Comment by pescap** [comment]
> It is strange that `net.trainable_variables` gives you `[]`. It should work, as we have used it here
> 
> https://github.com/lululxvi/deepxde/blob/707b0984312d0ad5f28522b2aa863fed923b424a/deepxde/model.py#L190

Has it to be called from within a `@tf.function` decorator maybe?

**Comment by lululxvi** [comment]
It should not. Even if it is a tensor, it should return a tensor, not a empty list. But you have to call compile first, else there is no weights yet.

**Comment by pescap** [comment]
I was just running
```model.compile("adam", lr=0.001)```.

It works fine once I call: `model.train(epochs=1000)`

**Comment by pescap** [comment]
Ok, so to sum up formulas:

**tf.compat.v1**

```
np.sum([np.prod(v.get_shape().as_list()) for v in tf.trainable_variables()]))
```

**tensorflow**
```
np.sum([np.prod(v.get_shape().as_list()) for v in net.trainable_variables])
```

**pytorch**

```
np.sum(v.numel() for v in net.parameters())
```

The most simple solution is that I just replace `self.num_trainable_parameters` formula in my first commit in `fnn.py` for each backend. Is it ok for you?

pescap:
Actually, `self.num_trainable_parameters` should be a method, as it will give `[]` if we define it within the `__init__` for `net`.

lululxvi:
Yes, sounds good. But instead in `fnn.py`, you can add it in `nn.py`.

pescap:
Moved to #593

PR #599, opened 2022-04-06
Support for global variable DDE_SEED fixes #585

### LINKED ISSUE #585: Reproductibility for sampler => Define a global DDESEED
Hi, 
I figured out that 

https://github.com/lululxvi/deepxde/blob/707b0984312d0ad5f28522b2aa863fed923b424a/deepxde/geometry/sampler.py#L28

was giving different data points despite having previously defined ``set_random_seed``. I didn't notice this before as I was using `Rectangle` with `uniform` generator.

I obtain the same results if I set a proper seed here, e.g.,

```
seed = 11
set_random_seed(seed)
```

and 

``
np.random.default_rng(seed).
``

So far, I didn't test with other random generators. I also noticed that seeds appeared for the initializers: 

https://github.com/lululxvi/deepxde/blob/707b0984312d0ad5f28522b2aa863fed923b424a/deepxde/nn/initializers.py#L29

Yet, I didn't have issues with that. Initializers seem to be within the ``set_random_seed(seed)`` domain.

To properly set up everything, I propose to set a new global variable ``DDESEED`` with default value `None`. Then, if `DDESEED` is an integer, we would ``set_random_seed(seed)`` and call the seed value in sampler.py.

If you are ok with the idea, I will work on a PR, best!

pescap:
Hi, here go the changes that I suggest to support the DDE_SEED backend.
The lines in `__init__` could be added to a `seed.py` file, but at least the code is working fine.

### DIFF
--- a/deepxde/geometry/sampler.py
+++ b/deepxde/geometry/sampler.py
@@ -4,6 +4,7 @@ import numpy as np
 import skopt
 
 from .. import config
+from .. import seed
 
 
 def sample(n_samples, dimension, sampler="pseudo"):
@@ -25,7 +26,7 @@ def sample(n_samples, dimension, sampler="pseudo"):
 
 def pseudo(n_samples, dimension):
     """Pseudo random."""
-    rng = np.random.default_rng()
+    rng = np.random.default_rng(seed)
     return rng.random(size=(n_samples, dimension), dtype=config.real(np))
 
 
--- a/deepxde/__init__.py
+++ b/deepxde/__init__.py
@@ -11,10 +11,29 @@ __all__ = [
     "Variable",
 ]
 
+import sys
+import os
+
 from .__about__ import __version__
 
 # Should import backend before importing anything else
 from . import backend
+
+thismod = sys.modules[__name__]
+
+# Set up random seed
+from .config import set_random_seed
+
+if "DDE_SEED" in os.environ:
+    seed = os.getenv("DDE_SEED")
+    try:
+        seed = int(seed)
+        set_random_seed(seed)
+    except:
+        seed = None
+    print("Using seed: %s\n" % seed, file=sys.stderr, flush=True)
+
+setattr(thismod, "seed", seed)
 
 from . import callbacks
 from . import data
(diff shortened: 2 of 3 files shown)


pescap:
The code allows to define `DDE_SEED`, e.g.:

`DDE_SEED=10 python`

Then, `dde.seed` is set to be equal to `None` or `DDE_SEED`, allowing to easily get the random seed value.

pescap:
Actually, it would be more judicious to create a `seed` or `random_seed.py` file. I would add the `set_random_seed` function here, as well as the seed set up I added to `__init__.

lululxvi:
There is already a `set_random_seed` to set the seed. Users can set the seed by calling the funciton.

Not sure if we need to support the seed via `DDE_SEED=xx python ...`. Does Python/numpy/TensorFlow/PyTorch support this way?

deepxde/geometry/sampler.py
```diff
@@ -25,7 +26,7 @@ def sample(n_samples, dimension, sampler="pseudo"):
 
 def pseudo(n_samples, dimension):
     """Pseudo random."""
-    rng = np.random.default_rng()
+    rng = np.random.default_rng(seed)
```
pescap, line 29: #618 could be related with this. 
`default_rng` is not whitin the `dde.config.set_random_seed(12345)`.

This is why I worked on this PR, in order to be able to set `seed` and access it afterward, to set `default_rng(seed)` appropriately.

Else, another solution could be to use something equivalent to `default_rng` (another function), which would be encompassed by the `set_random_seed`.

lululxvi:
- I would suggest adding a variable `random_seed` in `config.py`, instead in `__init__.py`.
- This `random_seed` can be set by calling `set_random_seed(seed)`, instead of using `DDE_SEED=10 python`. 

What do you think?

pescap:
> * I would suggest adding a variable `random_seed` in `config.py`, instead in `__init__.py`.
> * This `random_seed` can be set by calling `set_random_seed(seed)`, instead of using `DDE_SEED=10 python`.
> 
> What do you think?

Sounds good. Give me a few days to send a new version.

pescap:
Moved to #641

PR #635, opened 2022-04-23
Redesigns general framework for backend jax: utilizing vmap functionality, and adds auxiliary arguments to data.losses

ZongrenZou:


deepxde/model.py
```diff
@@ -1,6 +1,7 @@
 __all__ = ["Model", "TrainState", "LossHistory"]
 
 import pickle
+import functools
```
lululxvi, line 4:
```python
import functools
import pickle
```

Issue #618, opened 2022-04-14
Reproducability with Seeds


engsbk:
Hello, @lululxvi, Thank you for you efforts!

Is it normal that I may get different prediction results even when using `dde.config.set_random_seed(12345)` in the beginning of my code? 

I'm including it with my imports but I still get different results of prediction. Is this the correct way to use it?
```
################################################################
########################SET SEED################################
################################################################
import deepxde as dde
from deepxde.backend import tf
import numpy as np
from scipy.interpolate import griddata
import matplotlib.pyplot as plt
import matplotlib
print(tf.__version__)
dde.config.set_random_seed(12345)

```

lululxvi:
@pescap Any idea?

lululxvi:
Fixed in the updated code.

pescap:
I suggest that we open this issue again. I am working on a PR to fix the reproducibility issue for `tf.compat.v1` in GPU mode.

Reproducibility (tested with `Poisson_Lshape.py`).

| backend | GPU |  = | CPU |
| :---         |     :---:      |          :---: |  :---: |
| `tf.compat.v1`   |  x     |  x    | ✓
| `tensorflow`   |  ✓     | x     | ✓
| `pytorch`    | ✓       | ✓      |✓

✓: Reproducibility is ok
x: Reproducibility is not ok

Left column: Reproducibility in GPU mode
Center column: Results in GPU mode = Results in CPU mode 
Right column: Reproducibility in CPU modee

lululxvi:
@pescap Poisson L-shape problem could be tricky, due to the L-BFGS.

pescap:
Actually, I removed the L-BFGS lines. I will take the opportunity to test the behavior with the BFGS part.

pescap:
Reproducibility (tested with `Poisson_Lshape.py`).

| backend | GPU |  = | CPU |
| :---         |     :---:      |          :---: |  :---: |
| `tf.compat.v1`   |  ✓     |  x    | ✓
| `tensorflow`   |  ✓     | x     | ✓
| `pytorch`    | ✓       | ✓      |✓

✓: Reproducibility is ok
x: Reproducibility is not ok

jdellag:
> Reproducibility (tested with `Poisson_Lshape.py` over 5000 epochs) in #699.
> backend 	GPU 	= 	CPU
> `tf.compat.v1` 	✓ 	x 	✓
> `tensorflow` 	✓ 	x 	✓
[...]

PR #641, opened 2022-04-25
Seedv2: More simple version to seedv1 fixes #585 and #618

### LINKED ISSUE #585: Reproductibility for sampler => Define a global DDESEED

pescap:
Hi,
I started from the beginning #599 to solve #585 and #618

### DIFF
--- a/deepxde/config.py
+++ b/deepxde/config.py
@@ -1,4 +1,3 @@
-import sys
 import os
 import random
 
--- a/deepxde/config.py
+++ b/deepxde/config.py
@@ -1,3 +1,4 @@
+import sys 
 import os
 import random
 
@@ -6,6 +7,8 @@ import numpy as np
 from .backend import backend_name, tf, torch
 from .real import Real
 
+thismod = sys.modules[__name__]
+setattr(thismod, "random_seed", None)
 
 real = Real(32)
 
@@ -37,8 +40,7 @@ def set_default_float(value):
         real.set_float64()
     if backend_name in ["tensorflow.compat.v1", "tensorflow"]:
         tf.keras.backend.set_floatx(value)
-
-
+    
 def set_random_seed(seed):
     """Set the global random seeds of random, numpy, and backend.
 
@@ -58,3 +60,5 @@ def set_random_seed(seed):
     elif backend_name == "jax":
         global jax_random_seed
         jax_random_seed = seed
+    thismod = sys.modules[__name__]
+    setattr(thismod, "random_seed", seed)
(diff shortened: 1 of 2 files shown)


deepxde/config.py
```diff
@@ -6,6 +7,8 @@
 from .backend import backend_name, tf, torch
 from .real import Real
 
+thismod = sys.modules[__name__]
+setattr(thismod, "random_seed", None)
```
lululxvi, line 11: Directly use
```python
random_seed = None
```

deepxde/config.py
```diff
@@ -58,3 +61,5 @@ def set_random_seed(seed):
     elif backend_name == "jax":
         global jax_random_seed
         jax_random_seed = seed
+    thismod = sys.modules[__name__]
+    setattr(thismod, "random_seed", seed)
```
lululxvi, line 65:
```python
global random_seed
random_seed = seed
```

deepxde/config.py
```diff
@@ -1,3 +1,4 @@
+import sys
```
lululxvi, line 1: No need to import sys

pescap:
Tested successfully with: 
```
"""Backend supported: tensorflow.compat.v1, tensorflow, pytorch"""
import deepxde as dde

dde.config.set_random_seed(10)

def pde(x, y):
    return -dy_xx - dy_yy - 1


def boundary(_, on_boundary):
    return on_boundary


geom = dde.geometry.Polygon([[0, 0], [1, 0], [1, -1], [-1, -1], [-1, 1], [0, 1]])

data = dde.data.PDE(geom, pde, bc, num_domain=1200, num_boundary=120, num_test=1500)
net = dde.nn.FNN([2] + [50] * 4 + [1], "tanh", "Glorot uniform")
model = dde.Model(data, net)

model.train(epochs=2000)
```

PR #656, opened 2022-05-08
Add code for inverse problem of Navier-Stokes equations

zyz-rise:
Hi, Dr. Lu, 
     I wrote the code based on DeepXDE to solve the inverse problem of Navier-Stokes equations. Its flow field is the vortex street behind the flow around the cylinder with Re=100. The training dataset is from https://doi.org/10.1016/j.jcp.2018.10.045.

I have tested this code in three backends of tensorflow.compat.v1, tensorflow and pytorch using CPU and GPU respectively, and the loss of the model can be trained to around **1e-4**. The trained variables (C1, C2) are basically equal to the true values, and the flow field is also consistent with the reference data (PDEs mean residual: 3e-3).
train_loss
variables
flow_field


lululxvi:
Great! Could you
1. Reformat the code using black https://github.com/psf/black?
2. Add the example link at https://github.com/lululxvi/deepxde/blob/master/docs/demos/pinn_inverse.rst

examples/pinn_inverse/Navier_Stokes_inverse.py
```diff
@@ -0,0 +1,182 @@
+"""Backend supported: tensorflow.compat.v1, tensorflow, pytorch
+
+Inverse Problem of Navier Stokes Equation Based on Flow Around Cylinder
+
+training data source：https://doi.org/10.1016/j.jcp.2018.10.045
+"""
+
+from __future__ import absolute_import
```
lululxvi, line 8: Delete the import of `__future__`

examples/pinn_inverse/Navier_Stokes_inverse.py
```diff
@@ -0,0 +1,182 @@
+"""Backend supported: tensorflow.compat.v1, tensorflow, pytorch
+
+Inverse Problem of Navier Stokes Equation Based on Flow Around Cylinder
```
lululxvi, line 3:
```
An inverse problem of the Navier-Stokes equation of flow around cylinder

References: https://doi.org/10.1016/j.jcp.2018.10.045 Section 4.1.1
```

Is it Section 4.1.1?

examples/pinn_inverse/Navier_Stokes_inverse.py
```diff
@@ -0,0 +1,182 @@
...
+
+training data source：https://doi.org/10.1016/j.jcp.2018.10.045
+"""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import deepxde as dde
+# dde.backend.set_default_backend("tensorflow.compat.v1")
```
lululxvi, line 13: This is not the correct way to set the backend. Can you check the first line of the output in the screen for the backed being used?

examples/pinn_inverse/Navier_Stokes_inverse.py
zyz-rise: Yes, Modified the comments at the beginning and the code has been reformatted to black.

examples/pinn_inverse/Navier_Stokes_inverse.py
zyz-rise: Done.

examples/pinn_inverse/Navier_Stokes_inverse.py
zyz-rise: I retested the code in this way: `$python -m deepxde.backend.set_default_backend BACKEND`.
And it works fine in all three backends: `tensorflow.compat.v1`, `tensorflow`, `pytorch`.
Backend_tensorflow compat v1
Backend_tensorflow
Backend_pytorch

PR #667, opened 2022-05-13
PyTorch backend supports PFNN

mattragoza:
- added pytorch implementation of PFNN (parallel networks)
- tested and updated inverse problem examples that use PFNN


lululxvi:
Great! Please reformat the code using [black](https://github.com/psf/black)

deepxde/nn/pytorch/__init__.py
```diff
@@ -1,6 +1,6 @@
 """Package for pytorch NN modules."""
 
-__all__ = ["FNN", "NN"]
+__all__ = ["NN", "FNN", "PFNN"]
```
lululxvi, line 3:
```
__all__ = ["FNN", "NN", "PFNN"]
```

deepxde/nn/pytorch/__init__.py
```diff
@@ -1,6 +1,6 @@
...
 
-from .fnn import FNN
 from .nn import NN
+from .fnn import FNN, PFNN
```
lululxvi, line 6:
```
from .fnn import FNN, PFNN
from .nn import NN
```

deepxde/nn/pytorch/fnn.py
```diff
@@ -35,3 +35,115 @@ def forward(self, inputs):
         if self._output_transform is not None:
             x = self._output_transform(inputs, x)
         return x
+
+
+class PFNN(FNN):
+    """
+    Parallel fully-connected network that uses independent
```
lululxvi, line 42:
```
"""Parallel fully-connected network that uses independent
```

Also, the max length of each line is 88, which is the style of black.

deepxde/nn/pytorch/fnn.py
```diff
@@ -35,3 +35,115 @@ def forward(self, inputs):
...
+            it represents `len(layer_sizes[i])` sub-layers, each
+            of which is exclusively used by one output.
+
+            Note that `len(layer_sizes[i])` should equal the number
+            of outputs. Every number specifies the number of neurons
+            in that layer.
+    """
+
+    def __init__(self, layer_sizes, activation, kernel_initializer):
+        super(FNN, self).__init__()
```
lululxvi, line 60:
```
super().__init__()
```

deepxde/nn/pytorch/fnn.py
```diff
@@ -35,3 +35,115 @@ def forward(self, inputs):
...
+
+        self.activation = activations.get(activation)
```
lululxvi, line 62: Delete.

deepxde/nn/pytorch/fnn.py
```diff
@@ -35,3 +35,115 @@ def forward(self, inputs):
...
+        initializer = initializers.get(kernel_initializer)
+        initializer_zero = initializers.get("zeros")
+
+        assert len(layer_sizes) > 1, "must specify input and output sizes"
```
lululxvi, line 66: This error checking can be moved into `FNN`.

deepxde/nn/pytorch/fnn.py
mattragoza: This is done to avoid using FNN.__init__() function, which expects layer_sizes to be a list of ints and sets up the network in a fully sequential manner. I'm instead calling the NN __init__() function hear, i.e. the parent of the parent of PFNN.

examples/pinn_inverse/diffusion_reaction_rate.py
```diff
@@ -1,11 +1,10 @@
-"""Backend supported: tensorflow.compat.v1
+"""Backend supported: tensorflow.compat.v1 pytorch
```
lululxvi, line 1: add a comma before `pytorch`

examples/pinn_inverse/elliptic_inverse_field.py
```diff
@@ -1,4 +1,4 @@
-"""Backend supported: tensorflow.compat.v1"""
+"""Backend supported: tensorflow.compat.v1 pytorch"""
```
lululxvi, line 1: add a comma

deepxde/nn/pytorch/fnn.py
lululxvi: I see. Then should use
```
class PFNN(NN):
```

mattragoza: I was trying to avoid a rewrite of `FNN.__init__()` since I thought that would add complexity to the pull request.

mattragoza:
The most recent commit introduces an `initialize_layers()` method for pytorch FNN which is overridden by PFNN, instead of overriding the constructor. Therefore, the pytorch implementation still uses the same class hierarchy as the tensorflow v1 version.

deepxde/nn/pytorch/fnn.py
```diff
@@ -39,41 +42,32 @@ def forward(self, inputs):
 
 class PFNN(FNN):
     """
-    Parallel fully-connected network that uses independent
-    sub-networks for each network output.
+    Parallel fully-connected network that uses independent sub-networks for each network output.
```
lululxvi, line 45: Continue in the first line:
```
"""Parallel fully-connected network ...
```

lululxvi:
> The most recent commit introduces an `initialize_layers()` method for pytorch FNN which is overridden by PFNN, instead of overriding the constructor. Therefore, the pytorch implementation still uses the same class hierarchy as the tensorflow v1 version.

As `PFNN` doesn't reuse any methods from `FNN`, there is no need to use `PFNN` as a subclass of `FNN`. Your initial approach is good, and just modify as
```python
class PFNN(NN):
```
There is no need to be consistent as TF 1.x.

mattragoza:
It reuses the constructor!

lululxvi:
Yes, the current code reuses the constructor. But I prefer your original version, which doesn't reuse the constructor, because it only reuses 3 lines of code, but this makes things more complicated. So, I prefer the original version, but change the base class from `FNN` to `NN`.

mattragoza:
I reverted to the previous approach, where PFNN overrides the constructor and inherits from NN instead of FNN.

deepxde/nn/pytorch/fnn.py
```diff
@@ -35,3 +35,110 @@ def forward(self, inputs):
...
+        number specifies the number of neurons in that layer.
+    """
+
+    def __init__(self, layer_sizes, activation, kernel_initializer):
+        super().__init__()
+        self.activation = activations.get(activation)
+        initializer = initializers.get(kernel_initializer)
+        initializer_zero = initializers.get("zeros")
+
+        assert len(layer_sizes) > 1, "must specify input and output sizes"
```
lululxvi, line 62: Do not use `assert`, as the `assert` statements are ignored when running in optimize mode. Use `if` condition.

PR #699, opened 2022-05-25
Seedv3 resolves #618

### LINKED ISSUE #618: Reproducability with Seeds

pescap:


deepxde/config.py
```diff
@@ -57,9 +59,15 @@ def set_random_seed(seed):
     if backend_name == "tensorflow.compat.v1":
         tf.set_random_seed(seed)  # tf CPU seed
         os.environ["TF_DETERMINISTIC_OPS"] = "1"
+        os.environ["TF_CUDNN_DETERMINISTIC"] = "1"
+        os.environ["PYTHONHASHSEED"] = str(seed)
+        tf.config.threading.set_inter_op_parallelism_threads(1)
+        tf.config.threading.set_intra_op_parallelism_threads(1)
     elif backend_name == "tensorflow":
         tf.random.set_seed(seed)  # tf CPU seed
```
lululxvi, line 67: Do you need
```python
tf.keras.utils.set_random_seed(1)
```
as shown in https://www.tensorflow.org/api_docs/python/tf/config/experimental/enable_op_determinism

deepxde/config.py
```diff
@@ -49,6 +49,8 @@ def set_default_float(value):
 def set_random_seed(seed):
     """Set the global random seeds of random, numpy, and backend.
 
+    - For backend TensorFlow 2.x: Results might change if you run the model several times in the same terminal.
+
```
lululxvi, line 53: Add a warning:
```
Warning:
    Note that determinism in general comes at the expense of lower performance and so your model may run slower when determinism is enabled.
```

deepxde/config.py
```diff
@@ -49,6 +49,8 @@ def set_default_float(value):
 def set_random_seed(seed):
     """Set the global random seeds of random, numpy, and backend.
```
lululxvi, line 50: Set the global random seeds of random, numpy, and backend, so that your code runs deterministically.

deepxde/config.py
pescap: As stated in https://www.tensorflow.org/api_docs/python/tf/keras/utils/set_random_seed

"Calling this utility is equivalent to the following:"

```
import random
import numpy as np
import tensorflow as tf
random.seed(seed)
np.random.seed(seed)
tf.random.set_seed(seed)
```

I prefer not using `tf.keras.utils.set_random_seed(1)` as it was added only for tf>=1.17.

deepxde/config.py
```diff
@@ -47,7 +47,11 @@ def set_default_float(value):
 
 
 def set_random_seed(seed):
-    """Set the global random seeds of random, numpy, and backend.
+    """Set the global random seeds of random, numpy, and backend, so that your code runs deterministically.
+
+    - For backend TensorFlow 2.x: Results might change if you run the model several times in the same terminal.
+    Warning:
```
lululxvi, line 53: Format is not correct:
- Add an empty line before Warning
- Add 4 space indent before Note

deepxde/config.py
```diff
@@ -47,7 +47,12 @@ def set_default_float(value):
 
 
 def set_random_seed(seed):
-    """Set the global random seeds of random, numpy, and backend.
+    """Set the global random seeds of random, numpy, and backend, so that your code runs deterministically.
```
lululxvi, line 50: Sets all random seeds for the program (Python random, NumPy, and backend).

You can use this to make the program fully deterministic. This means that if the program is run multiple times with the same inputs on the same hardware, it will have the exact same outputs each time. This is useful for debugging models.

pescap: Sets all random seeds for the program (Python random, NumPy, and backend).

You can use this to make the program fully deterministic. This means that if the program is run multiple times with the same inputs on the same hardware, it will have the exact same outputs each time. This is useful for debugging models, and to obtain fully reproducible results for a given setting.

pescap: Sets all random seeds for the program (Python random, NumPy, and backend).

You can use this to make the program fully deterministic. This means that if the program is run multiple times with the same inputs on the same hardware, it will have the exact same outputs each time. This is useful for debugging models, and for obtaining fully reproducible results.

deepxde/config.py
```diff
@@ -57,9 +64,15 @@ def set_random_seed(seed):
     if backend_name == "tensorflow.compat.v1":
         tf.set_random_seed(seed)  # tf CPU seed
         os.environ["TF_DETERMINISTIC_OPS"] = "1"
+        os.environ["TF_CUDNN_DETERMINISTIC"] = "1"
+        os.environ["PYTHONHASHSEED"] = str(seed)
+        tf.config.threading.set_inter_op_parallelism_threads(1)
```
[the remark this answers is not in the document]
pescap: From what I observed, the code did not seem to go much slower.

lululxvi:
> line too long: I don't know if you have a specific rule to "go to line" (see my last commit)

88 characters per line

deepxde/config.py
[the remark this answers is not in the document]
lululxvi: in both CPU and GPU?

deepxde/config.py
```diff
@@ -47,7 +47,18 @@ def set_default_float(value):
 
 
 def set_random_seed(seed):
-    """Set the global random seeds of random, numpy, and backend.
+    """Sets all random seeds for the program (Python random, NumPy, and backend).
+
+    You can use this to make the program fully deterministic. This means that if
+        the program is run multiple times with the same inputs on the same hardware,
```
lululxvi, line 53: Don't need the indent from line 53 to line 55.

deepxde/config.py
```diff
@@ -57,9 +68,15 @@ def set_random_seed(seed):
     if backend_name == "tensorflow.compat.v1":
         tf.set_random_seed(seed)  # tf CPU seed
         os.environ["TF_DETERMINISTIC_OPS"] = "1"
+        os.environ["TF_CUDNN_DETERMINISTIC"] = "1"
```
lululxvi, line 71: Based on https://docs.nvidia.com/deeplearning/frameworks/tensorflow-release-notes/rel_19.06.html, If you set TF_DETERMINISTIC_OPS=1 then there is no need to also set TF_CUDNN_DETERMINISM=1.

deepxde/config.py
```diff
@@ -47,7 +47,18 @@ def set_default_float(value):
 
 
 def set_random_seed(seed):
-    """Set the global random seeds of random, numpy, and backend.
+    """Sets all random seeds for the program (Python random, NumPy, and backend).
```
lululxvi, line 50: Sets all random seeds for the program (Python random, NumPy, and backend), and configures the program to run deterministically.

deepxde/config.py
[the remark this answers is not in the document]
pescap: Let me run a few numerical experiments.

deepxde/config.py
lululxvi: OK. Then you can add it back.

deepxde/config.py
```diff
@@ -47,19 +47,32 @@ def set_default_float(value):
...
     if backend_name == "tensorflow.compat.v1":
-        tf.set_random_seed(seed)  # tf CPU seed
         os.environ["TF_DETERMINISTIC_OPS"] = "1"
+        os.environ["TF_CUDNN_DETERMINISTIC"] = "1"
+        os.environ["PYTHONHASHSEED"] = str(seed)
+        tf.set_random_seed(seed)  # tf CPU seed
     elif backend_name == "tensorflow":
-        tf.random.set_seed(seed)  # tf CPU seed
         os.environ["TF_DETERMINISTIC_OPS"] = "1"
+        os.environ["PYTHONHASHSEED"] = str(seed)
```
lululxvi, line 74: Do you also need to add `TF_CUDNN_DETERMINISTIC` here?

pescap: No

deepxde/config.py
```diff
@@ -47,19 +47,32 @@ def set_default_float(value):
...
 
     Args:
         seed (int): The desired seed.
     """
     random.seed(seed)  # python random
     np.random.seed(seed)  # numpy
     if backend_name == "tensorflow.compat.v1":
-        tf.set_random_seed(seed)  # tf CPU seed
         os.environ["TF_DETERMINISTIC_OPS"] = "1"
+        os.environ["TF_CUDNN_DETERMINISTIC"] = "1"
```
lululxvi, line 69: Add a comment here:
```
Based on https://docs.nvidia.com/deeplearning/frameworks/tensorflow-release-notes/rel_19.06.html, if we set TF_DETERMINISTIC_OPS=1 then there is no need to also set TF_CUDNN_DETERMINISM=1. However, our experiment shows that TF_CUDNN_DETERMINISM=1 is still required.
```

deepxde/config.py
```diff
@@ -47,19 +47,32 @@ def set_default_float(value):
...
+        #os.environ["PYTHONHASHSEED"] = str(seed)
+        tf.set_random_seed(seed)  # tf CPU seed
     elif backend_name == "tensorflow":
-        tf.random.set_seed(seed)  # tf CPU seed
         os.environ["TF_DETERMINISTIC_OPS"] = "1"
+        os.environ["PYTHONHASHSEED"] = str(seed)
```
lululxvi, line 74: Also `os.environ["PYTHONHASHSEED"] = str(seed)` here

deepxde/config.py
```diff
@@ -47,19 +47,32 @@ def set_default_float(value):
...
     Args:
         seed (int): The desired seed.
     """
     random.seed(seed)  # python random
     np.random.seed(seed)  # numpy
     if backend_name == "tensorflow.compat.v1":
-        tf.set_random_seed(seed)  # tf CPU seed
         os.environ["TF_DETERMINISTIC_OPS"] = "1"
+        os.environ["TF_CUDNN_DETERMINISTIC"] = "1"
+        os.environ["PYTHONHASHSEED"] = str(seed)
```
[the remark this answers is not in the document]
lululxvi: OK. Then delete them.

deepxde/config.py
```diff
@@ -47,19 +47,33 @@ def set_default_float(value):
...
 
 def set_random_seed(seed):
-    """Set the global random seeds of random, numpy, and backend.
+    """Sets all random seeds for the program (Python random, NumPy, and backend), and configures the program to run deterministically.
+
+    You can use this to make the program fully deterministic. This means that if the program is run multiple times with the same
+    inputs on the same hardware, it will have the exact same outputs each time. This is useful for debugging models, and for
+    obtaining fully reproducible results.
+    
+    - For backend TensorFlow 1.x: Based on https://docs.nvidia.com/deeplearning/frameworks/tensorflow-release-notes/rel_19.06.html
```
lululxvi, line 56: This comment should go to line 71, as it is a technical comment, not useful for users.

deepxde/config.py
```diff
@@ -47,19 +47,33 @@ def set_default_float(value):
...
+
+    You can use this to make the program fully deterministic. This means that if the program is run multiple times with the same
+    inputs on the same hardware, it will have the exact same outputs each time. This is useful for debugging models, and for
+    obtaining fully reproducible results.
+
+    - For backend TensorFlow 2.x: Results might change if you run the model several times in the same terminal.
+
+    Warning:
+        Note that determinism in general comes at the expense of lower performance and so your model
+            may run slower when determinism is enabled.
```
lululxvi, line 60: No indent is needed here.

Issue #320, opened 2021-06-24
SGD/small-batch


Ryszard2:
Hello @lululxvi and @smao-astro 

I can’t figure out if SGD or small-batch is definitively implemented and available in DeepXDE, without any adjustment to the source code by the user.

I’m trying to train a `[3] + [128]*4 + [3]` NN in a 2D Riemann problem where I would like to use some tens of thousands of training points. Doing it with the full gradient descent would take like one hour for a single 1000 epochs step (the max power available to me is the Google Colab GPU, whatever GPU it is).

#

I tried a small-batch-like approach by myself, without touching the source code, also because I’m not sure of what to do.
It’s more or less like this:

1. I train a small batch of domain/BC/IC, like 200/100/100 for like 10000 epochs, saving the model, `save_better_only`.

2. I start over the training, restoring the model, for a brand-new set of 200/100/100 points for another 10000 epochs.

And so on...

One thing I do not like of this approach is that I would like to save really an improved model, but every time I start over the training, after the first new 1000 epochs I get the message of the train loss improved from `inf` to the new loss, say:

`Epoch 1000: train loss improved from inf to 8.91e-01, saving model to ...`

where `8.91e-01` is way higher than the loss I got to at the end of the previous 10000 epochs run. I feel like I’m updating bad the hyperparameters every time I start over with the training.

#

Another approach I’m testing, inspired from #305, is:

1. I train a medium-size set of IC points, say 5000 points for 10000 epochs, I set to zero the weights and the points for PDE and BC. I save the model, save_better_only.
2. I start over the training, restoring the model, for a brand-new set of 0/0/5000 points for another 10000 epochs.

And so on...

3. When I’m satisfied with the results with the IC, I take on the BC part, keeping always some points for the IC, say 0/5000/100
4. When I’m satisfied with the results with the BC, I take on the PDE part, keeping always some points for BC and IC, say 5000/100/100

That’s what I’m thinking about.
I’m not sure it is a sound way to train a NN.

I’m still in the middle of it, it still _“improves”_ the train loss from `inf` to the new loss every time it starts over, but it seems to work better than the first approach.

#

Thank you
Riccardo

lululxvi:
Yes, SGD/mini-batch is available in DeepXDE. You can use `dde.callbacks.PDEResidualResampler()`, see an example https://github.com/lululxvi/deepxde/blob/master/examples/diffusion_1d_resample.py

Ryszard2:
[...]

Issue #579, opened 2022-03-24
Variable boundary conditions


chang-change:
Dear lulu,DeepXDE is a great tool.However, I have some questions. I am solving a 2-dimensional heat equation,  and I want to use spatial coordinates and boundary conditions as input to the neural network,  and the output is the temperature field of the solution domain. My goal is to train with different boundary conditions  and then input a new boundary condition to the trained neural network so that the trained neural network can directly  input a temperature field. Can DeepXDE do this?  Thank you very much!

praksharma:
Hi, I also use DeepXDE. I guess you can do this with parametrization. You need to heavily tweak the source code.
But if you think this from a physics perspective. well-posed 2D heat equation has a unique solution. You are asking a neural network to throw multiple unique solutions for each well-posed problem for a different BC vector. This would require a huge amount of data and I would suggest you make your problem underdetermined i.e. put parameters more than the number of samples.

From my experience of solving heat equations, especially steady-state. Be very careful with your BCs. If you have two adjacent boundaries with conflicting BCs, then you are literally introducing a discontinuity. Well, Nvidia Modulus uses signed distance function to kick those points out of the training process.

For example, I solved steady state 3D heat conduction problem with different temperature at each face. It took me 120 layers and 150 neuron in each layer. Basically, an underdetermined system (assume we remove the PDE loss) for my training set of 20k points.

This needs an HPC cluster for sure.

lululxvi:
More comments:

- There are several discussions on parametric PDE. Check FAQ "Q: Solve parametric PDEs.", but I haven't listed all the related issues, and you can also search in the issues.
- Depending on exactly what you try to do, you may need or not to tweak the code.
- For the conflicting adjacent BCs, the current version of DeepXDE should remove all the points of those corners. @praksharma Do you have a specific example (2D/3D), where DeepXDE didn't remove the corner points?

praksharma:
@lululxvi Thanks for the instant reply. I was trying to see whether PINNs can solve 2D and 3D problems with conflicting adjacent BCs. Of course, those baseline PINNs, learning rate annealing with encoder, Neural Tangent kernel were designed for this.

Surprisingly, 2D heat conduction with conflicting adjacent BCs did work with DeepXDE. Here are the [results](https://github.com/lululxvi/deepxde/issues/564#issue-1165412222) in one of the closed issues on this repo.

When I tried the 3D heat conduction, things became very weird with DeepXDE. I applied different temperatures on each face.

[...]

Issue #39, opened 2020-05-10
Code discussion


kpratik41:
Thank you very much for your amazing work. Had a few questions.

1) I was able to successfully run it on single GPU but it gave me an OOM error when i increased the domain points to a large value. How do I make it run on a multiple GPU (using TF 1.14). Could you provide me some direction regarding this? Have you ever ran this code on multiple GPU's?

2) After going through the code I believe that currently you have not implemented mini batch. What sort of effort do you think would be required to implement that?

3) Dropouts are used in neural networks in order to prevent overiftting. If we are solving a heat equation then we want the network to learn all the small variations in space and time over the entire domain. Essentially we want to overfit the neural network on that domain to get the most accurate solution so should we use dropouts during training?

lululxvi:
1. I haven't tried multiple GPUs, but it should be not hard. TF supports multiple GPUs, and you can also use [Horovod[(https://github.com/horovod/horovod). This may require a small modification of the code.
2. Which problem you are solving? For most problems, I believe you don't need so many points. Compared to finite difference/element, we need much less points to achieve the same accuracy. You can also use `residual-based adaptive refinement` method to reduce the number of points, see the paper (https://arxiv.org/abs/1907.04502).
3. I don't use mini batch, due to the previous reason. But it is very easy to use mini batch. The only thing to modify is the function `train_next_batch(self, batch_size=None)` in the class class `PDE`. As you can see, I actually have the argument `batch_size`. During training, we call `train_next_batch` for each SGD iteration. Currently, I just ignore `batch_size`, and return the whole dataset.
4. You are right. We don't have the overiftting issue, so we don't need dropout.

kpratik41:
I was just trying to implement the NSFnets paper(March 2020) as part of my learning process. They have used very large number of points in the domain (144,000). The GPU that I am using has 12 GB memory and I have 4 such GPU's. If I use just one then I am get OOM error for  more than 40,000 points.

The results I am getting for less points are not satisfactory so I wanted to use more number of points. Also currently I am using using anchors functionality in your code to specify points rather than specifying num_domains and num_boundary and num_initial because I felt that uniform number of points in time and space might lead to lower loss values.

`data = dde.data.TimePDE(
    geomtime, 3, pde, [ic1, ic2, bc1, bc2,
                       bc4, bc5,  bc8, bc9, bc10],  anchors=points)`

[...]

PR #703, opened 2022-05-25
Add support for Horovod (tensorflow)

pescap:
Work in progress. Add the hvd support, to use multiple GPUs. Sounds to be working properly (tested with 1 up to 8 GPUs) and is non-intrusive! Later, we could add more the hvd-accelerated `tf.compat.v1` and `pytorch` backends.

To do:

- [ ] Avoid multiple imports of `import horovod.tensorflow as hvd`. I have to understand how to import horovod only once. I wonder where I shall put this import. I do not want the code to depend on `hvd`, if one does not want to use multiple GPU acceleration. 
- [ ] Define properly the collocation points. The idea would be to generate independent samples for each GPU. This is not done properly so far. With the current code, I am sending the same points to all the GPUs, and I add a resampling process `resampler = dde.callbacks.PDEResidualResampler(period= epochs // hvd.size())`. To my opinion, the most simple option would be to set properly the mini_batch option (see #320 )... Do you agree? 

Tested with the 2D Helmhlotz case (notice that I only added the resampler and `dde.config.set_hvd()`):

```
"""Backend supported: tensorflow.compat.v1, tensorflow, pytorch"""
import deepxde as dde
import numpy as np

dde.config.set_hvd()

# Set seed on 11
dde.config.set_random_seed(11)
# General parameters
n = 2
precision_train = 10
precision_test = 30
hard_constraint = True
weights = 100  # if hard_constraint == False
epochs = 5000
parameters = [1e-3, 3, 150, "sin"]

# Define sine function
if dde.backend.backend_name == "pytorch":
    sin = dde.backend.pytorch.sin
else:

    sin = tf.sin

learning_rate, num_dense_layers, num_dense_nodes, activation = parameters


def pde(x, y):

    f = k0 ** 2 * sin(k0 * x[:, 0:1]) * sin(k0 * x[:, 1:2])
    return -dy_xx - dy_yy - k0 ** 2 * y - f


def func(x):
    return np.sin(k0 * x[:, 0:1]) * np.sin(k0 * x[:, 1:2])


def transform(x, y):
    res = x[:, 0:1] * (1 - x[:, 0:1]) * x[:, 1:2] * (1 - x[:, 1:2])
    return res * y


def boundary(_, on_boundary):
    return on_boundary


geom = dde.geometry.Rectangle([0, 0], [1, 1])
k0 = 2 * np.pi * n
wave_len = 1 / n

hx_train = wave_len / precision_train
nx_train = int(1 / hx_train)

hx_test = wave_len / precision_test
nx_test = int(1 / hx_test)

if hard_constraint == True:
    bc = []
else:


data = dde.data.PDE(
    geom,
    pde,
    bc,
    num_domain=nx_train ** 2,
    num_boundary=4 * nx_train,
    solution=func,
    num_test=nx_test ** 2,
)

net = dde.nn.FNN(
    [2] + [num_dense_nodes] * num_dense_layers + [1], activation, "Glorot uniform"
)

if hard_constraint == True:
    net.apply_output_transform(transform)

model = dde.Model(data, net)

if hard_constraint == True:
    model.compile("adam", lr=learning_rate, metrics=["l2 relative error"])
else:
    loss_weights = [1, weights]
    model.compile(
        "adam",
        lr=learning_rate,
        metrics=["l2 relative error"],
        loss_weights=loss_weights,
    )
    
import horovod.tensorflow as hvd    
resampler = dde.callbacks.PDEResidualResampler(period= epochs // hvd.size())
losshistory, train_state = model.train(epochs=epochs, callbacks = [resampler])
```


pescap:
#579 #39

pescap:
FYI: https://github.com/horovod/horovod/issues/114

I am currently working on it. The first milestone would be to dispatch the training points over the GPUs. Maybe via a data generator: 

1. generate n_GPU*random_data of same size [N_domain, N_boundary]
2. or just generate n_GPU*random_data of same size [N_domain] and use the same boundary points for all the GPUs

Next, as soon as the horovod architecture is deployed, one could go towards domain decomposition-based solutions.

pescap:
Notes (work in progress): 

Train next batch for PDEs:
https://github.com/lululxvi/deepxde/blob/770e7c1f703682633fe182a6de984987fd579afa/deepxde/data/pde.py#L168

pescap:
I'll start from the beginning, as this PR is too outdated. A quick update:

- I am more familiar with `tf.compat.v1` and it has more models implemented. Therefore, I will implement the Horovod acceleration for this backend first (and not `tensorflow`, as for this original draft)
- At the end, the data point generation is straightforward. One uses random o pseudo-random generation, and each GPU is assigned its own `data`. Uniform distribution would lead to to the same training points over all GPUs...
- I keep accelerating the domain collocation points. I assume that the boundary points do not need to be distributed over GPU.

lululxvi:
@pescap You can also check https://github.com/lululxvi/deepxde/pull/1094, although it is only for backend paddle, but it has many useful functions.

pescap:
> @pescap You can also check #1094, although it is only for backend paddle, but it has many useful functions.

Thank you, I'll try to follow the same structure.

deepxde/config.py
```diff
@@ -7,6 +7,7 @@
 from .backend import backend_name, tf, torch, paddle
 from .real import Real
 
+hvd_dist = False
```
pescap, line 10: @lululxvi I'll keep using this variable to add or not Horovod. Let me know if the variable name is fine.

lululxvi: We can use this for now.

PR #735, opened 2022-06-14
A Makefile to discover and quickly run all examples

tvial:
Hi!

I noticed there was no automated test suite and since I would love to contribute to the package someday, I decided to use the examples as integration tests. They will cover a fraction of the code, hopefully a good fraction, so they can help discover regressions in face of e.g. a change in the DeepXDE API.

The idea is not to check the numerical results, rather to make sure that all scripts run end-to-end. For this, the sample scripts are slightly modified to accept a "test mode" where the number of iterations is very small and no side effects (file accesses, plot windows) are triggered. The code thus runs quickly and non interactively. It takes less than 10 min on my GPU-less laptop, this is enough for integration tests.

Here is the usage:
- `cd examples && make function/dataset.py`: run one example
- `cd examples && make function/*.py`: run all examples in `function` subdirectory
- `cd examples && make`: discover all Python sources and run them in turn (could go to a CI task)

There is one caveat: there is no provision to change whatever backend each sample requests, so much of the backend-specific code is likely not covered. I hope it's a good start though before introducing more testing layers, if needed.

What do you think?
Thomas


lululxvi:
Good idea. Testing is important. But one concern is that this modification may make the users confused when they read the example codes.

tvial:
Thanks, I appreciate the feedback!

Regarding files, I wanted to avoid leaving them behind and pollute the git status, but I can delete them instead or add them to `.gitignore`, either way. And apparently plotting can be disabled with a dummy matplotlib backend, I'll look into that, it's much better.

For the epochs override, if you'd rather have the scripts unmodified, I'll try to tweak them on the fly at run time, that should not be too troublesome. Also it will make writing new examples easier, without consideration for the unattended testing context.

lululxvi:
Yes, it sounds great.

lululxvi:
Good. Format the code via [black](https://github.com/psf/black)

examples/sample_to_test.py
```diff
@@ -0,0 +1,51 @@
...
+
+import matplotlib
+matplotlib.use('template')
+"""
+
+
+def transform(line, file_name):
+    """
+    Apply transformations to line
+    """
```
lululxvi, line 28: """Apply transformations to line."""

examples/sample_to_test.py
```diff
@@ -0,0 +1,51 @@
...
+    m = re.match(EPOCHS_RE, line)
+    if m is not None:
+        line = m.expand(r"\1epochs=1\3")
+
+    # Burgers_RAR.py has an additional convergence loop: force 1 single pass
+    if line.startswith("while") and "Burgers_RAR.py" in file_name:
```
lululxvi, line 34:
```python
if file_name = "Burgers_RAR.py" and line.startswith("while"):
```

PR #757, opened 2022-06-23
Add Pytorch implementation of AdamW optimizer

Jerry-Jzy:
## what is the purpose of this change
Add Pytorch implementation of AdamW optimizer


Jerry-Jzy:
> 


> AdamW has a parameter `weight_decay`, which should be problem-dependent. So it might not be proper to add AdamW.
> 
> Or you may need to use the weight decay parameter `decay` to set the `weight_decay`

PR #758, opened 2022-06-23
Add Pytorch implementation of regularization

Jerry-Jzy:
## what is the purpose of this change
Add Pytorch implementation of regularization


deepxde/optimizers/pytorch/optimizers.py
```diff
@@ -31,16 +31,18 @@ def get(params, optimizer, learning_rate=None, decay=None):
         if learning_rate is None:
             raise ValueError("No learning rate for {}.".format(optimizer))
         if optimizer == "sgd":
-            optim = torch.optim.SGD(params, lr=learning_rate)
+            optim = torch.optim.SGD(params, lr=learning_rate, weight_decay = decay["regular"])
```
lululxvi, line 34: Don't add `weight_decay` for SGD, RMSprop, Adam, etc. which are uncommon.

deepxde/model.py
```diff
@@ -124,7 +124,11 @@ def compile(
         elif backend_name == "tensorflow":
             self._compile_tensorflow(lr, loss_fn, decay, loss_weights)
         elif backend_name == "pytorch":
-            self._compile_pytorch(lr, loss_fn, decay, loss_weights)
+            if decay == None:
```
lululxvi, line 127: You shouldn't modify here. Only modify `self._compile_pytorch`

deepxde/nn/regularizers.py
```diff
@@ -2,7 +2,6 @@
 
 
 def get(identifier):
-    # TODO: pytorch
     if identifier is None:
```
lululxvi, line 5: # TODO: other backends

deepxde/optimizers/pytorch/optimizers.py
```diff
@@ -31,16 +31,18 @@ def get(params, optimizer, learning_rate=None, decay=None):
...
         elif optimizer == "rmsprop":
-            optim = torch.optim.RMSprop(params, lr=learning_rate)
+            optim = torch.optim.RMSprop(params, lr=learning_rate, weight_decay = decay["regular"])
         elif optimizer == "adam":
-            optim = torch.optim.Adam(params, lr=learning_rate)
+            optim = torch.optim.Adam(params, lr=learning_rate, weight_decay = decay["regular"])
+        elif optimizer == "adamw":
+            optim = torch.optim.AdamW(params, lr=learning_rate, weight_decay = decay["regular"])
```
lululxvi, line 40: Oh. `decay` is used for learning rate decay, not for regularization ... We should not use `decay`. sorry...

lululxvi:
Oh. `decay` is used for learning rate decay, not for regularization ... We should not use `decay`. sorry... So there is no variable for the weight decay in AdamW.

We might not add AdamW in DeepXDE. AdamW should be used directly in the user's main code.

Jerry-Jzy:
> 

how about using other optimizers with regularization?

lululxvi:
The standard regularization methods can be supported in a similar way as TensorFlow.

PR #764, opened 2022-06-24
Mini batch training using point set boundary condition

mattragoza:
This pull request implements `BatchPointSetBC`, as described in #744 . This is a subclass of `PointSetBC` that takes additional constructor arguments of `batch_size` and `shuffle`. It uses a `deepxde.data.sampler.BatchSampler` to maintain the internal state while iterating over the data points. Each call to `collocation_points` advances the iterator and returns the next batch of data points, then subsequent calls to `error` compute the boundary condition residual with respect to the data values for the current batch.

I also created an example script `examples/pinn_inverse/elliptic_inverse_field_batch.py` and verified that the performance is similar to the non-minibatch version.


mattragoza:
I am aware that the PR contains additional commits from my dev branch that were already merged in a previous PR. I'm not sure what the best practice is for omitting these commits from future pull requests given that this repo uses squash merging. Should I just create a separate branch for each pull request?

lululxvi:
> I am aware that the PR contains additional commits from my dev branch that were already merged in a previous PR. I'm not sure what the best practice is for omitting these commits from future pull requests given that this repo uses squash merging. Should I just create a separate branch for each pull request?

Yes, that would be better. It is fine this time.

mattragoza:
I have only tested with pytorch so far, but I will test with other backends.

mattragoza:
Maybe we can add an additional method `pde.resample_bc_points` to PDE, along with a BCResidualResampler callback?

lululxvi:
> Maybe we can add an additional method `pde.resample_bc_points` to PDE, along with a BCResidualResampler callback?

Yes, adding a callback is a good idea.

lululxvi:
It might be better to combine `BatchPointSetBC` and `PointSetBC` into one class, by checking if `batch_size` is `None`.

deepxde/callbacks.py
```diff
@@ -510,3 +510,12 @@ def on_epoch_end(self):
...
             )
+
+
+class BatchResampler(Callback):
+    """Resample the training points for both PDE and boundary condition losses after
+    each step."""
+
+    def on_batch_end(self):
+        self.model.data.train_x_bc = None
+        self.model.data.resample_train_points()
```
lululxvi, line 521: This will also resample the PDE training points. It might be better to add a function `resample_bc_points` in `dde.data.PDE`, similar to
https://github.com/lululxvi/deepxde/blob/09c74ab6a5c3269e248b09f27518a05d506b0327/deepxde/data/pde.py#L196

but probably a little more complicated .

mattragoza: I agree but this requires more changes to the PDE class. The approach I would take is to move the `run_if_all_none("train_x", "train_y", "train_aux_vars")` decorator from `pde.train_next_batch` onto `pde.train_points`. That way the caching of the PDE domain points and the BC points are cached separately, but `train_next_batch` runs each time and recomputes whichever points are needed.

mattragoza: Otherwise `pde.bc_points` won't ever get called to recompute the BC points. However this approach would also call the true solution function every time, which could be expensive. Any better ideas?

mattragoza: For my use case I want to resample both the PDE domain and BC points anyway.

lululxvi: Yes, it is a good idea to cache `pde.train_points` in `self.train_x_all`. But you can still keep `run_if_all_none("train_x", "train_y", "train_aux_vars")`.

If we resample, then re-calling the true solution function seems unavoidable.

deepxde/data/pde.py
```diff
@@ -113,11 +113,12 @@ def __init__(
 
         self.auxiliary_var_fn = auxiliary_var_function
 
-        # TODO: train_x_all is used for PDE losses. It is better to add train_x_pde explicitly.
-        self.train_x_all = None
-        self.train_x, self.train_y = None, None
+        self.train_x_pde = None
```
lululxvi, line 116: Don't modify the behavior of `train_x_all`, otherwise many other classes depending on `PDE` will break.

deepxde/callbacks.py
```diff
@@ -502,11 +502,19 @@ def on_epoch_end(self):
...
                 "`num_bcs` changed! Please update the loss function by `model.compile`."
             )
+
+
+class BatchResampler(Callback):
+    """Resample the training points for both PDE and boundary condition losses after
+    each step."""
+
+    def on_batch_end(self):
+        self.model.data.resample_train_points()
```
lululxvi, line 520: Resample on every itereation seems expensive. Maybe add `period` as `PDEResidualResampler`?

deepxde/data/pde.py
mattragoza: Ok. I found the variable and method names confusing, since `train_x_all` only contains the PDE points and `train_x` contains all the training points including PDE and BC points. I would be willing to open a separate pull request that makes this change and any necessary changes to other classes that would be affected (I think just `TimePDE`, `FPDE`, and `TimeFPDE`).

deepxde/callbacks.py
mattragoza: I will make this change.

deepxde/data/pde.py
lululxvi: Yes, use a separate PR. There are more classes using it, such as `PDEOperator`, `PDEOperatorCartesianProd`.

deepxde/callbacks.py
mattragoza: What do you think about combining into a single class that takes arguments for resampling the PDE points, BC points or both?

mattragoza: I.e. the constructor signature would be something like `BatchResampler(period=100, pde_points=True, bc_points=True)`.

lululxvi: Good idea. But we can do step by step. Maybe this can be another PR.

deepxde/callbacks.py
```diff
@@ -502,11 +502,28 @@ def on_epoch_end(self):
...
 
         if not np.array_equal(self.num_bcs_initial, self.model.data.num_bcs):
             print("Initial value of self.num_bcs:", self.num_bcs_initial)
             print("self.model.data.num_bcs:", self.model.data.num_bcs)
             raise ValueError(
                 "`num_bcs` changed! Please update the loss function by `model.compile`."
             )
+
+
+class BatchResampler(Callback):
```
lululxvi, line 515: How about `PDEResampler`?

deepxde/callbacks.py
lululxvi: Actually I agree with you it is good to combine them now, because there are so many same code.

deepxde/data/pde.py
```diff
@@ -225,6 +241,7 @@ def replace_with_anchors(self, anchors):
                 config.real(np)
             )
 
+    @run_if_all_none("train_x_all")
```
lululxvi, line 244: Do you also need to add this in line 311?

deepxde/icbc/boundary_conditions.py
```diff
@@ -17,6 +17,7 @@
 import numpy as np
 
 from .. import backend as bkd
+from .. import data
 from .. import config
```
lululxvi, line 21:
```
from .. import config
from .. import data
```

deepxde/data/pde.py
mattragoza: Yes, you're right

deepxde/callbacks.py
```diff
@@ -484,12 +484,20 @@ def on_train_end(self):
                 )
 
 
-class PDEResidualResampler(Callback):
-    """Resample the training points for PDE losses every given period."""
+class PDEResampler(Callback):
```
lululxvi, line 487: How about `PDEPointResampler`? so that it is easier to understand.

examples/pinn_forward/heat.py
```diff
@@ -1,3 +1,4 @@
+"""Backend supported: tensorflow.compat.v1, pytorch"""
```
lululxvi, line 1: Why not "tensorflow"?

mattragoza: I have not tested with tensorflow yet because I had trouble installing tensorflow_probability in my conda environment.

deepxde/icbc/boundary_conditions.py
```diff
@@ -180,11 +183,31 @@ def __init__(self, points, values, component=0):
...
     def collocation_points(self, X):
+        if self.batch_size is not None:
+            self.batch_indices = self.batch_sampler.get_next(self.batch_size)
+            return self.points[self.batch_indices]
         return self.points
 
     def error(self, X, inputs, outputs, beg, end, aux_var=None):
+        if self.batch_size is not None:
+            if backend_name == "tensorflow.compat.v1":
+                batch_values = bkd.tf.gather(self.values, self.batch_indices)
```
lululxvi, line 204: I am not sure if this would work or not, because in TF 1, `self.batch_indices` may only be initialized once in the graph.

Also, in TF 2, we use `@tf.function` for graph mode...

mattragoza: I really am not very familiar with tensorflow. Maybe I should just keep support for pytorch for now.

examples/pinn_forward/heat.py
lululxvi: OK. I just tested it, and made a few changes, see https://github.com/lululxvi/deepxde/commit/b927780de668ccb8f5b01c15985cf04cc905665c
You can remove it from your PR.

deepxde/icbc/boundary_conditions.py
lululxvi: Yes, I agree. Check the backend; if it is not PyTorch, then throw an error.

deepxde/callbacks.py
```diff
@@ -484,12 +484,20 @@ def on_train_end(self):
...
+    """Resample the training points for PDE and/or BC losses every given period.
 
-    def __init__(self, period=100):
+    Args:
+        period: How often to resample the training points (default is 100 iterations).
+        pde_points: If True, resample the training points for PDE losses.
+        bc_points: If True, resample the training points for BC losses.
+    """
+
+    def __init__(self, period=100, pde_points=True, bc_points=True):
```
lululxvi, line 496: Use `bc_points=False` as default, as it only supports PyTorch.

deepxde/icbc/boundary_conditions.py
```diff
@@ -169,9 +170,11 @@ class PointSetBC:
             used for training.
         values: An array of values that gives the exact solution of the problem.
         component: The output component satisfying this BC.
+        batch_size: The number of points per minibatch, or `None` to return all points.
```
lululxvi, line 173:
```
batch_size: The number of points per minibatch, or `None` to return all points. This is only supported for the backend PyTorch.
```

examples/pinn_forward/heat_resample.py
```diff
@@ -0,0 +1,112 @@
+"""Backend supported: tensorflow.compat.v1, pytorch"""
+import deepxde as dde
+import matplotlib.pyplot as plt
```
lululxvi, line 3: Delete `import matplotlib.pyplot as plt`

deepxde/callbacks.py
lululxvi: @mattragoza

lululxvi: @mattragoza Would you like to finish the step of this PR?

mattragoza: Done

Issue #805, opened 2022-07-22
Dump external trainable variable for external optimizer e.g. L-FGBS


tsarikahin:
### Overview
As discussed in issue #395  it is not implemented that one can not print out the external variable when an external optimizer is used such as `L-BFGS`. 

Question?
How can I get the external variable through the epoch? 

I went through the source code: I think I should change (loss_callback) or add a new callback?

```python
def _train_tensorflow_compat_v1_scipy(self, display_every):
        def loss_callback(loss_train):
            self.train_state.epoch += 1
            self.train_state.step += 1
            if self.train_state.step % display_every == 0:
                self.train_state.loss_train = loss_train
                self.train_state.loss_test = None
                self.train_state.metrics_test = None
                self.losshistory.append(
                    self.train_state.step, self.train_state.loss_train, None, None
                )
                display.training_display(self.train_state)
                print(self.callbacks.callbacks[0])

        self.train_state.set_data_train(*self.data.train_next_batch(self.batch_size))
        feed_dict = self.net.feed_dict(
            True,
            self.train_state.X_train,
            self.train_state.y_train,
            self.train_state.train_aux_vars,
        )
        self.train_step.minimize(
            self.sess,
            feed_dict=feed_dict,
            fetches=[self.outputs_losses[1]],
            loss_callback=loss_callback,
        )
        self._test()
```

**NOT:** External variable will be dumped only at the beginning of the external optimizer. I think it is forgotten:

https://github.com/lululxvi/deepxde/blob/master/deepxde/callbacks.py#L285

 ### Interested parties
@lululxvi

tsarikahin:
I added the print(self.callbacks.callbacks[0].value) but this command prints always the same value

lululxvi:
Good idea. You can change the callback `VariableValue` by adding `on_train_end`

tsarikahin:
@lululxvi thx for the reply. What about getting the results after each epoch? With adam it is possible. I could not figure it out how one can achieve that using an external optimizer?

tsarikahin:
I think I figured it out. I added step_callback function in `_train_tensorflow_compat_v1_scipy` from model.py

```python
def step_callback(y):
            if self.train_state.step % display_every == 0:
                print(y[:len(self.external_trainable_variables)])
```
This function prints out the trainable external variables on the terminal. Of course, one can write the results on the existing file as well. I will think about that and I will make a MR/PR so that it might help many people.

lululxvi:
[...]

Issue #395, opened 2021-10-11
callback for the optimized variable in inverse problem


sashahexe:
Dear Lu Lu,

could you please give me a hint. 

1) I am compiling the model first with adam and then with L-BFGS-B. I put the `dde.callbacks.VariableValue()` before the final `losshistory, train_state` parts but the file for the variable gets updated only once after adam is trained and never gets next optimized values from the  L-BFGS-B. Could you please help me what is there to change in the code?

`   model = dde.Model(data, net)

    model.compile('adam', lr=0.001, external_trainable_variables=[gamma])

    model.train(epochs=5000)   

    model.compile("L-BFGS-B", external_trainable_variables=[gamma])   
    
    model.train_step.optimizer_kwargs = {'options': {'maxfun': 1e5, 'ftol': 1e-20, 'gtol': 1e-20, 'eps': 1e-20, 'iprint': -1, 'maxiter': 1e5}} 

    variable = dde.callbacks.VariableValue([gamma], period=1000, filename="gamma_2.dat")

    losshistory, train_state = model.train(epochs=50000, callbacks=[variable]) `

Here is my code https://colab.research.google.com/drive/18I6pRggrKomVuJ2ZIM0OchrIPt_2YePH?usp=sharing

Thank you in advance for your time!

Pranay-Nagrani:
Hi Dr. Lulu,

Does this mean while solving an inverse problem to identify unknown model parameters, I can only use Adam optimizer to train my model and not L-BFGS-B?

Thanks in advance!

lululxvi:
You could. But you can only show the variable values at the end.

Pranay-Nagrani:
Hi Dr. Lulu,

I train my model first with Adam and then using L-BFGS-B. My train and test loss reduces when I optimize with L-BFGS-B but model parameters are the same as the last iteration of Adam. Am I missing something? The snippet of my code where I implement L-BFGS-B is:

model.compile("L-BFGS-B",external_trainable_variables=[G, eta_p, yield_stress])
variable = dde.callbacks.VariableValue([G, eta_p, yield_stress], period=10, filename="variables_1.dat")
losshistory, trainstate = model.train(callbacks = [variable])
dde.saveplot(losshistory, trainstate, issave=True, isplot=True)

lululxvi:
The code seems good.

okhsallam:
How can I know the last value for the variable trained by the LBFGS  ?

lululxvi:
[...]

PR #817, opened 2022-07-25
Dump out external trainable variable for external optimizer

tsarikahin:
closing #805 #395

### DIFF
--- a/deepxde/callbacks.py
+++ b/deepxde/callbacks.py
@@ -324,7 +324,7 @@ class VariableValue(Callback):
         if self.epochs_since_last >= self.period:
             self.epochs_since_last = 0
             self.on_train_begin()
-    
+
     def on_train_end(self):
         self.on_train_begin()
 
--- a/deepxde/callbacks.py
+++ b/deepxde/callbacks.py
@@ -304,6 +304,7 @@ class VariableValue(Callback):
         self.file = sys.stdout if filename is None else open(filename, "w", buffering=1)
         self.value = None
         self.epochs_since_last = 0
+        self.epochs_counter = 0
 
     def on_train_begin(self):
         if backend_name == "tensorflow.compat.v1":
@@ -321,9 +322,11 @@ class VariableValue(Callback):
 
     def on_epoch_end(self):
         self.epochs_since_last += 1
+        self.epochs_counter += 1
         if self.epochs_since_last >= self.period:
-            self.epochs_since_last = 0
-            self.on_train_begin()
+            if not (self.epochs_counter == self.model.num_iterations):
+                self.epochs_since_last = 0
+                self.on_train_begin()
 
     def on_train_end(self):
         self.on_train_begin()
--- a/deepxde/callbacks.py
+++ b/deepxde/callbacks.py
@@ -304,7 +304,6 @@ class VariableValue(Callback):
         self.file = sys.stdout if filename is None else open(filename, "w", buffering=1)
         self.value = None
         self.epochs_since_last = 0
-        self.epochs_counter = 0
 
     def on_train_begin(self):
         if backend_name == "tensorflow.compat.v1":
@@ -322,14 +321,13 @@ class VariableValue(Callback):
 
     def on_epoch_end(self):
         self.epochs_since_last += 1
-        self.epochs_counter += 1
         if self.epochs_since_last >= self.period:
-            if not (self.epochs_counter == self.model.num_iterations):
-                self.epochs_since_last = 0
-                self.on_train_begin()
+            self.epochs_since_last = 0
+            self.on_train_begin()
 
     def on_train_end(self):
-        self.on_train_begin()
+        if not self.epochs_since_last == 0:
+            self.on_train_begin()
 
     def get_value(self):
         """Return the variable values."""
--- a/deepxde/callbacks.py
+++ b/deepxde/callbacks.py
@@ -324,6 +324,9 @@ class VariableValue(Callback):
         if self.epochs_since_last >= self.period:
             self.epochs_since_last = 0
             self.on_train_begin()
+    
+    def on_train_end(self):
+        self.on_train_begin()
 
     def get_value(self):
         """Return the variable values."""
--- a/deepxde/model.py
+++ b/deepxde/model.py
@@ -16,6 +16,7 @@ from .backend import backend_name, tf, torch, jax, paddle
 from .callbacks import CallbackList
 from .utils import list_to_str
 
+
 class Model:
     """A ``Model`` trains a ``NN`` on a ``Data``.
 
@@ -616,21 +617,25 @@ class Model:
                     self.train_state.step, self.train_state.loss_train, None, None
                 )
                 display.training_display(self.train_state)
-        
+
         def ext_var_callback(y):
-            if len(self.external_trainable_variables)>0: # activated only for external optimizers since it is done adam already
+            # check if external variables are defined
+            if len(self.external_trainable_variables) > 0:
                 cb = self.callbacks.callbacks[0]
                 cb.epochs_since_last += 1
-                
+
                 if cb.epochs_since_last >= cb.period:
                     cb.epochs_since_last = 0
 
                     print(
-                    cb.model.train_state.epoch,
-                    list_to_str(y[:len(self.external_trainable_variables)], precision=cb.precision),
-                    file=cb.file,
+                        cb.model.train_state.epoch,
+                        list_to_str(
+                            y[: len(self.external_trainable_variables)],
+                            precision=cb.precision,
+                        ),
+                        file=cb.file,
                     )
-                    
+
                     cb.file.flush()
 
         self.train_state.set_data_train(*self.data.train_next_batch(self.batch_size))
@@ -645,7 +650,7 @@ class Model:
             feed_dict=feed_dict,
             fetches=[self.outputs_losses_train[1]],
             loss_callback=loss_callback,
-            step_callback=ext_var_callback
+            step_callback=ext_var_callback,
         )
         self._test()
 
--- a/deepxde/model.py
+++ b/deepxde/model.py
@@ -14,7 +14,6 @@ from . import optimizers
 from . import utils
 from .backend import backend_name, tf, torch, jax, paddle
 from .callbacks import CallbackList
-from .utils import list_to_str
 
 
 class Model:
@@ -617,7 +616,7 @@ class Model:
                 )
                 display.training_display(self.train_state)
 
-            if self.external_trainable_variables:
+                self.callbacks.on_batch_end()
                 self.callbacks.on_epoch_end()
         self.train_state.set_data_train(*self.data.train_next_batch(self.batch_size))
         feed_dict = self.net.feed_dict(
@@ -634,7 +633,7 @@ class Model:
             self.sess,
             feed_dict=feed_dict,
             fetches=fetches,
-            loss_callback=loss_callback
+            loss_callback=loss_callback,
         )
         self._test()
 
--- a/deepxde/model.py
+++ b/deepxde/model.py
@@ -30,6 +30,7 @@ class Model:
         self.net = net
 
         self.opt_name = None
+        self.num_iterations = None
         self.batch_size = None
         self.callbacks = None
         self.metrics = None
@@ -538,7 +539,7 @@ class Model:
                 " Use iterations instead."
             )
             iterations = epochs
-
+        self.num_iterations = iterations
         self.batch_size = batch_size
         self.callbacks = CallbackList(callbacks=callbacks)
         self.callbacks.set_model(self)
@@ -606,7 +607,7 @@ class Model:
                 break
 
     def _train_tensorflow_compat_v1_scipy(self, display_every):
-        def loss_callback(loss_train):
+        def loss_callback(loss_train, *args):
             self.train_state.epoch += 1
             self.train_state.step += 1
             if self.train_state.step % display_every == 0:
@@ -618,9 +619,7 @@ class Model:
                 )
                 display.training_display(self.train_state)
 
-        def ext_var_callback(y):
-            # check if external variables are defined
-            if len(self.external_trainable_variables) > 0:
+            if self.external_trainable_variables:
                 cb = self.callbacks.callbacks[0]
                 cb.epochs_since_last += 1
 
@@ -630,12 +629,11 @@ class Model:
                     print(
                         cb.model.train_state.epoch,
                         list_to_str(
-                            y[: len(self.external_trainable_variables)],
+                            [float(arg) for arg in args],
                             precision=cb.precision,
                         ),
                         file=cb.file,
                     )
-
                     cb.file.flush()
 
         self.train_state.set_data_train(*self.data.train_next_batch(self.batch_size))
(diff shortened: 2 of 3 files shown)


tsarikahin:
This merge request includes the following changes:

 - If an external optimizer function (e.g. L-BFGS) is used, it was not outputting or writing the external trainable variables after each epoch or in the training end. `In the training end` part was forgotten to implement.  Now one can output the external trainable variables after each epoch and at the very end as well
 - One example is tested if the algorithm works as expected

lululxvi:
Format your code via [black](https://github.com/psf/black)

tsarikahin:
@lululxvi I could not get why the tests could not pass? Is it because of formatting?

lululxvi:
> @lululxvi I could not get why the tests could not pass? Is it because of formatting?

No. It seems an issue of TensorFlow installation in the Travis system.

deepxde/callbacks.py
```diff
@@ -325,6 +325,9 @@ def on_epoch_end(self):
             self.epochs_since_last = 0
             self.on_train_begin()
 
+    def on_train_end(self):
```
lululxvi, line 328: Would adding this line makes two duplicated lines when training with Adam? Because the last result would be saved in both `on_epoch_end` and `on_train_end`.

deepxde/model.py
```diff
@@ -617,6 +618,26 @@ def loss_callback(loss_train):
...
+                cb = self.callbacks.callbacks[0]
+                cb.epochs_since_last += 1
+
+                if cb.epochs_since_last >= cb.period:
+                    cb.epochs_since_last = 0
+
+                    print(
+                        cb.model.train_state.epoch,
+                        list_to_str(
+                            y[: len(self.external_trainable_variables)],
```
lululxvi, line 633: There is no guarantee that the `external_trainable_variables` will be in the beginning.

deepxde/model.py
```diff
@@ -629,6 +650,7 @@ def loss_callback(loss_train):
             feed_dict=feed_dict,
             fetches=[self.outputs_losses_train[1]],
             loss_callback=loss_callback,
+            step_callback=ext_var_callback,
```
lululxvi, line 653: Do not use `step_callback`, as discussed above. Add the `external_trainable_variables` into `fetches`, and then use `loss_callback`.

examples/pinn_inverse/Lorenz_inverse.py
```diff
@@ -73,9 +73,20 @@ def boundary(_, on_initial):
...
+
+# train adam
+model.compile(
+    "adam", lr=0.001, external_trainable_variables=external_trainable_variables
 )
-losshistory, train_state = model.train(iterations=60000, callbacks=[variable])
+losshistory, train_state = model.train(iterations=20000, callbacks=[variable])
+
+# train lbfgs
+model.compile("L-BFGS-B", external_trainable_variables=external_trainable_variables)
```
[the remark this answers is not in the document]
tsarikahin: I am working on solid mechanic examples and here is the improvement:

image

As can be seen, there is a huge improvement with L-BFGS

deepxde/callbacks.py
tsarikahin: This can be avoided by modifying the `on_epoch_end` function. If the epoch number is not the last epoch, then continue.

deepxde/model.py
tsarikahin: I thought they should be at the very end which might make sense. I run a couple of examples and it was always at the very beginning. How can we be sure? Is there any way to do that?

examples/pinn_inverse/Lorenz_inverse.py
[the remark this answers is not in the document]
lululxvi: Good. How about this Lorenz inverse problem you modified?

deepxde/callbacks.py
lululxvi: Yes. You can modify it as you suggested.

deepxde/model.py
lululxvi: I don't know. That is why I suggest to use `loss_callback` as my next comment.

examples/pinn_inverse/Lorenz_inverse.py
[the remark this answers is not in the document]
tsarikahin: Sorry, I was busy. Here are the results for Lorenz's problem. Initially, only ADAM is used and here are the results for c1, c2 and c3: (actually values 10, 15 and 8/3) 

```bash
19900 [9.74e+00, 1.50e+01, 2.67e+00]
19910 [9.74e+00, 1.50e+01, 2.67e+00]
19920 [9.74e+00, 1.50e+01, 2.67e+00]
19930 [9.74e+00, 1.50e+01, 2.67e+00]
19940 [9.74e+00, 1.50e+01, 2.67e+00]
19950 [9.74e+00, 1.50e+01, 2.67e+00]
19960 [9.74e+00, 1.50e+01, 2.67e+00]
19970 [9.75e+00, 1.50e+01, 2.67e+00]
19980 [9.75e+00, 1.50e+01, 2.67e+00]
19990 [9.75e+00, 1.50e+01, 2.67e+00]
20000 [9.75e+00, 1.50e+01, 2.67e+00]
20000 [9.75e+00, 1.50e+01, 2.67e+00]
```

Adam + LBFGS, it took 2608 steps more for LBFGS. And we can see that C1 is improved and it is perfectly matching with analytical results.

```bash
22426 [1.00e+01, 1.50e+01, 2.67e+00]
22436 [1.00e+01, 1.50e+01, 2.67e+00]
22447 [1.00e+01, 1.50e+01, 2.67e+00]
22457 [1.00e+01, 1.50e+01, 2.67e+00]
22467 [1.00e+01, 1.50e+01, 2.67e+00]
22477 [1.00e+01, 1.50e+01, 2.67e+00]
22487 [1.00e+01, 1.50e+01, 2.67e+00]
22497 [1.00e+01, 1.50e+01, 2.67e+00]
22507 [1.00e+01, 1.50e+01, 2.67e+00]
22517 [1.00e+01, 1.50e+01, 2.67e+00]
22527 [1.00e+01, 1.50e+01, 2.67e+00]
22537 [1.00e+01, 1.50e+01, 2.67e+00]
22547 [1.00e+01, 1.50e+01, 2.67e+00]
22558 [1.00e+01, 1.50e+01, 2.67e+00]
22568 [1.00e+01, 1.50e+01, 2.67e+00]
22578 [1.00e+01, 1.50e+01, 2.67e+00]
22588 [1.00e+01, 1.50e+01, 2.67e+00]
22608 [1.00e+01, 1.50e+01, 2.67e+00]
22608 [1.00e+01, 1.50e+01, 2.67e+00]
````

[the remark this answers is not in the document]
Comment (tsarikahin) [reply examples/pinn_inverse/Lorenz_inverse.py:89]: Also, the initial number of epochs is set to 60000, which is a huge number. I think adam + LBFGS has a huge improvement in training.

File: deepxde/callbacks.py (hunk 1 above)
Comment (tsarikahin) [reply deepxde/callbacks.py:328]: I can not reach the number of `epochs`, or iterations inside of the callbacks class. Therefore, I will introduce `num_iterations` in model class as None and assign that variable using epochs in training.

Comment (tsarikahin) [reply deepxde/callbacks.py:328]: Done.

File: deepxde/model.py (hunk 3 above)
Comment (tsarikahin) [reply deepxde/model.py]: Done.

File: examples/pinn_inverse/Lorenz_inverse.py (hunk 4 above)
Comment (lululxvi) [inline examples/pinn_inverse/Lorenz_inverse.py:89]: Replace "L-BFGS-B" with "L-BFGS"

File: deepxde/callbacks.py [hunk 5]
```diff
@@ -321,9 +322,14 @@ def on_train_begin(self):
 
     def on_epoch_end(self):
         self.epochs_since_last += 1
+        self.epochs_counter += 1
         if self.epochs_since_last >= self.period:
-            self.epochs_since_last = 0
-            self.on_train_begin()
+            if not (self.epochs_counter == self.model.num_iterations):
```
lululxvi, line 327: We don't need `self.epochs_counter` and `self.model.num_iterations` to check if it is the last step. We can use the following trick. In `on_train_end`, we check if `self.epochs_since_last == 0`: if it is true, then we don't run `self.on_train_begin()`, otherwise, we run it. You can double check the logic.

deepxde/model.py
```diff
@@ -617,18 +619,39 @@ def loss_callback(loss_train):
                 )
                 display.training_display(self.train_state)
 
+            if self.external_trainable_variables:
+                cb = self.callbacks.callbacks[0]
```
lululxvi, line 623: There is no guarantee `self.callbacks.callbacks[0]` is the varialbe callback. You should call `self.callbacks.on_epoch_end()`

tsarikahin: Done.

deepxde/callbacks.py
tsarikahin: Done.

examples/pinn_inverse/Lorenz_inverse.py
tsarikahin: Done.

tsarikahin:
@lululxvi I could not comment on the file but line 562 in model.py I have comment.

```python
.
.
.
self.train_state.set_data_train(*self.data.train_next_batch(self.batch_size))
self.train_state.set_data_test(*self.data.test())
self._test()
self.callbacks.on_train_begin() # Line 562 
.
.
.
```
Line 562 also repeats the last epoch of the previous optimizer if more than one optimizer is used. I think it is not considered when it is implemented. I think this should be skipped if at least two optimizers are used. 

actually, line 562 is not needed at all. Because it is the initial guess before optimization starts. Do we really need it?

deepxde/model.py
```diff
@@ -14,6 +14,7 @@
 from . import utils
 from .backend import backend_name, tf, torch, jax, paddle
 from .callbacks import CallbackList
+from .utils import list_to_str
```
lululxvi, line 17: Delete

lululxvi:
> Line 562 also repeats the last epoch of the previous optimizer if more than one optimizer is used. I think it is not considered when it is implemented and it should be skipped if at least two optimizers are used.
> 
> Actually, line 562 is not needed at all. Because it is the initial guess before optimization starts. Do we really need it?

Good point. I suggest keeping it, as the initial values are useful and we may tune the initial values in some cases. It is a future work to resolve this repeated output between two optimizers.

deepxde/model.py
```diff
@@ -617,18 +617,24 @@ def loss_callback(loss_train):
...
+            fetches = [self.outputs_losses_train[1]] + self.external_trainable_variables
+        else:
+            fetches = [self.outputs_losses_train[1]]
         self.train_step.minimize(
             self.sess,
             feed_dict=feed_dict,
-            fetches=[self.outputs_losses_train[1]],
-            loss_callback=loss_callback,
+            fetches=fetches,
+            loss_callback=loss_callback
```
lululxvi, line 637: `loss_callback=loss_callback,`

deepxde/model.py
tsarikahin: Done.

deepxde/model.py
tsarikahin: Done.

deepxde/model.py
```diff
@@ -617,18 +617,24 @@ def loss_callback(loss_train):
                 )
                 display.training_display(self.train_state)
 
+            if self.external_trainable_variables:
+                self.callbacks.on_epoch_end()
```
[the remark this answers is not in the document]
tsarikahin: Done.

deepxde/model.py
```diff
@@ -617,18 +617,24 @@ def loss_callback(loss_train):
...
         self.train_state.set_data_train(*self.data.train_next_batch(self.batch_size))
         feed_dict = self.net.feed_dict(
             True,
             self.train_state.X_train,
             self.train_state.y_train,
             self.train_state.train_aux_vars,
         )
+        if self.external_trainable_variables:
+            fetches = [self.outputs_losses_train[1]] + self.external_trainable_variables
```
lululxvi, line 630: Because in the callback, we recompute the variable values, do we need to fetch them here?

deepxde/model.py
```diff
@@ -617,17 +616,23 @@ def loss_callback(loss_train):
                 )
                 display.training_display(self.train_state)
 
+                self.callbacks.on_batch_end()
+                self.callbacks.on_epoch_end()
```
lululxvi, line 620: These two lines should be outside the `if self.train_state.step % display_every == 0:`. callback has its own period.

tsarikahin: ahs sorry, yes.

deepxde/model.py
tsarikahin: If we don't fetch them, it will not update values so it will be always the last iteration of the adam optimizer.

deepxde/model.py
```diff
@@ -620,22 +618,7 @@ def loss_callback(loss_train, *args):
...
-                        cb.model.train_state.epoch,
-                        list_to_str(
-                            [float(arg) for arg in args],
-                            precision=cb.precision,
-                        ),
-                        file=cb.file,
-                    )
-                    cb.file.flush()
-
+                self.callbacks.on_epoch_end()
```
tsarikahin, line 620: It seems like this does not. Replacing lines 623-638 with `on_epoch_end` does not update the external variables so they don't change.

deepxde/model.py
tsarikahin: Can you please check the last comment?

deepxde/model.py
lululxvi: You are right. Then let us use the lines 623-638. You can check if there is a VariableValue callback.

tsarikahin: Done.

tsarikahin:
I have another remark regarding on outputting of the loss values in case of an external optimizer is used. Fetches contain only the training loss so it does not include the test loss. I think we should fetch the test losses as well right? It was lie this before, but I just realized it while plotting losses

deepxde/model.py
```diff
@@ -616,18 +616,38 @@ def loss_callback(loss_train):
                     self.train_state.step, self.train_state.loss_train, None, None
                 )
                 display.training_display(self.train_state)
-
+            for cb in self.callbacks.callbacks:
+                if type(cb).__name__=="VariableValue":
```
lululxvi, line 620: Refactor via black

deepxde/model.py
```diff
@@ -616,18 +616,38 @@ def loss_callback(loss_train):
...
+                    cb = self.callbacks.callbacks[0]
```
lululxvi, line 621: `cb = self.callbacks.callbacks[0]` why we need this?

lululxvi:
> I have another remark regarding on outputting of the loss values in case of an external optimizer is used. Fetches contain only the training loss so it does not include the test loss. I think we should fetch the test losses as well right? It was like this before, but I just realized it while plotting losses

Yes, you are right.

deepxde/model.py
tsarikahin: Sorry, I missed it. Done.

tsarikahin:
> > I have another remark regarding on outputting of the loss values in case of an external optimizer is used. Fetches contain only the training loss so it does not include the test loss. I think we should fetch the test losses as well right? It was like this before, but I just realized it while plotting losses
> 
> Yes, you are right.

Done.

deepxde/model.py
```diff
@@ -624,10 +642,17 @@ def loss_callback(loss_train):
             self.train_state.y_train,
             self.train_state.train_aux_vars,
         )
+        if self.external_trainable_variables:
+            fetches = [
+                self.outputs_losses_train[1],
+                self.outputs_losses_test[1],
+            ] + self.external_trainable_variables
+        else:
+            fetches = [self.outputs_losses_train[1], self.outputs_losses_test[1]]
```
lululxvi, line 651: test loss should also be added here.

lululxvi, line 651:
```python
fetches = [self.outputs_losses_train[1], self.outputs_losses_test[1]]
if self.external_trainable_variables:
    fetches += self.external_trainable_variables
```

PR #840, opened 2022-08-15
Add merger net and ouput merger net in MIONetCartesianProd and add monitor in ModelCheckpoint

Jerry-Jzy:
## purpose of this pull request
Add merger net and ouput merger net in MIONetCartesianProd and add monitor in ModelCheckpoint


deepxde/callbacks.py
```diff
@@ -119,14 +119,21 @@ class ModelCheckpoint(Callback):
...
     """
 
-    def __init__(self, filepath, verbose=0, save_better_only=False, period=1):
+    def __init__(
+        self,
+        filepath,
+        verbose=0,
+        save_better_only=False,
+        period=1,
+        monitor="train loss",
```
lululxvi, line 128: Update the docstring above.

lululxvi:
Only update the callback in this PR.

deepxde/callbacks.py
```diff
@@ -154,6 +161,16 @@ def on_epoch_end(self):
         else:
             self.model.save(self.filepath, verbose=self.verbose)
 
+    def get_monitor_value(self):
+        if self.monitor == "train loss":
+            result = sum(self.model.train_state.best_loss_train)
```
lululxvi, line 166: `result = sum(self.model.train_state.loss_train)`

deepxde/callbacks.py
```diff
@@ -154,6 +161,16 @@ def on_epoch_end(self):
...
+        elif self.monitor == "test loss":
+            result = sum(self.model.train_state.best_loss_test)
```
lululxvi, line 168: `result = sum(self.model.train_state.loss_test)`

PR #920, opened 2022-09-20
Update model.py

mitchelldaneker:
Allows pytorch to compute derivatives in the output transform.

### DIFF
--- a/deepxde/model.py
+++ b/deepxde/model.py
@@ -256,12 +256,14 @@ class Model:
 
         def outputs(training, inputs):
             self.net.train(mode=training)
-            with torch.no_grad():
-                if isinstance(inputs, tuple):
-                    inputs = tuple(map(torch.as_tensor, inputs))
-                else:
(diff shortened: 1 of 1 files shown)


deepxde/model.py
```diff
@@ -256,12 +256,14 @@ def _compile_pytorch(self, lr, loss_fn, decay, loss_weights):
 
         def outputs(training, inputs):
             self.net.train(mode=training)
-            with torch.no_grad():
-                if isinstance(inputs, tuple):
-                    inputs = tuple(map(torch.as_tensor, inputs))
-                else:
-                    inputs = torch.as_tensor(inputs)
-                return self.net(inputs)
+            if isinstance(inputs, tuple):
```
lululxvi, line 259: Try still keeping `with torch.no_grad():` and see if it works.

mitchelldaneker: It works, I have updated the code.

PR #1036, opened 2022-11-15
Documentation on how to approximate a function from a dataset

Jhayford:


docs/demos/FunctionApproximation/function.dataset.rst
```diff
@@ -0,0 +1,75 @@
+Approximating a Function From a Dataset
```
lululxvi, line 1: Rename this file as dataset.rst

lululxvi:
You need to merge Jonathan's code.

PR #1052, opened 2022-11-27
Add new publication based on DeepXDE

eslee1685:
added https://www.mdpi.com/1996-1073/15/20/7697


lululxvi:
Why there is PINN doc?

PR #1041, opened 2022-11-21
PointSetBC allows multi-component outputs

mitchelldaneker:
Allows PointSetBC to take a single input that multiple components share. This will save on computational cost for a large number of PointSetBCs.

PR #1058, opened 2022-12-01
Revert "PointSetBC supports multi-component outputs (#1041)"

mitchelldaneker:
This reverts commit fc0782f8764f3522dbf9239fece6b21fcdc63904.

As of now, all backend except pytorch do not allow for multi-components outputs.

### DIFF
--- a/deepxde/icbc/boundary_conditions.py
+++ b/deepxde/icbc/boundary_conditions.py
@@ -221,8 +221,8 @@ class PointSetBC:
                 outputs[beg:end, self.component]
                 - self.values[self.batch_indices]
             )
-        if backend_name in ["tensorflow.compat.v1", "tensorflow", "jax","paddle"]:
-            return outputs[beg:end, self.component[0]] - self.values
+        if backend_name in ["pytorch"]:
+            return outputs[beg:end, self.component] - self.values
         #When a concat is provided, the following code works 'fast' in paddle cpu, 
         #and slow in both tensorflow backends, jax untested.
         #tf.gather can be used instead of for loop but is also slow
@@ -235,7 +235,7 @@ class PointSetBC:
         #    calculated_error = outputs[beg:end, self.component[0]] - self.values
         #return calculated_error
         else:
-            return outputs[beg:end, self.component] - self.values
+            return outputs[beg:end, self.component[0]] - self.values
 
 
 class PointSetOperatorBC:
--- a/deepxde/icbc/boundary_conditions.py
+++ b/deepxde/icbc/boundary_conditions.py
@@ -167,11 +167,11 @@ class PointSetBC:
     Compare the output (that associates with `points`) with `values` (target data).
     If more than one component is provided via a list, the resulting loss will
     be the addative loss of the provided componets.
-    
+
     Args:
         points: An array of points where the corresponding target values are known and
             used for training.
-        values: An array of values that gives the exact solution of the problem.
+        values: A scalar or a 2D-array of values that gives the exact solution of the problem.
         component: Integer or a list of integers. The output components satisfying this BC.
             List of integers only supported for the backend PyTorch.
         batch_size: The number of points per minibatch, or `None` to return all points.
@@ -187,15 +187,20 @@ class PointSetBC:
         if isinstance(component, numbers.Number):
             self.component = [component]
         else:
-            if backend_name in ["tensorflow.compat.v1", "tensorflow", "jax", "paddle"]:
-               #TODO: Add support for multiple components in other backends
-               raise RuntimeError(
-                   "multiple components only implemented for pytorch backend"
-               ) 
+            if backend_name in [
+                "tensorflow.compat.v1",
+                "tensorflow",
+                "jax",
+                "paddle",
+            ]:
+                # TODO: Add support for multiple components in other backends
+                raise RuntimeError(
+                    "multiple components only implemented for pytorch backend"
+                )
             else:
                 self.component = component
         self.batch_size = batch_size
-        
+
         if batch_size is not None:  # batch iterator and state
             if backend_name != "pytorch":
                 raise RuntimeError(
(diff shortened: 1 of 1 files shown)


mitchelldaneker:
Currently, PointSetBC does not function for any backend except pytorch. Implementation for multiple components in pytorch only will be provided soon but reverting back to allow other backends to function properly.

lululxvi:
No need to totally revert back. Make it work for PyTorch.

tsarikahin:
image

lululxvi:
Format the code via black https://github.com/psf/black .

deepxde/icbc/boundary_conditions.py
```diff
@@ -167,12 +167,13 @@ class PointSetBC:
     Compare the output (that associates with `points`) with `values` (target data).
     If more than one component is provided via a list, the resulting loss will
     be the addative loss of the provided componets.
-
+    
     Args:
         points: An array of points where the corresponding target values are known and
             used for training.
-        values: A 2D-array of values that gives the exact solution of the problem.
+        values: An array of values that gives the exact solution of the problem.
```
lululxvi, line 174: values: A scalar or a 2D-array of values that gives the exact solution of the problem.

deepxde/icbc/boundary_conditions.py
```diff
@@ -214,7 +226,21 @@ def error(self, X, inputs, outputs, beg, end, aux_var=None):
...
+        # if len(self.component) > 1:
+        #    calculated_error = outputs[beg:end, self.component[0]] - self.values[:,0]
+        #    for i in range(1,len(self.component)):
+        #        tmp = outputs[beg:end, self.component[i]] - self.values[:,i]
+        #        calculated_error = bkd.lib.concat([calculated_error,tmp],axis=0)
+        # else:
+        #    calculated_error = outputs[beg:end, self.component[0]] - self.values
+        # return calculated_error
+        else:
+            return outputs[beg:end, self.component[0]] - self.values
```
lululxvi, line 243: Doesn't seem correct. `outputs[beg:end, self.component[0]:self.component[0]+1]`?
Check the original code

deepxde/icbc/boundary_conditions.py
```diff
@@ -186,7 +187,18 @@ def __init__(
         if isinstance(component, numbers.Number):
             self.component = [component]
```
lululxvi, line 188: Don't change `component` to `[component]`

mitchelldaneker: I now check is component is a list rather than automatically converting it to a list. As such, this is no longer needed and has been replaced

deepxde/icbc/boundary_conditions.py
mitchelldaneker: I believe the difference between `outputs[beg:end, self.component[0]:self.component[0]+1]` and `outputs[beg:end, self.component[0]]` is a difference in dimensionality. To prevent this from happening I go back to the original code when `self.component` is a single value. Also, since `self.component` is no longer a list I do not need to call it so I use `outputs[beg:end, self.component:self.component+1]`

deepxde/icbc/boundary_conditions.py
```diff
@@ -183,8 +184,14 @@ def __init__(
     ):
         self.points = np.array(points, dtype=config.real(np))
         self.values = bkd.as_tensor(values, dtype=config.real(bkd.lib))
-        if isinstance(component, numbers.Number):
-            self.component = [component]
+        if isinstance(component, list):
```
lululxvi, line 187:
```python
self.component = component
if isinstance(component, list) and backend_name != "pytorch":
    raise ...
```

deepxde/icbc/boundary_conditions.py
```diff
@@ -210,12 +217,36 @@ def collocation_points(self, X):
 
     def error(self, X, inputs, outputs, beg, end, aux_var=None):
         if self.batch_size is not None:
-            return (
-                outputs[beg:end, self.component]
-                - self.values[self.batch_indices]
-            )
-        return outputs[beg:end, self.component] - self.values
-
+            if isinstance(self.component, list):
```
lululxvi, line 220: use `if isinstance(component, numbers.Number):` to swap the order of list and number

deepxde/icbc/boundary_conditions.py
```diff
@@ -210,12 +217,36 @@ def collocation_points(self, X):
...
+                return (
+                    outputs[beg:end, self.component]
+                    - self.values[self.batch_indices]
+                )
+            else:
```
lululxvi, line 221: No need to use `else`

deepxde/icbc/boundary_conditions.py
```diff
@@ -210,12 +217,36 @@ def collocation_points(self, X):
...
+                return (
+                    outputs[beg:end, self.component : self.component + 1]
+                    - self.values[self.batch_indices]
+                )
+        else:
```
lululxvi, line 226: No need `else`

deepxde/icbc/boundary_conditions.py
```diff
@@ -210,12 +217,36 @@ def collocation_points(self, X):
...
+            # tf.gather can be used instead of for loop but is also slow
+            # if len(self.component) > 1:
+            #    calculated_error = outputs[beg:end, self.component[0]] - self.values[:,0]
+            #    for i in range(1,len(self.component)):
+            #        tmp = outputs[beg:end, self.component[i]] - self.values[:,i]
+            #        calculated_error = bkd.lib.concat([calculated_error,tmp],axis=0)
+            # else:
+            #    calculated_error = outputs[beg:end, self.component[0]] - self.values
+            # return calculated_error
+            else:
```
lululxvi, line 240: No need `else`

deepxde/icbc/boundary_conditions.py
mitchelldaneker: done for both batch and non-batch

deepxde/icbc/boundary_conditions.py
```diff
@@ -210,10 +213,31 @@ def collocation_points(self, X):
...
+                    outputs[beg:end, self.component : self.component + 1]
+                    - self.values[self.batch_indices]
+                )
             return (
                 outputs[beg:end, self.component]
                 - self.values[self.batch_indices]
             )
+        if isinstance(self.component, numbers.Number):
+            return (
+            outputs[beg:end, self.component : self.component + 1]
```
lululxvi, line 227: Format via black.

PR #1070, opened 2022-12-07
Support more API and example in PaddlePaddle

HydrogenSulfate:
1. support mor API in paddle and pytorch, tensorflow.v1 backend
2. unify API from `framework.xxx` to `bkd.xxx`
3. polish docs and comments in related file header

PR #1078, opened 2022-12-08
feat(utils): replace framework.xxx with bkd.xxx

HydrogenSulfate:
from PR https://github.com/lululxvi/deepxde/pull/1070

1. replace framework.xxx with bkd.xxx in array_ops_compat.py
2. refine cache stategy when alpha is an Variable

### DIFF
--- a/deepxde/data/fpde.py
+++ b/deepxde/data/fpde.py
@@ -124,8 +124,8 @@ class FPDE(PDE):
 
     def train_next_batch(self, batch_size=None):
         # do not cache train data when alpha is a learnable parameter
-        if backend_name == "tensorflow.compat.v1" or not is_tensor(self.alpha):
-            if hasattr(self, "train_x") or hasattr(self, "train_y"):
+        if not is_tensor(self.alpha) or backend_name == "tensorflow.compat.v1":
+            if self.train_x is not None:
                 return self.train_x, self.train_y
         if self.disc.meshtype == "static":
             if self.geom.idstr != "Interval":
@@ -154,8 +154,8 @@ class FPDE(PDE):
 
     def test(self):
         # do not cache test data when alpha is a learnable parameter
-        if backend_name == "tensorflow.compat.v1" or not is_tensor(self.alpha):
-            if hasattr(self, "test_x") or hasattr(self, "test_y"):
+        if not is_tensor(self.alpha) or backend_name == "tensorflow.compat.v1":
+            if self.test_x is not None:
                 return self.test_x, self.test_y
 
         if self.disc.meshtype == "static" and self.num_test is not None:
--- a/deepxde/data/fpde.py
+++ b/deepxde/data/fpde.py
@@ -2,7 +2,6 @@ __all__ = ["Scheme", "FPDE", "TimeFPDE"]
 
 import math
 import numpy as np
-import types
 
 from .pde import PDE
 from .. import config
--- a/deepxde/data/fpde.py
+++ b/deepxde/data/fpde.py
@@ -1,12 +1,12 @@
 __all__ = ["Scheme", "FPDE", "TimeFPDE"]
 
 import math
-
 import numpy as np
+import types
 
 from .pde import PDE
 from .. import config
-from ..backend import is_tensor, tf
+from ..backend import is_tensor, tf, backend_name
 from ..utils import array_ops_compat, run_if_all_none
 
 
@@ -94,6 +94,11 @@ class FPDE(PDE):
             solution=solution,
             num_test=num_test,
         )
+
+        # do not use cache when alpha is a learnable parameter
+        if is_tensor(self.alpha) and backend_name != "tensorflow.compat.v1":
+            self.train_next_batch = types.MethodType(self.train_next_batch.__wrapped__, self)
+            self.test = types.MethodType(self.test.__wrapped__, self)
 
     def losses_train(self, targets, outputs, loss_fn, inputs, model, aux=None):
         bcs_start = np.cumsum([0] + self.num_bcs)
--- a/deepxde/data/fpde.py
+++ b/deepxde/data/fpde.py
@@ -95,11 +95,6 @@ class FPDE(PDE):
             num_test=num_test,
         )
 
-        # do not use cache when alpha is a learnable parameter
-        if is_tensor(self.alpha) and backend_name != "tensorflow.compat.v1":
-            self.train_next_batch = types.MethodType(self.train_next_batch.__wrapped__, self)
-            self.test = types.MethodType(self.test.__wrapped__, self)
-
     def losses_train(self, targets, outputs, loss_fn, inputs, model, aux=None):
         bcs_start = np.cumsum([0] + self.num_bcs)
         int_mat = self.get_int_matrix(True)
@@ -128,8 +123,11 @@ class FPDE(PDE):
             loss_fn(tf.zeros(tf.shape(fi), dtype=config.real(tf)), fi) for fi in f
         ] + [tf.constant(0, dtype=config.real(tf)) for _ in self.bcs]
 
-    @run_if_all_none("train_x", "train_y")
     def train_next_batch(self, batch_size=None):
+        # do not cache train data when alpha is a learnable parameter
+        if backend_name == "tensorflow.compat.v1" or not is_tensor(self.alpha):
+            if hasattr(self, "train_x") or hasattr(self, "train_y"):
+                return self.train_x, self.train_y
         if self.disc.meshtype == "static":
             if self.geom.idstr != "Interval":
                 raise ValueError("Only Interval supports static mesh.")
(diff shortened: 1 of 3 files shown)


deepxde/utils/internal.py
```diff
@@ -33,7 +33,7 @@ def decorator(func):
         @wraps(func)
         def wrapper(self, *args, **kwargs):
             x = [getattr(self, a) for a in attr]
-            if all(i is None for i in x):
+            if all(i is None for i in x) or (hasattr(self, "alpha") and bkd.is_tensor(self.alpha)):
```
lululxvi, line 36: What is alpha?

deepxde/utils/array_ops_compat.py
```diff
@@ -2,8 +2,9 @@
 
 import numpy as np
 
+from .. import backend as bkd
 from .. import config
-from ..backend import is_tensor, tf
+from ..backend import is_tensor, tf, get_preferred_backend
```
lululxvi, line 7: Use backend_name, instead of get_preferred_backend

deepxde/utils/internal.py
HydrogenSulfate: > What is alpha?

In FPDE inverse problem such as `fractional_Poisson_2d_inverse.py`, `int_mat` will be calculated with an learnable parameter `alpha`, so I guess in dygraph based framwork, those `int_mat` should not be cached when alpha is an learnable parameter. Same for `FPDE.test`

HydrogenSulfate: > What is alpha?

Although It seems that tensorflow do not have this problem...

deepxde/utils/array_ops_compat.py
HydrogenSulfate: Done

deepxde/utils/internal.py
lululxvi: `run_if_all_none` has a clear definition: if all the attributes are None, then turn the function and return, otherwise directly return. If it doesn't work for FPDE, then we should modify FPDE, not this function.

deepxde/utils/array_ops_compat.py
```diff
@@ -13,21 +14,23 @@ def istensorlist(values):
 def convert_to_array(value):
     """Convert a list to numpy array or tensorflow tensor."""
```
lululxvi, line 15: Convert a list of numpy arrays or tensors to a numpy array or a tensor.

deepxde/utils/array_ops_compat.py
```diff
@@ -13,21 +14,23 @@ def istensorlist(values):
...
     if istensorlist(value):
-        return tf.convert_to_tensor(value, dtype=config.real(tf))
+        if backend_name == "paddle":
```
lululxvi, line 17: Why bkd.as_tensor cannot be used for paddle? paddle.to_tensor works for a list of tensor.

deepxde/utils/array_ops_compat.py
```diff
@@ -13,21 +14,23 @@ def istensorlist(values):
...
+        return bkd.as_tensor(value, dtype=config.real(bkd.lib))
     value = np.array(value)
     if value.dtype != config.real(np):
         return value.astype(config.real(np))
     return value
 
 
 def hstack(tup):
-    if not is_tensor(tup[0]) and tup[0] == []:
+    if not is_tensor(tup[0]) and isinstance(tup[0], list) and tup[0] == []:
```
lululxvi, line 27: Why needs `isinstance(tup[0], list)`?

deepxde/utils/array_ops_compat.py
HydrogenSulfate: > Why bkd.as_tensor cannot be used for paddle? paddle.to_tensor works for a list of tensor.

It may cause an error in backward progress
3FB799A034419E3FB816DE69968246AF
when input type is List[Tensor], it is more suitable to use `paddle.concat` and `paddle.stack` than `paddle.to_tensor` according to their usages

deepxde/utils/array_ops_compat.py
HydrogenSulfate: Deleted

deepxde/utils/internal.py
HydrogenSulfate: > `run_if_all_none` has a clear definition: if all the attributes are None, then turn the function and return, otherwise directly return. If it doesn't work for FPDE, then we should modify FPDE, not this function.

It seems `self.method = types.MethodType(self.method.__wrapped__, self)` is an easy way to achieve this.

deepxde/data/fpde.py
```diff
@@ -95,6 +95,11 @@ def __init__(
             num_test=num_test,
         )
 
+        # do not use cache when alpha is a learnable parameter
+        if is_tensor(self.alpha) and backend_name != "tensorflow.compat.v1":
+            self.train_next_batch = types.MethodType(self.train_next_batch.__wrapped__, self)
```
lululxvi, line 100: I am sort of understanding how this works. But `types.MethodType` is not really designed for this purpose. I am not sure if this is an abuse `types.MethodType`. Another simple solution is

- Delete `@run_if_all_none("train_x", "train_y")`
- In the beginning of `train_next_batch`, we check if alpha is tensor, and then check if `self.train_x` and `self.train_y` are None.

I think this makes the code easier to understand. What do you think?

HydrogenSulfate: > I am sort of understanding how this works. But `types.MethodType` is not really designed for this purpose. I am not sure if this is an abuse `types.MethodType`. Another simple solution is
> 
> * Delete `@run_if_all_none("train_x", "train_y")`
> * In the beginning of `train_next_batch`, we check if alpha is tensor, and then check if `self.train_x` and `self.train_y` are None.
> 
> I think this makes the code easier to understand. What do you think?

I agree with that👍

deepxde/utils/array_ops_compat.py
HydrogenSulfate: > Oh, this is tricky. It is not clear from the docs https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/to_tensor_en.html Is there any plan to make paddle.to_tensor work for this?

`paddle.to_tensor` is not an computational OP, and will not create edges on the graph, so `to_tensor` should not be used for concatenating tensors, which is consistent with pytorch

deepxde/utils/array_ops_compat.py
```diff
@@ -2,18 +2,21 @@
...
 
 
 def convert_to_array(value):
-    """Convert a list to numpy array or tensorflow tensor."""
+    """Convert a list of numpy arrays or tensors to a numpy array or a tensor."""
     if istensorlist(value):
-        return tf.convert_to_tensor(value, dtype=config.real(tf))
+        if backend_name == "paddle":
+            return bkd.concat(value, axis=0)
+        return bkd.as_tensor(value, dtype=config.real(bkd.lib))
```
lululxvi, line 19: How about also using `bkd.concat` for tf?

deepxde/data/fpde.py
```diff
@@ -123,8 +122,11 @@ def losses_test(self, targets, outputs, loss_fn, inputs, model, aux=None):
             loss_fn(tf.zeros(tf.shape(fi), dtype=config.real(tf)), fi) for fi in f
         ] + [tf.constant(0, dtype=config.real(tf)) for _ in self.bcs]
 
-    @run_if_all_none("train_x", "train_y")
     def train_next_batch(self, batch_size=None):
+        # do not cache train data when alpha is a learnable parameter
+        if backend_name == "tensorflow.compat.v1" or not is_tensor(self.alpha):
+            if hasattr(self, "train_x") or hasattr(self, "train_y"):
```
lululxvi, line 128:
```
if self.train_x is None:
```

deepxde/utils/array_ops_compat.py
HydrogenSulfate: Done, and it works well with **fractional_diffusion_1d** and **fractional_Poisson_1d_inverse**

deepxde/data/fpde.py
HydrogenSulfate: Done, and changed same as in `FPDE.test`

deepxde/data/fpde.py
```diff
@@ -1,12 +1,11 @@
 __all__ = ["Scheme", "FPDE", "TimeFPDE"]
 
 import math
```
lululxvi, line 3: Keep this empty line 4

deepxde/utils/array_ops_compat.py
```diff
@@ -2,18 +2,19 @@
 
 import numpy as np
 
+from .. import backend as bkd
 from .. import config
-from ..backend import is_tensor, tf
+from ..backend import is_tensor, tf, backend_name
```
lululxvi, line 7: No need to import `backend_name`

deepxde/data/fpde.py
HydrogenSulfate: Done

deepxde/utils/array_ops_compat.py
HydrogenSulfate: Done

lululxvi:
Please resolve the conflicts.

HydrogenSulfate:
> Please resolve the conflicts.

Done

PR #1097, opened 2022-12-27
add schrodinger.py

xiaoguoguo626807:
add Schrodinger example


lululxvi:
There is already Schrodinger eq at https://github.com/lululxvi/deepxde/blob/master/examples/pinn_forward/Schrodinger.ipynb
Merge this with yours.

PR #1108, opened 2023-01-12
 Allow specifying amount of boundary condition test points 

g-w1:
Something similar can be done in `data/pde_operator.py` and I may do this in a future pull request.

Should I add `num_test_boundary` to the examples? I'm not sure if this is a problem in the real-world. It seems to me that doing this would prevent (or at least detect) overfitting to the boundary conditions selected.

I also fixed a trivial TODO in `pde.py` (rename `train_x_all` to `train_x_pde`). Let me know if you want me to split this into a seperate PR.

### DIFF
--- a/deepxde/data/pde.py
+++ b/deepxde/data/pde.py
@@ -25,9 +25,10 @@ class PDE(Data):
             `num_boundary` sampled points.
         exclusions: A Numpy array of points to be excluded for training.
         solution: The reference solution.
-        num_test: The number of points sampled inside the domain for testing PDE loss.
-            The testing points for BCs/ICs are the same set of points used for training.
+        num_test_domain: The number of points sampled inside the domain for testing PDE loss.
             If ``None``, then the training points will be used for testing.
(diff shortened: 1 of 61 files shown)


lululxvi:
> I also fixed a trivial TODO in `pde.py` (rename `train_x_all` to `train_x_pde`). Let me know if you want me to split this into a seperate PR.

Yes, too many changes in this PR. Please split.

g-w1:
> Yes, too many changes in this PR. Please split.

I have split it into #1113 .

lululxvi:
There might be some classes that also use `num_test`. You may double check.

lululxvi:
Another main issue is that in `losses`, we use `self.num_bcs` for both training and testing. If you generate test boundary points separately, then `self.num_bcs` doesn't work for test points.

g-w1:
I will modify `ide.py` and `fpde.py` to add `num_test_domain` and `num_test_boundary` and will creat `self.num_bcs_train` and `self.num_bcs_test`. Thank you for the feedback.

PR #1113, opened 2023-01-15
rename train_x_all to train_x_pde

g-w1:
This resolves a TODO

Was split off from #1108

### DIFF
--- a/deepxde/data/pde.py
+++ b/deepxde/data/pde.py
@@ -53,18 +53,18 @@ class PDE(Data):
             error= dde.metrics.l2_relative_error(y_true, y_pred)
 
     Attributes:
-        train_x_all: A Numpy array of points for PDE training. `train_x_all` is
+        train_x_pde: A Numpy array of points for PDE training. `train_x_pde` is
             unordered, and does not have duplication. If there is PDE, then
-            `train_x_all` is used as the training points of PDE.
+            `train_x_pde` is used as the training points of PDE.
         train_x_bc: A Numpy array of the training points for BCs. `train_x_bc` is
-            constructed from `train_x_all` at the first step of training, by default it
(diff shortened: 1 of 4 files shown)


lululxvi:
have you double-checked if all classes using `train_x_all` has been revised?

g-w1:
Grepping for `train_x_all` returns nothing on this branch. However, I just realized that in #1108 I use `train_x_pde`, so it would make sense to merge this pull request before that one since it depends on this one.

deepxde/data/pde.py
```diff
@@ -100,9 +100,7 @@ def __init__(
 
         self.auxiliary_var_fn = auxiliary_var_function
 
-        # TODO: train_x_all is used for PDE losses. It is better to add train_x_pde
-        # explicitly.
```
lululxvi, line 104: As commented here, it is better to add `train_x_pde`, not replace `train_x_all` with `train_x_pde`. In some cases, we may actually use the meaning of `train_x_all`, but I am not sure which case. Need to look at all the code carefully.

g-w1: I have looked over the code, and I can't find a place where `train_x_all` does not contain PDE residual points.

If I added `train_x_pde` in addition to `train_x_all`, what would be different about them? The only thing that I can think of is that `train_x_pde` would only contain points in the domain as opposed to `train_x_all` also containing points on the boundary (or on the initial conditions in TimePDE).

What do you recommend for how to add it?

lululxvi: `train_x_all` contains all points (unordered, and does not have duplication). `train_x_bc` is the points for BC losses (may have have duplication).

There are cases where there are only BC losses and no PDE losses. So, it is better to add `train_x_pde`.

Issue #1111, opened 2023-01-15
Enhancement: Improve DeepXDE's packaging and build infrastructure


Saransh-cpp:
This is a placeholder issue for the infrastructure developments of DeepXDE in continuation of a discussion I had with Prof. Lu. 

Things I am planning to work on in the next couple months -

- [ ] Migrating DeepXDE's CI from Travis to GitHub Actions
- [ ] Automating DeepXDE's PyPI release using GitHub Actions
- [ ] Update the build configuration files along with the metadata (migrate metadata from `setup.py` and `setup.cfg` to `pyproject.toml` - [PEP 621](https://peps.python.org/pep-0621/), [PEP 660](https://peps.python.org/pep-0660/), https://github.com/pypa/setuptools/issues/3214, etc, etc.)
- [ ] Maybe update the build backend from setuptools to [hatch](https://github.com/pypa/hatch) (as DeepXDE is a pure Python project and setuptools+distutils is in a very messy state right now (given that Python 3.12 is dropping distutils for good and setuptools monkeypatches distutils - [PEP 632](https://peps.python.org/pep-0632/)))

Long term -

- [ ] Add tests
- [ ] Add [Codecov](https://about.codecov.io) integration for coverage statistics

Saransh-cpp:
Closing this as complete 🥳 

See #1244 for the testing developments.

PR #1117, opened 2023-01-18
Switch to `pyproject.toml`

Saransh-cpp:
XRef #1111

Migrates build config and metadata to pyproject.toml

## Description
This PR migrates -
- DeepXDE's metadata to the `pyproject.toml` file.
  - This is now extremely encouraged because of [PEP 621](https://peps.python.org/pep-0621/), [PEP 660](https://peps.python.org/pep-0660/), and https://github.com/pypa/setuptools/issues/3214.
- DeepXDE's build backend from `setuptools` to `hatch`.
  - `Hatch` is a modern build backend and is the recent best choice for pure Python projects. `setuptools` has been lagging in meeting modern PEPs. `setuptools` also uses `distutils` under the hood which should be avoided unless extremely necessary (for instance, a `Python - C++ project`). Python `3.12` is deprecating `distutils` for good and this has made the `setuptools-distutils` ecosystem a bit messy.
  - For instance, `SciPy` has already moved away from `setuptools+distutils`, and it now uses the `meson` build system. `NumPy` is planning to migrate away very soon, before the release of Python `3.12`.
  - A lot of big pure Python projects are migrating to hatch - https://hatch.pypa.io/dev/community/users/

## Feedback required
It would be nice if `DeepXDE` users can install `DeepXDE` from this branch and let me know if they face any issues -
```
pip install .
```

## Fixes required
Once this PR is in, and once a new release is created, the conda build will fail. `deepxde-feedstock`'s maintainer will need to  add `hatchling` here -
https://github.com/conda-forge/deepxde-feedstock/blob/e39744e8ef4b58d5f602b7655794ac4315d0000f/recipe/meta.yaml#L17-L21

## Future work in this direction
- Automate versioning
  - Right now `DeepXDE` requires manual updation of the version number on every release. This can be automated using `hatch-vcs`. `hatch-vcs` pull the version from `git`, removing the need to update it manually.


Saransh-cpp:
Merging #1122 before this PR would be better. That way we would be able to test the hatch build in GH Actions.

.gitignore
```diff
@@ -10,3 +10,6 @@ docs/_build/
 
 # VSCode
 .vscode/
+
+# Build files
+dist/
```
lululxvi, line 15: Re-Format the order and comments using the template at https://github.com/github/gitignore/blob/main/Python.gitignore

pyproject.toml
```diff
@@ -0,0 +1,55 @@
...
+[project]
+name = "DeepXDE"
+dynamic = ["version"]
+description = "A library for scientific machine learning"
+readme = "README.md"
+license = "LGPL-2.1"
+authors = [
+    { name = "Lu Lu", email = "lululxvi@gmail.com" },
+]
+requires-python = ">=3.7,<3.10"
```
lululxvi, line 14: Remove "<3.10"

pyproject.toml
```diff
@@ -0,0 +1,55 @@
...
+    "scikit-optimize>=0.9.0",
+    "scipy",
+]
+
+[project.urls]
+Homepage = "https://github.com/lululxvi/deepxde"
+"Bug Tracker" = "https://github.com/lululxvi/deepxde/issues"
+Changelog = "https://github.com/lululxvi/deepxde/releases"
+Documentation = "https://deepxde.readthedocs.io/en/latest/"
+
```
lululxvi, line 49: Why two empty lines here?

eli-schwartz:
> - setuptools also uses distutils under the hood which should be avoided unless extremely necessary (for instance, a Python - C++ project). Python 3.12 is deprecating distutils for good and this has made the setuptools-distutils ecosystem a bit messy.
> - For instance, SciPy has already moved away from setuptools+distutils, and it now uses the meson build system. NumPy is planning to migrate away very soon, before the release of Python 3.12.

This isn't really a reason to avoid setuptools. The distutils migration was specifically about putting setuptools in charge of it -- setuptools' copy of distutils is NOT deprecated.

The reason that SciPy and NumPy moved away is not because setuptools is inherently bad for using distutils - rather, it's because `numpy.distutils` was a very complex bit of internals, which would need to be redeveloped if it was going to be viable after setuptools adopted and updated distutils.

And they cannot use setuptools on its own, because setuptools doesn't support Fortran. So the setuptools ecosystem is indeed a bit messy... for fortran users. Which seems a bit niche compared to pure-python software. ;)

Saransh-cpp:
(Thank you for the insights! Please correct me if I am wrong. I am always learning.)

> The distutils migration was specifically about putting setuptools in charge of it -- setuptools' copy of distutils is NOT deprecated.

Yes, but doesn't setuptools monkeypatch large portions of distutils? I have had personal experiences of running into packaging problems because a stable release of setuptools decided to remove a module, which monkey-patched a public `distutils` module. Pushing a bug to prod is human, but why depend on `setuptools` when there is a build backend which works like magic for pure python projects?

> The reason that SciPy and NumPy moved away is not because setuptools is inherently bad for using distutils - rather, it's because numpy.distutils was a very complex bit of internals, which would need to be redeveloped if it was going to be viable after setuptools adopted and updated distutils.

This is indeed what I was trying to convey here, but I did not go into detail (and it might've come out wrong). Messy setuptools-distutils dependency -> breaks numpy.distutils -> breaks a lot of research software -> propagates downstream. In fact, I was the one who fixed this issue in a few research software. Since then, I have been encouraging users to switch to hatch if they have a pure-python project.

> And they cannot use setuptools on its own, because setuptools doesn't support Fortran. So the setuptools ecosystem is indeed a bit messy... for fortran users. Which seems a bit niche compared to pure-python software. ;)

True, but again, why not simply use `hatch` for pure Python projects and `setuptools` for projects binding C/C++ code with Python? I am definitely not against `setuptools`, I do work on projects that use `setuptools`, I just find `hatch` much better and updated for pure Python projects. `setuptools` has also been a bit slow in adopting modern PEPs (for instance, editable installs with `pyproject.toml`).

---

But yes, most of these build failures will never happen in DeepXDE because it is a pure-python library. In the end, it would be the maintainer's decision. I will be more than happy to switch back to `setuptools` while keeping the `pyproject.toml` file! :)

eli-schwartz:
> `setuptools` has also been a bit slow in adopting modern PEPs (for instance, editable installs with `pyproject.toml`).

And on the flip side, it invented editable installs and was, for most of python history, the only way to do them. :P

> True, but again, why not simply use `hatch` for pure Python projects and `setuptools` for projects binding C/C++ code with Python?

There may indeed be reasons to want to move off of setuptools that are unrelated to distutils, indeed. :)

Aside: I really like flit for this. Because flit-core has zero dependencies, or one if you count a vendored tomli on versions of python less than 3.11 -- hatchling has 5, and 7 on python 3.7, and a couple of those need setuptools to build as well.

ofek:
> Aside: I really like flit for this. Because flit-core has zero dependencies, or one if you count a vendored tomli on versions of python less than 3.11 -- hatchling has 5, and 7 on python 3.7, and a couple of those need setuptools to build as well.

As far as I understand the only context in which there is a benefit to a build backend having zero dependencies is for projects that everyone depends on like `packaging` and in that case really only for distributions like Debian to avoid cycles when building everything from scratch.

eli-schwartz:
Downloading fewer things into an isolated venv when building sdist -> wheel is nice, but perhaps most people don't care about that.

ofek:
> I think what hatchling is doing there with dependencies has more to do with its non-pure-python support via plugins than it has to do with anything a pure-python project cares about?

I have heard that before but actually no https://github.com/pypa/hatch/blob/hatchling-v1.12.2/backend/pyproject.toml#L37-L44

On Python 3.8 there are 5 dependencies and on the latest Python that includes reading TOML there are only 4:

- `editables` - this is for editable installations that want to use the opt-in targeted behavior (the default is simply PTH files)
- `packaging` - this is for utilizing standards for project name normalization, dependency parsing, etc.
- `pathspec` - this is the library that provides Git-style pattern matching for file inclusion/exclusion
- `pluggy` - this is the only dependency for the plugin system

eli-schwartz:
Somehow flit manages the same functionality without needing any of this? But granted, I did not check what you're using packaging for. I've now checked -- you, as expected, use it to manage non-pure tags via `packaging.tags`, but also for some bits I don't fully understand, like `python -m hatchling dep synced` which has no documentation, but seems to have something to do with environments, and as a CLI tool, definitely nothing to do with building an sdist or wheel?

Maybe it makes sense to do this stuff with packaging, once it's already a mandatory backend dependency due to needing tags support.

So, let us say 2 out of 4 are for non-pure python -- and a third for file inclusion/exclusion that this PR doesn't implement anyway.

...

Regardless, it's certainly 5 dependencies in most places (4 on py3.11) just like I said, whatever the reasons for them. Obviously, adding dependencies also allows adding features, which some people want and/or need, so there's definite value there.

But, like I said, the flip side is that building in an isolated environment requires downloading more artifacts from PyPI, and unpacking more files to disk, which is more work (and also on some OSes, can be relatively slow).

Thus, my point stands:
- flit is super simple -- "if you don't need any of that power, why bother?"
- @eli-schwartz happens to like flit for cases like this

ofek:
> and a third for file inclusion/exclusion that this PR doesn't implement anyway

Every project uses that functionality by default because Hatchling respects `.gitignore` unlike other backends, which is one of the ways in which the defaults are easier to reason about and less surprising for users.

eli-schwartz:
> Every project uses that functionality by default because Hatchling respects `.gitignore` unlike other backends, which is one of the ways in which the defaults are easier to reason about and less surprising for users.

hatchling should not do this, it is fundamentally the incorrect way to interact with gitignore files inside a git repo. Use `git check-ignore` for this.

Of course it's still surprising behavior -- gitignore files don't affect files checked into git, so regardless of what you choose, some hatchling users will be surprised if gitignore is respected to remove files tracked by git (I assume hatch does this today), or else other users will be surprised that it isn't respected (`git check-ignore`).

Personally, I am on the side that finds this default highly confusing, difficult to reason about, and surprising, so I would prefer `git check-ignore`.

You cannot change to the technologically correct approach, because that would be a breaking change.

You could use `git check-ignore --no-index`, but you still need the library to handle manually specified git-style patterns, because changing that would *also* be a breaking change. Even though there's no particular advantage to that, except sharing the internal implementation.

Meanwhile, git has functionality which is totally unrelated to gitignore that is intended precisely for this use case ("when creating distribution archives, exclude certain tracked files") that is used by git's native "create distribution archives" functionality -- Meson, for example, uses this to prep an sdist before running sdist hooks. It incurs no additional dependencies, because it just uses git (or mercurial). This makes the common case easy, while making the uncommon case (taking artifacts built by `build_wheel()`, building them during an sdist as well, and then including them inside an sdist) more complicated.

Nearly all complexity falls under the "not pure python" category anyway, so I suspect flit and setuptools_scm's "if it is in git, it is distributed in the sdist" default is actually what people would really find simple, and you don't need a dependency for that because you can just run git ls-files. You only really want that power when you're doing nontrivial codegen.

...

This is exactly the opposite of simple, which is fine if you want a non-simple backend, for which there are legitimate reasons.

Saransh-cpp:
I don't have any strong opinions here. I'll maybe leave it for @lululxvi to decide. I think all of the backends should be fine for DeepXDE, and would only require a few minor changes in `pyproject.toml`. We can maybe start with `setuptools` as DeepXDE currently runs on it and then iterate further if required.

ofek:
> hatchling should not do this, it is fundamentally the incorrect way to interact with gitignore files inside a git repo

I do understand your perspective! However, similar to what Henry was talking about in [that other thread](https://github.com/pyOpenSci/python-package-guide/pull/23#discussion_r1087430451) using the ignore file actually leads to more consistent behavior because building from an actual Git checkout is in fact the least common scenario. If one is building from the source distribution for example by using pip or distribution configuration for a project as is common with Conda, Nix, etc. there is often no Git functionality, only the included ignore file.

eli-schwartz:
Yeah sorry, my bad.

Any further discussion of hatchling design decisions or whether it is or isn't a simple backend are probably best discussed on the hatchling issue tracker -- if @ofek is interested in my opinions then feel free to open a discussion there and ping me into it. It seems to have gone quite beyond the point of offtopic here.

Saransh-cpp:
> that other thread

(Very informative thread. Thanks!)

pyproject.toml
```diff
@@ -0,0 +1,57 @@
+[build-system]
+requires = ["setuptools>=65"]
+build-backend = "setuptools.build_meta"
```
Saransh-cpp, line 3: This PR now shifts the metadata to `pyproject.toml` while keeping the build backend same - `setuptools`.

ofek: Do you have any recommendations for improvements to Hatchling?

Saransh-cpp: Hatchling is amazing! We decided to make the migration incremental -

1. Use this PR to move the metadata to `pyproject.toml` while keeping the backend the same.
2. Check if `setuptools` works okay and migrate to `hatchling` if we see any drawbacks.

pyproject.toml
```diff
@@ -0,0 +1,57 @@
...
+    "scikit-learn",
+    "scikit-optimize>=0.9.0",
+    "scipy",
+]
+
+[project.urls]
+Homepage = "https://github.com/lululxvi/deepxde"
+"Bug Tracker" = "https://github.com/lululxvi/deepxde/issues"
+Changelog = "https://github.com/lululxvi/deepxde/releases"
+Documentation = "https://deepxde.readthedocs.io/en/latest/"
```
lululxvi, line 50: Documentation = "https://deepxde.readthedocs.io"

PR #1132, opened 2023-02-05
Add 16 bit float support to deepXDE

g-w1:


deepxde/config.py
```diff
@@ -37,14 +37,19 @@ def set_default_float(value):
...
+        print("Set the default float type to float16")
+        real.set_float16()
+    elif value == "float32":
         print("Set the default float type to float32")
         real.set_float32()
     elif value == "float64":
         print("Set the default float type to float64")
         real.set_float64()
+    else:
+        raise (f"ValueError: {value} not supported in deepXDE")
```
lululxvi, line 52: Delete `(` and `)`

PR #1137, opened 2023-02-08
A copy button added in the docs.

praksharma:
Hi, I just added a copy button in the docs to copy the code with just one click. This is particularly useful when copying the full-code at the end of the examples.

Here is an example: 

image

This copybutton appear in each code cell. I have build the docs so you can have a quick look.

HTML build: https://deepxde-copybutton.readthedocs.io/en/latest/demos/pinn_forward/poisson.1d.dirichlet.html

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -3,8 +3,6 @@ numpy
 scikit-learn
 scikit-optimize>=0.9.0
 scipy
-docutils<0.18  # https://github.com/readthedocs/readthedocs.org/issues/8616
-sphinx-copybutton
 
 # TensorFlow 1.x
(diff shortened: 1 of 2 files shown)


docs/requirements.txt
```diff
@@ -17,3 +17,4 @@ paddlepaddle>=2.3.0
 jax
 flax
 optax
+sphinx-copybutton
```
lululxvi, line 20: Reorder as
```
optax

docutils<0.18 # https://github.com/readthedocs/readthedocs.org/issues/8616
sphinx-copybutton
```

praksharma: done

PR #1140, opened 2023-02-14
optimize spase matrix initialization for FPDE

HydrogenSulfate:
Cache `int_mat` when FPDE.alpha is an unlearnable constant to speed up training process, such as `fractional_Poisson_3d.py`

### DIFF
--- a/deepxde/data/fpde.py
+++ b/deepxde/data/fpde.py
@@ -83,6 +83,7 @@ class FPDE(PDE):
         self.alpha = alpha
         self.disc = Scheme(meshtype, resolution)
         self.frac_train, self.frac_test = None, None
+        self.int_mat = None
 
         super().__init__(
             geometry,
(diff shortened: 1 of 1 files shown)


deepxde/data/fpde.py
```diff
@@ -83,6 +83,7 @@ def __init__(
         self.alpha = alpha
         self.disc = Scheme(meshtype, resolution)
         self.frac_train, self.frac_test = None, None
+        self.int_mat = None
```
lululxvi, line 86: Use `self.int_mat_train`

HydrogenSulfate: Done

deepxde/data/fpde.py
```diff
@@ -98,7 +99,18 @@ def __init__(
 
     def losses_train(self, targets, outputs, loss_fn, inputs, model, aux=None):
         bcs_start = np.cumsum([0] + self.num_bcs)
-        int_mat = self.get_int_matrix(True)
+        # do not cache int_mat when alpha is a learnable parameter
+        if not is_tensor(self.alpha):
```
lululxvi, line 103:
```
if is_tensor(self.alpha):
    ...
else
    ...
```

HydrogenSulfate: Done

PR #1149, opened 2023-02-20
speed up construction of indices for sparse tensor

HydrogenSulfate:
replace list(zip()) with nested list comprehension to speed up construction of indices for sparse tensor

python3.7 

```python
import time
import random

n = 1252319
array = [[random.randint(0, n), random.randint(0, n)] for _ in range(n)]

s = time.perf_counter()
res1 = zip(*array)
e1 = time.perf_counter()
print(f"zip(...) cost: {e1 - s}")

s = time.perf_counter()
res1 = list(res1)
e2 = time.perf_counter()
print(f"list(...) cost: {e2 - s}")

s = time.perf_counter()
res2 = [tuple([p[0] for p in array]), tuple([p[1] for p in array])]
e2 = time.perf_counter()
print(f"list comprehension cost: {e2 - s}")

print(res1 == res2)

"""
zip(...) cost: 0.6884705810807645
list(...) cost: 0.0964503069408238
list comprehension cost: 0.1501797849778086
True
"""
```

### DIFF
--- a/examples/pinn_forward/fractional_Poisson_1d.py
+++ b/examples/pinn_forward/fractional_Poisson_1d.py
@@ -32,7 +32,12 @@ def fpde(x, y, int_mat):
 #     """(D_{0+}^alpha + D_{1-}^alpha) u(x) = f(x)"""
 #     if isinstance(int_mat, (list, tuple)) and len(int_mat) == 3:
 #         indices, values, shape = int_mat
-#         int_mat = paddle.sparse.sparse_coo_tensor(list(zip(*indices)), values, shape, stop_gradient=False)
+#         int_mat = paddle.sparse.sparse_coo_tensor(
+#             [[p[0] for p in indices], [p[1] for p in indices]],
+#             values,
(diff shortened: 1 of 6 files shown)


lululxvi:
How about new versions, such as Python 3.11?

HydrogenSulfate:
> How about new versions, such as Python 3.11?

zip still slower than list comprehension
```
zip(...) cost: 0.6185596429277211
list(...) cost: 0.08116732793860137
list comprehension cost: 0.12146345409564674
True
```

lululxvi:
Personally, I like `list(zip(*indices))` as it is more elegant, but of course, speed is very important. The difference is about 0.6s. How does this affect the total cost of training?

HydrogenSulfate:
> Personally, I like `list(zip(*indices))` as it is more elegant, but of course, speed is very important. The difference is about 0.6s. How does this affect the total cost of training?

52876s vs 2888s，with display_every=1

PR #1150, opened 2023-02-20
Sphinx-copybutton : Strip bash input prompt and some error handling

praksharma:
Hi. I just realised that when I added a copybutton in the docs (#1137), it will copy the bash prompt in [installation.html](https://deepxde.readthedocs.io/en/latest/user/installation.html#installation) 

For example. this will be copied `$ pip install deepxde`, if we use the copybutton. So, I added a fix to the conf.py. Now it will strip the bash prompt and will copy `pip install deepxde` .


praksharma:
I've also added checks so the user can't just pass anything to create geometry. They will get meaningful errors. I've also add dimension checks.

deepxde/geometry/geometry_2d.py
```diff
@@ -12,6 +12,14 @@
 
 class Disk(Geometry):
     def __init__(self, center, radius):
+        if not isinstance(center, list):
```
lululxvi, line 15: `center` can also be numpy array or tuple.
The same for many other cases.

praksharma: well there are some [problems](https://stackoverflow.com/questions/43081809/len-of-a-numpy-array-in-python/43082119#43082119) but this won't happen unless the user is very careless. Btw I've removed all the explicit checks for  `list`.

deepxde/geometry/geometry_nd.py
```diff
@@ -10,7 +10,7 @@
 
 
 class Hypercube(Geometry):
-    def __init__(self, xmin, xmax):
+    def __init__(self, xmin, xmax):
```
lululxvi, line 13: Reformat here.

docs/conf.py
```diff
@@ -178,3 +178,5 @@ def get_version(rel_path: str) -> str:
 
 
 # -- Extension configuration -------------------------------------------------
+# Sphinx-copybutton : Strip bash input prompt from copied text, if exist
+copybutton_prompt_text = "$ "
```
lululxvi, line 182: Add an empty line at the end.

deepxde/geometry/geometry_2d.py
```diff
@@ -12,6 +12,12 @@
 
 class Disk(Geometry):
     def __init__(self, center, radius):
+        if not isinstance(radius, (int, float)):
+            raise TypeError("radius must be an integer or float")
+
+        if len(center) != 2:
+            raise ValueError("Dimension of center should be 2.")
```
lululxvi, line 19: Length of center

deepxde/geometry/geometry_2d.py
```diff
@@ -78,6 +84,9 @@ class Rectangle(Hypercube):
 
     def __init__(self, xmin, xmax):
         super().__init__(xmin, xmax)
+        if not all([len(xmin) == 2,len(xmax) == 2]):
```
lululxvi, line 87: Format via black.

deepxde/geometry/geometry_2d.py
praksharma: done

docs/conf.py
praksharma: done

deepxde/geometry/geometry_2d.py
praksharma: I used black 22.6.0, conda doesn't install the latest 23.1.0. It changes some other parts of the code.

deepxde/geometry/geometry_nd.py
praksharma: done

deepxde/geometry/geometry_nd.py
```diff
@@ -164,4 +167,4 @@ def background_points(self, x, dirn, dist2npt, shift):
         n = max(dist2npt(dx), 1)
         h = dx / n
         pts = x - np.arange(-shift, n - shift + 1, dtype=config.real(np))[:, None] * h * dirn
-        return pts
+        return pts
```
lululxvi, line 170: This line shouldn't change.

deepxde/geometry/geometry_2d.py
```diff
@@ -78,6 +84,9 @@ class Rectangle(Hypercube):
 
     def __init__(self, xmin, xmax):
         super().__init__(xmin, xmax)
+        if not all([len(xmin) == 2, len(xmax) == 2]):
```
lululxvi, line 87: `if len(xmin) != 2 or len(xmax) != 2:`

deepxde/geometry/geometry_2d.py
```diff
@@ -161,7 +170,14 @@ class Triangle(Geometry):
     """
 
     def __init__(self, x1, x2, x3):
+        if not all([len(x1) == 2, len(x2) == 2, len(x3) == 2]):
+            raise ValueError("Dimension of x1, x2 and x3 should be 2.")
+
         self.area = polygon_signed_area([x1, x2, x3])
+        if self.area == 0:
```
lululxvi, line 177: You cannot compare floating number by "==". Use `isclose`.

deepxde/geometry/geometry_3d.py
```diff
@@ -15,6 +15,8 @@ class Cuboid(Hypercube):
 
     def __init__(self, xmin, xmax):
         super().__init__(xmin, xmax)
+        if not all([len(xmin) == 3,len(xmax) == 3]):
```
lululxvi, line 18: `if len(xmin) != 3 or len(xmax) != 3:`

PR #1158, opened 2023-02-22
add jax support for Lotka Volterra example

HGangloff:


examples/pinn_forward/Lotka_Volterra.py
```diff
@@ -1,128 +1,159 @@
...
+# def input_transform(t):
+#     return paddle.concat(
+#         (
+#             paddle.sin(t),
+#         ),
+#         axis=1,
+#     )
+# Backend jax
+# def input_transform(t):
+#     if t.ndim == 1:
```
lululxvi, line 94: Why do we need `if t.ndim == 1:`?

examples/pinn_forward/Lotka_Volterra.py
```diff
@@ -1,128 +1,159 @@
...
+#     y2 = y[:, 1:2]
+#     return torch.cat([y1 * torch.tanh(t) + 100 / ub, y2 * torch.tanh(t) + 15 / ub], dim=1)
+# Backend paddle
+# def output_transform(t, y):
+#     y1 = y[:, 0:1]
+#     y2 = y[:, 1:2]
+#     return paddle.concat([y1 * paddle.tanh(t) + 100 / ub, y2 * paddle.tanh(t) + 15 / ub], axis=1)
+# Backend jax
+# def output_transform(t, y):
+#     if y.ndim == 1:
```
lululxvi, line 122: Why `if y.ndim == 1:`?

HGangloff:
We need those tests because before vmapping the input for jax function has no batch dimension. Note that in `output_transform`  the condition is not necessary but the `.squeeze()` remains.

lululxvi:
In the code, `t.ndim == 1` should be either always `True`, or always `False`. You can keep `t = t[None]` if needed, but `if t.ndim == 1:` should be removed.

HGangloff:
No it varies, I admit that I still don't get why... something in the Jax backend complicates the use of input_transform and output_transform

PR #1142, opened 2023-02-14
Add a workflow to automate the release process

Saransh-cpp:
XRef #1111 

Now, creating a new release on GitHub will trigger GitHub Actions to automatically build DeeoXDE and publish it to PyPI, without any manual work. One will also be able to trigger the release process manually from the Actions tab (`workflow_dispatch`).

This will require adding a new PyPI API token as a repository secret in DeepXDE's GitHub repository. This secret should be named `PYPI_API_TOKEN`.

PR #1176, opened 2023-03-08
Use setuptools_scm for VCS versioning

Saransh-cpp:
Xref: #1111

[setuptools_scm](https://github.com/pypa/setuptools_scm) provides a neat way of versioning a library through the available `git` tags. This should ideally be merged after #1142. This is how the release process will look like after both of these PRs are merged (most of these steps will occur simultaneously) -
1. A new release is drafted and published on GitHub through the releases page
2. This release will automatically create a new tag in the git repository (as specified while making the release)
3. This release will also trigger the release workflow added in #1142, which will prepare and upload the `sdist` and `wheel` on PyPI
4. During the release `setuptools_scm` will automatically pull up the version from the git tag created in step 2 and add it as DeepXDE's version

This PR eliminates the need of updating the version number manually before every release. It also goes hand-in-hand with #1142, aiming to automate the release process completely.

> **Warning**
> We will have to add `setuptools_scm` as a dependency in DeepXDE's conda feedstock recipe.


docs/conf.py
```diff
@@ -14,24 +14,10 @@
...
-
-def get_version(rel_path: str) -> str:
-    for line in read(rel_path).splitlines():
-        if line.startswith("__version__"):
-            # __version__ = "0.11.2"
-            delim = '"' if '"' in line else "'"
-            return line.split(delim)[1]
-    raise RuntimeError("Unable to find version string.")
-
+dde_version = version('deepxde')
```
Saransh-cpp, line 20: Revert this and split in another PR.

.readthedocs.yaml
```diff
@@ -19,3 +19,5 @@ sphinx:
 python:
    install:
    - requirements: docs/requirements.txt
+   - method: pip
+     path: .
```
Saransh-cpp, line 23: This is now required as `_version.py` is generated only when `DeepXDE` is installed. Previously, the version string could be accessed without having `DeepXDE` installed, by just reading the `__about__.py` file.

Saransh-cpp:
I will mark this PR for review once the doc build is fixed, to ensure that the build passes on this PR too.

Saransh-cpp:
Works! One thing that I forgot to mention in my meet with Prof. Lu -

This will require updating DeepXDE's `conda` feedstock to include `setuptools_scm` in `recipe/meta.yaml` as a `host` requirement. Something like this - https://github.com/conda-forge/zarr-feedstock/blob/cfc7f71b6178581a6144a647a85ac2c37b962310/recipe/meta.yaml#L26-L31

lululxvi:
How does setuptools_scm know the current version? from the tag?

Saransh-cpp:
Yes, it uses `git` tags to pull the version information. More precisely, quoting their docs -

Default versioning scheme
---------------------------

In the standard configuration ``setuptools_scm`` takes a look at three things:

1. latest tag (with a version number)
2. the distance to this tag (e.g. number of revisions since latest tag)
3. workdir state (e.g. uncommitted changes since latest tag)

and uses roughly the following logic to render the version:

no distance and clean:
    ``{tag}``
distance and clean:
    ``{next_version}.dev{distance}+{scm letter}{revision hash}``
no distance and not clean:
    ``{tag}+dYYYYMMDD``
distance and not clean:
    ``{next_version}.dev{distance}+{scm letter}{revision hash}.dYYYYMMDD``

The next version is calculated by adding ``1`` to the last numeric component of the tag.

For Git projects, the version relies on `git describe <https://git-scm.com/docs/git-describe>`_, so you will see an additional ``g`` prepended to the ``{revision hash}``.


Semantic Versioning (SemVer)
------------------------------

Due to the default behavior it's necessary to always include a patch version (the ``3`` in ``1.2.3``), or else the automatic guessing will increment the wrong part of the SemVer (e.g. tag ``2.0`` results in ``2.1.devX`` instead of ``2.0.1.devX``). So please make sure to tag accordingly.

> **Note**
> Future versions of ``setuptools_scm`` will switch to `SemVer
> <http://semver.org/>`_ by default hiding the the old behavior as a configurable option.

---

> What if in the future we switch from setuptools to something else like hatch? Do we need to then change setuptools_scm?

There are similar plugin-like libraries for other build backends as well. For instance, `hatch` has [`hatch-vcs`](https://github.com/ofek/hatch-vcs) which works identically.

lululxvi:
This is ready to merge, right? How can we test this?

PR #1178, opened 2023-03-09
model: add loss scaling optimizer support for tensorflow

g-w1:


lululxvi:
Does this require updates?

g-w1:
I don't think so.

lululxvi:
Is this PR still needed?

g-w1:
I don't think so.

PR #1193, opened 2023-03-17
Update model.py

OverKit:
1. Remove the code that changes the property `requires_grad` of the `net`.
2. Let `trainable_variables` in pytorch more resonable, considering that the user's code may change `requires_grad` and recompile.

### DIFF
--- a/deepxde/model.py
+++ b/deepxde/model.py
@@ -309,7 +309,7 @@ class Model:
         # https://pytorch.org/docs/stable/optim.html#per-parameter-options,
         # but not all optimizers (such as L-BFGS) support this.
         trainable_variables = (
-            list(self.net.parameters()) + self.external_trainable_variables
+            list(filter(lambda p: p.requires_grad, self.net.parameters())) + self.external_trainable_variables
         )
         if self.net.regularizer is None:
             self.opt, self.lr_scheduler = optimizers.get(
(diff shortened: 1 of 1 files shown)


deepxde/model.py
```diff
@@ -518,9 +518,7 @@ def _outputs_losses(self, training, inputs, targets, auxiliary_vars):
             outs = outputs_losses(inputs, targets, auxiliary_vars)
         elif backend_name == "pytorch":
             # TODO: auxiliary_vars
-            self.net.requires_grad_(requires_grad=False)
```
lululxvi, line 521: Why remove this? If not training, then this speeds up the prediction.

OverKit: For a custom class, where some layers are frozen inside, this will make frozen layers trainable. For the acceleration of prediction, I will carefully reconsider whether it's worth removing the code.

lululxvi: Why this will make frozen layers trainable?

OverKit: In `nn.pytorch.fnn`, all parameters are collected from `self.linears = torch.nn.ModuleList()`. `net.requires_grad_` let these parameters all trainable or untrainable. If I want one layer to freeze in my custom class during training, I have to keep that layer from being collected by paramters(), so that it wont'be affected by `net.requires_grad_`. One way I've come up with is to put this layer into a plain list from `self.linears`.

lululxvi: Sounds good.

PR #1196, opened 2023-03-20
Backend Paddle change line_search_fn config of LBFGS optimizer

lijialin03:


deepxde/optimizers/paddle/optimizers.py
```diff
@@ -35,7 +35,7 @@ def get(params, optimizer, learning_rate=None, decay=None):
             tolerance_grad=LBFGS_options["gtol"],
             tolerance_change=LBFGS_options["ftol"],
             history_size=LBFGS_options["maxcor"],
-            line_search_fn='strong_wolfe',
+            line_search_fn=None,
```
lululxvi, line 38: If set as `None`, then what line search method will be used?

lijialin03: It will use a fixed-length step which is a default way. Torch also use this way now.

PR #1094, opened 2022-12-23
Add DistributedDataParallel training with paddle

HydrogenSulfate:
Add DistributedDataParallel training for 5 model with paddle：`Poisson_Dirichlet_1d`、`diffusion_reaction_rate`、`Poisson_Dirichlet_1d_exactBC`、`Poisson_multiscale_1d`、`wave_1d`

Main logic is depicted below
deepxde多卡支持示意图

update:
1. raise Error when length of data can't be divided by world_size, error is as below.
    ```python
      File "deepxde_sd/deepxde/data/pde.py", line 259, in train_points
        X = array_ops_compat.split_in_rank(X)
      File "deepxde/utils/array_ops_compat.py", line 57, in split_in_rank
        f"The data length({n_total}) must be an "
    ValueError: The data length(41) must be an integer multiple of world_size(2).
    ```

result(examples/pinn_forward/Poisson_multiscale_1d.py, seed=42)
|1GPU|2GPU|
|--- |--- |
| image | image|

PR #1205, opened 2023-03-25
(Weak) data-parallel Horovod acceleration for tensorflow.compat.v1

pescap:
This is a first version. 
- Check that data distribution is ok
- See whether I change the structure of `data_parallel` variable
- Avoid multiple `hvd` imports: I am importing it for each function. Is there a way to import once `hvd`?
- `hvd` does not work for `tf.compat.v1` in eager mode. Is it an issue @lululxvi ?


deepxde/backend/tensorflow_compat_v1/tensor.py
```diff
@@ -23,7 +23,8 @@
 # features in TensorFlow 2.x are useful such as `Tensor shapes`. Actually we use `Tensor
 # shapes` in DeepXDE.
 tf.disable_v2_behavior()
-tf.enable_v2_tensorshape()
+#tf.enable_v2_tensorshape()
```
pescap, line 26: add it

deepxde/config.py
```diff
@@ -16,12 +16,13 @@
     jax_random_seed = random.randint(iinfo.min, iinfo.max)
 # XLA
 xla_jit = False
-if backend_name in ["tensorflow.compat.v1", "tensorflow"]:
-    xla_jit = bkd.is_gpu_available()
-elif backend_name == "jax":
-    xla_jit = True
-if xla_jit:
-    print("Enable just-in-time compilation with XLA.\n")
+# Remember
```
pescap, line 19: Todo: Check if the `xla_jit` and `hvd` run together.

lululxvi:
> * `hvd` does not work for `tf.compat.v1` in eager mode. Is it an issue @lululxvi ?

Tf v1 is not eager mode, right?

deepxde/config.py
```diff
@@ -33,9 +33,7 @@ def default_float():
 
 def set_default_float(value):
     """Sets the default float type.
-
```
lululxvi, line 36: Why delete these lines? Keep them.

docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM horovod/horovod:master
+FROM horovod/horovod:sha-9b42eda
```
lululxvi, line 1: This has already been merged in another PR.

docker/requirements.txt
```diff
@@ -3,6 +3,7 @@ scipy
 sklearn
 matplotlib
 scikit-optimize
+seaborn
```
lululxvi, line 6: Also this file.

deepxde/model.py
```diff
@@ -19,7 +19,6 @@
 
 class Model:
     """A ``Model`` trains a ``NN`` on a ``Data``.
-
```
lululxvi, line 22: Also keep these lines.

pescap: black added the lines. Ok, let me keep these lines.

deepxde/model.py
```diff
@@ -64,6 +64,7 @@ def compile(
         decay=None,
         loss_weights=None,
         external_trainable_variables=None,
+        data_parallel=False,
```
pescap, line 67: Codacy suggests that I `import horovod` at the beginning of `model.py`. I don't know which would be the best manner to do this, as I just want `hvd` to be imported for `model(data_parallel=True)`

deepxde/model.py
```diff
@@ -142,7 +157,9 @@ def compile(
         metrics = metrics or []
         self.metrics = [metrics_module.get(m) for m in metrics]
 
-    def _compile_tensorflow_compat_v1(self, lr, loss_fn, decay, loss_weights):
+    def _compile_tensorflow_compat_v1(
+        self, lr, loss_fn, decay, loss_weights, data_parallel
```
lululxvi, line 161: Do we need `data_parallel` as the argument? We have `self.data_parallel`.

deepxde/model.py
lululxvi: Yes, we only import hvd if it is parallel.

deepxde/model.py
```diff
@@ -109,8 +110,20 @@ def compile(
                 physics systems that need to be recovered. If the backend is
                 tensorflow.compat.v1, `external_trainable_variables` is ignored, and all
                 trainable ``dde.Variable`` objects are automatically collected.
+            data_parallel: If True, a Horovod-based data-parallel acceleration is performed.
```
lululxvi, line 113: Instead of using this argument, can we automatically detect if there are more than 1 GPU? If so, then we use hvd?

deepxde/model.py
pescap: you're right, sorry

deepxde/model.py
pescap: - One can get `CUDA_VISIBLE_DEVICES`. Yet, I wouldn't use Horovod by default is there are more than 1 GPUs available. I was thinking about adding a `dde.config.set_hvd()` function, that would return `hvd`.
- Horovod supports multiple node. This PR would handle multiple node with no additional change. The `horovodrun` from command line has to be adapted (check https://horovod.readthedocs.io/en/stable/running_include.html). However, I never used several nodes so far.

pescap:
I had defined in `config.py`:
```
hvd = None

 def set_hvd():
    """Sets horovod.

    - Only supported for backend Tensorflow 1.x.
    """
        import horovod.tensorflow as hvd
    else:
        raise NotImplementedError(
                    "The data parallel acceleration is only implemented in backend tensorflow.compat.v1"
                )
    global hvd
    hvd = hvd
```

Then, in model:

```from .config import hvd```

And in the main script, I was running:

```
dde.config.set_hvd()
```

However, it does not work for `model.py` (gives me `None`). Any help?

**Comment by pescap** [comment]
At least the actual version is working fine.

**Comment by pescap** [comment]
I shall add more details concerning Horovod acceleration. I found it optimal to keep the `data_parallel` parameter, as it allows to have more usages of Horovod acceleration:

If we run:
```
horovodrun -np 2 -H localhost:2 python Poisson_Lshape.py
```
with `data_parallel=True`, it performs the data-parallel acceleration over two GPUs, with allreduce operation.

Else, if we run the same command with `data_parallel=False`, it runs the model separately over each rank. This can be useful for model performance analysis over several concurrent runs.

Therefore, on can:
-> Run with different `data` points over each rank to compar points distribution and trainings
-> Run with the same `data` points over each rank to compare trainings

File: deepxde/config.py [hunk 10]
```diff
@@ -7,6 +7,15 @@
 from .backend import backend_name, tf, torch, paddle
 from .real import Real
 
+# Import Horovod if a mpirun process is ongoing
+global hvd
```
lululxvi, line 11: Do you need `global`?

pescap: I did not need it. Edited the code, hvd is now imported once if a `mpirun`-type process is ongoing.

deepxde/config.py
```diff
@@ -16,14 +24,13 @@
...
 xla_jit = False
-if backend_name in ["tensorflow.compat.v1", "tensorflow"]:
+if backend_name in ["tensorflow.compat.v1", "tensorflow"] and hvd is None:
     xla_jit = bkd.is_gpu_available()
 elif backend_name == "jax":
     xla_jit = True
 if xla_jit:
     print("Enable just-in-time compilation with XLA.\n")
 
-
```
lululxvi, line 26: Keep this line.

deepxde/config.py
```diff
@@ -7,6 +7,14 @@
 from .backend import backend_name, tf, torch, paddle
 from .real import Real
 
+# Import Horovod if a mpirun process is ongoing
+try:
+    mpi = os.environ['OMPI_COMM_WORLD_SIZE']
+    import horovod.tensorflow as hvd
+    tf.compat.v1.disable_eager_execution()
```
lululxvi, line 14: You can use `hvd.init()` here.

deepxde/config.py
```diff
@@ -7,6 +7,14 @@
 from .backend import backend_name, tf, torch, paddle
 from .real import Real
 
+# Import Horovod if a mpirun process is ongoing
+try:
+    mpi = os.environ['OMPI_COMM_WORLD_SIZE']
```
lululxvi, line 12: As currently hvd only works for tensorflow 1.x, check if the backend is tensorflow 1.x

deepxde/config.py
lululxvi, line 14: `tf.compat.v1.disable_eager_execution()` doesn't seem necessary, as it is already disabled in

https://github.com/lululxvi/deepxde/blob/c2e05ac8e97b89ab3247280678093dfb546faa14/deepxde/backend/tensorflow_compat_v1/tensor.py#L25

deepxde/model.py
```diff
@@ -625,12 +652,15 @@ def train(
...
-        display.training_display.summary(self.train_state)
+        if not self.data_parallel or (self.data_parallel and hvd.rank() == 0):
+            display.training_display.summary(self.train_state)
         if model_save_path is not None:
             self.save(model_save_path, verbose=1)
         return self.losshistory, self.train_state
 
     def _train_sgd(self, iterations, display_every):
+        if self.data_parallel:
+            import horovod.tensorflow as hvd
```
lululxvi, line 663: Why do we need to import hvd? Use config.hvd?

lululxvi:
For data parallel, different GPU uses different data, but here, which code ganrantee this? How the data is split into different gpus?

deepxde/config.py
pescap: One has to add both lines, or the code doesn't run.

deepxde/model.py
pescap: thank you! Removed

deepxde/config.py
pescap: > For data parallel, different GPU uses different data, but here, which code ganrantee this? How the data is split into different gpus?

As long as we work with with rank-dependent seeds, random data generation gives different data for different GPUs.
This directly leads to different random observations in:

https://github.com/lululxvi/deepxde/blob/c7371b29c90cf0402c4e3c54d2a50a45dc0a8056/examples/pinn_inverse/Navier_Stokes_inverse.py#L48

and random collocation points in:

https://github.com/lululxvi/deepxde/blob/c7371b29c90cf0402c4e3c54d2a50a45dc0a8056/examples/pinn_inverse/Navier_Stokes_inverse.py#L101

deepxde/config.py
pescap: I have been struggling for a month with an error because of this line: https://github.com/horovod/horovod/discussions/3806

pescap:
> For data parallel, different GPU uses different data, but here, which code ganrantee this? How the data is split into different gpus?

As long as we work with random points, random data generation gives different data for different GPUs.
This directly leads to different random observations in:

https://github.com/lululxvi/deepxde/blob/c7371b29c90cf0402c4e3c54d2a50a45dc0a8056/examples/pinn_inverse/Navier_Stokes_inverse.py#L48

and different random collocation points in:

https://github.com/lululxvi/deepxde/blob/c7371b29c90cf0402c4e3c54d2a50a45dc0a8056/examples/pinn_inverse/Navier_Stokes_inverse.py#L101

Therefore, for data-parallel hvd: use random collocation points generation, and random slices for observations.

This is done without changing the main scripts.

Once this PR is ok, I plan to work on:
- Adding rank-dependent random seed (if horovod is running, and a seed is defined, define a rank dependent seed as for example `seed+rank`.
- Detailing the data-parallel acceleration for Lshape (forward) and NavierStokes (inverse) problems.

deepxde/config.py
```diff
@@ -7,6 +7,20 @@
...
+        import horovod.tensorflow as hvd
+        tf.compat.v1.disable_eager_execution()
+        hvd.init()    
+    else:
+        raise NotImplementedError(
+            "The data parallel acceleration is only implemented in backend tensorflow.compat.v1"
+        )
+except KeyError: 
+    hvd = None
+
```
pescap, line 23: maybe add a else condition, to close this try except.

deepxde/config.py
lululxvi: That is strange, as `tf.disable_v2_behavior()` includes `disable_eager_execution`. Then you can add a comment here to indicate that this line is necessary

lululxvi:
> As long as we work with random points, random data generation gives different data for different GPUs.
> Therefore, for data-parallel hvd: use random collocation points generation, and random slices for observations.

This is more complicated than this. First, it requires that the training data is randomly generated, but we also use many other sampling methods, such as the default Hammersley sequence. But even for random sampling, in some cases, it might work, and in general, it doesn't work.

In Horovod, each worker does the forward pass and backward pass to compute the gradients, and then hvd computes the average of gradients, which is used to update the weights. Due to this "average" operation, it is easy to show that in order to guarantee the parallel version is consistent with non-parallel version, the number of training points for each loss must be the same in all works. Note that PINN loss is the sum of many PDE losses, BC/IC losses, etc. By using random points, there is no guarantee that two works have the same number of training points for each BC. Check the following code

https://github.com/lululxvi/deepxde/blob/c7371b29c90cf0402c4e3c54d2a50a45dc0a8056/deepxde/data/pde.py#L255

You can check the current code at https://github.com/lululxvi/deepxde/pull/1094 for data splitting, although it also has some issues now...

pescap:
I run the Lshape case over 5 GPUs.

Case 1:
`num_boundary=120`.
The ranks are with `129,126,130,127,128` bc points (same order as how the prints show up).
Simulation takes 45s to run 5000 epochs.

Case 2:
`n_boundary = (hvd.rank() + 1) * 120`
The ranks are with `611,369,128,248,488` bc points (same order as how the prints show up).
Simulation takes 45s to run 5000 epochs.

lululxvi:
> Are we sure of this? Having more points for a given rank could delay the allreduce process (by creating a bottleneck for the rank with more points). But in our case, we are dealing with a few points more (or less).
> 
> I am ok to say that it is not optimal mathematically. But I guess that it leads to negligible differences in terms of performance.
> 
> For example, for the Lshape case with `num_boundary=120` and 2 ranks, I get `123` and `127` bc points for each rank, respectively.

This example is exactly what I am trying to say: different ranks have different numbers of training points. You can mathematically show that this is not equivalent to a single rank with 250 (=123+127) points. Well, the difference could be small or negligible in practice. But I prefer it is mathematically good. This is what we try to do in another PR. Once that one is fixed, it can also be used here.

deepxde/config.py
pescap: I added a comment.

deepxde/config.py
```diff
@@ -7,6 +7,20 @@
 from .backend import backend_name, tf, torch, paddle
 from .real import Real
 
+# Import Horovod if a mpirun process is ongoing
+try:
+    mpi = os.environ['OMPI_COMM_WORLD_SIZE']
```
lululxvi, line 12: Use
```
if "OMPI_COMM_WORLD_SIZE" in os.environ:
```
so you don't need `try`

pescap:
As you stated:
- Option 1: BCs out of the data-parallel scheme
- Option 2: Force always that `num_boundary=120` => exactly 120 points on the boundary.
- Option 2: Add a binary parameter to `PDE` to force `num_boundary=120` => exactly 120 points on the boundary.

Example:
https://github.com/lululxvi/deepxde/blob/64f51521e8dc6e272ec979557000d3e9a26b7ba5/examples/pinn_forward/Poisson_Lshape.py#L22

lululxvi:
> If we look for consistency for the entire process, `num_boundary=120` should imply that there are 120 points on the boundary.
> 
> I am not sure whether I am understanding exactly how the bc points generation works. Is the process always leading to equal or more bc training points (e.g. `>=20` points)?

Yes, equal or more BC points could be generated. We can control to always generate equal points. But this is not the key issue. The main issue is that there are more than one BCs. Let us say our geometry is a circle, and we have 2 BCs: BC1 is applied on the left half circle, and BC2 is applied on the right half circle. Now we sample 100 random points on the circle in 2 workers. As the sampling is random, worker 1 may sample 51 points on the left and 49 points on the right, so in this case, BC1 has 51 points, and BC2 has 49 points. But worker 2 may sample 49 points on the left and 51 points on the right, so in this case, BC1 has 49 points, and BC2 has 51 points.

If each worker samples points independently, I think it is very difficult to guarantee the equal number of points for each BC. One solution is that we generate all the points in the root rank, and then split the points, and do a broadcast to other ranks.

pescap:
Update: So far, I didn't manage to broadcast the bc training points using Horovod.

Another option would be to use `mpi4py`, which allows to simply broadcast np arrays: https://pythonprogramming.net/mpi-broadcast-tutorial-mpi4py/.

pescap:
> 



> > If we look for consistency for the entire process, `num_boundary=120` should imply that there are 120 points on the boundary.
> > I am not sure whether I am understanding exactly how the bc points generation works. Is the process always leading to equal or more bc training points (e.g. `>=20` points)?
> 
> Yes, equal or more BC points could be generated. We can control to always generate equal points. But this is not the key issue. The main issue is that there are more than one BCs. Let us say our geometry is a circle, and we have 2 BCs: BC1 is applied on the left half circle, and BC2 is applied on the right half circle. Now we sample 100 random points on the circle in 2 workers. As the sampling is random, worker 1 may sample 51 points on the left and 49 points on the right, so in this case, BC1 has 51 points, and BC2 has 49 points. But worker 2 may sample 49 points on the left and 51 points on the right, so in this case, BC1 has 49 points, and BC2 has 51 points.
> 
> If each worker samples points independently, I think it is very difficult to guarantee the equal number of points for each BC. One solution is that we generate all the points in the root rank, and then split the points, and do a broadcast to other ranks.

Isn't this issue independent of the data-parallel acceleration? If I run the same experiment twice with traditional DeepXDE, the same problem appears (run 1: 51 on the left and 49 on the right; run 2: 49 on the left and 51 on the right).

lululxvi:
> Another option would be to use `mpi4py`, which allows to simply broadcast np arrays: https://pythonprogramming.net/mpi-broadcast-tutorial-mpi4py/.

Yes, mpi4py is a good option.

> Isn't this issue independent of the data-parallel acceleration? If I run the same experiment twice with traditional DeepXDE, the same problem appears (run 1: 51 on the left and 49 on the right; run 2: 49 on the left and 51 on the right). => same consistency issue?

No, non-parallel one is not a problem.

lululxvi:
Sounds great. I think the easiest way is doing broadcast here
https://github.com/lululxvi/deepxde/blob/c7371b29c90cf0402c4e3c54d2a50a45dc0a8056/deepxde/data/pde.py#L171
After all points have been constructed, including PDE points and BC points. Then this also support non-random PDE points, such as grid PDE points. After broadcasting, then do a split, as https://github.com/lululxvi/deepxde/pull/1094

pescap:
#1094 is using a strong scaling approach: Define the total number of points, and split. 
I prefer the weak scaling approach: Set `num_domain` as the number of points for each GPU.

deepxde/data/pde.py
```diff
@@ -160,7 +161,15 @@ def losses(self, targets, outputs, loss_fn, inputs, model, aux=None):
     @run_if_all_none("train_x", "train_y", "train_aux_vars")
     def train_next_batch(self, batch_size=None):
         self.train_x_all = self.train_points()
-        self.train_x = self.bc_points()
+        if hvd is not None:
+            from mpi4py import MPI
+            comm = MPI.COMM_WORLD
+            rank = comm.Get_rank()
+            train_x = self.bc_points()[:self.num_boundary,:]
```
pescap, line 168: I used [:self.num_boundary,:] to obtain broadcast safely the inforation (fixed-size arrays) .

deepxde/data/pde.py
```diff
@@ -160,7 +161,15 @@ def losses(self, targets, outputs, loss_fn, inputs, model, aux=None):
...
+            comm.Bcast(train_x, root=0)
+            self.train_x = train_x
+        else:
+            self.train_x = self.bc_points()[:self.num_boundary,:]
```
pescap, line 172: Keep `[:self.num_boundary,:]` or remove it for the unaccelerated case?

lululxvi:
Actually for python random and numpy random, it seems that the random seeds are the same in different processors.
https://github.com/lululxvi/deepxde/pull/1094#issuecomment-1514536127

deepxde/data/pde.py
lululxvi: Why we need `[:self.num_boundary,:]`?

deepxde/config.py
```diff
@@ -7,6 +7,24 @@
 from .backend import backend_name, tf, torch, paddle
 from .real import Real
 
+# Import Horovod if a mpirun process is ongoing
+hvd = None
+mpi4py = None
```
lululxvi, line 12: `mpi4py` is never used.

deepxde/config.py
```diff
@@ -7,6 +7,24 @@
...
+MPI = None
```
lululxvi, line 13: `comm = None`

deepxde/data/pde.py
pescap: To be sure that `train_x` is of size == `num_boundary` for each rank, so that the broadcast does not throw error.

pescap: I removed it for the unaccelerated case.

deepxde/config.py
pescap: corrected

deepxde/data/pde.py
pescap: I removed it for the unaccelerated case.

deepxde/data/pde.py
```diff
@@ -3,6 +3,7 @@
 from .data import Data
 from .. import backend as bkd
 from .. import config
+from ..config import hvd, comm
```
lululxvi, line 6: Use `config.hvd` and `config.comm`

deepxde/config.py
```diff
@@ -7,6 +7,23 @@
...
+if "OMPI_COMM_WORLD_SIZE" in os.environ:
+    if backend_name == "tensorflow.compat.v1":
+        import horovod.tensorflow as hvd
+        from mpi4py import MPI
+        comm = MPI.COMM_WORLD
+        tf.compat.v1.disable_eager_execution() # Without this line, Horovod broadcasting fails.
+        hvd.init()    
+    else:
+        raise NotImplementedError(
+        "The data parallel acceleration is only implemented in backend tensorflow.compat.v1"
```
lululxvi, line 23: Reformat via black https://github.com/psf/black

deepxde/model.py
```diff
@@ -14,6 +14,7 @@
 from . import utils
 from .backend import backend_name, tf, torch, jax, paddle
 from .callbacks import CallbackList
+from .config import hvd
```
lululxvi, line 17: Use `config.hvd`

deepxde/model.py
```diff
@@ -109,8 +111,11 @@ def compile(
                 physics systems that need to be recovered. If the backend is
                 tensorflow.compat.v1, `external_trainable_variables` is ignored, and all
                 trainable ``dde.Variable`` objects are automatically collected.
+            data_parallel: If True, a Horovod-based data-parallel acceleration is performed.
```
lululxvi, line 114: Why we need `data_parallel`? As we can check `if config.hvd is None`

deepxde/config.py
```diff
@@ -7,6 +7,24 @@
 from .backend import backend_name, tf, torch, paddle
 from .real import Real
 
+# Import Horovod if a mpirun process is ongoing
+hvd = None
+comm = None
+
```
lululxvi, line 13: Delete this empty line

deepxde/model.py
pescap: Great!

deepxde/optimizers/tensorflow_compat_v1/optimizers.py
```diff
@@ -3,14 +3,15 @@
...
+from ...config import hvd
 
 
 def is_external_optimizer(optimizer):
     scipy_opts = ["L-BFGS", "L-BFGS-B"]
     return optimizer in scipy_opts
 
 
-def get(loss, optimizer, learning_rate=None, decay=None):
+def get(loss, optimizer, learning_rate=None, decay=None, data_parallel=False):
```
lululxvi, line 14: No `data_parallel`

deepxde/model.py
```diff
@@ -182,8 +188,10 @@ def losses(losses_fn):
         self.outputs_losses_train = [self.net.outputs, losses_train]
         self.outputs_losses_test = [self.net.outputs, losses_test]
         self.train_step = optimizers.get(
-            total_loss, self.opt_name, learning_rate=lr, decay=decay
-        )
+            total_loss,
+            self.opt_name,
+            learning_rate=lr,
+            decay=decay)
```
lululxvi, line 194: format via black

deepxde/data/pde.py
```diff
@@ -160,7 +160,12 @@ def losses(self, targets, outputs, loss_fn, inputs, model, aux=None):
     @run_if_all_none("train_x", "train_y", "train_aux_vars")
     def train_next_batch(self, batch_size=None):
         self.train_x_all = self.train_points()
-        self.train_x = self.bc_points()
+        if config.hvd is not None:
+            train_x = self.bc_points()[: self.num_boundary, :]
+            config.comm.Bcast(train_x, root=0)
```
lululxvi, line 165: I am a little confused now. As the random seeds for different processors are actually the same, then `train_x` should be the same in different processors, so no Bcase is needed.

pescap: - If no random seed is set, `train_x` differ over different processors (see the figure above) => Bcast is needed
- If one sets a random seed, for example with `deepxde.config.set_random_seed(1234)`, therefore `train_x` are the same over all process. This is why in next step I shall define rank-dependent seeds, to obtain different training points over the ranks. 

Am I right?

lululxvi: That is why I got confused. See comment by another one https://github.com/lululxvi/deepxde/pull/1094#issuecomment-1514536127 and https://stackoverflow.com/questions/23742550/mpi-rand-gives-the-same-constant-numbers-across-all-processes-in-every-run

Now I am not sure if the random state is the same or different in different processors.

pescap: By default, if I set a random state, it will be shared across all ranks.

Check out:
https://github.com/pescap/horovodPINNs/blob/46aa5faee24f9512d2a9e7ae136286c558d9f94b/Laplace/hvd_hval.py#L60

pescap:
> Actually for python random and numpy random, it seems that the random seeds are the same in different processors. [#1094 (comment)](https://github.com/lululxvi/deepxde/pull/1094#issuecomment-1514536127)

It is true. This is why `dde.config.set_random_seed` should be adapted soon, to define manually the seed as for example:
`
seed + hvd.rank() 
`

pescap:
It is working as expected.

Three configurations:
-  No random seed
-  Random seed of 1234
-  Random seed of 1234+rank

In all configurations, the bc points are shared in `data.train_next_batch()[0]`.

final


The code:

```
"""Backend supported: tensorflow.compat.v1, tensorflow, pytorch, jax, paddle"""
import deepxde as dde
import numpy as np
from deepxde.config import hvd


def pde(x, y):
    # Most backends
    # Backend jax
    # dy_xx, _ = dde.grad.hessian(y, x, i=0, j=0, component=0)
    # dy_yy, _ = dde.grad.hessian(y, x, i=1, j=1, component=0)
    return -dy_xx - dy_yy - 1


def boundary(_, on_boundary):
    return on_boundary


seed = '1234+rank'
dde.config.set_random_seed(1234+hvd.rank())

geom = dde.geometry.Polygon([[0, 0], [1, 0], [1, -1], [-1, -1], [-1, 1], [0, 1]])

data = dde.data.PDE(geom, pde, bc, num_domain=200, num_boundary=100, num_test=1500)

to_save = data.train_next_batch()[0]
np.savetxt('data_' + seed +'_' + str(hvd.local_rank()) , to_save)


quit()
net = dde.nn.FNN([2] + [50] * 4 + [1], "tanh", "Glorot uniform")
model = dde.Model(data, net)

model.train(iterations=5000)
```

deepxde/model.py
```diff
@@ -595,19 +601,27 @@ def train(
 
         if backend_name == "tensorflow.compat.v1":
             if self.train_state.step == 0:
-                print("Initializing variables...")
+                if config.hvd is None or (
```
lululxvi, line 604: Now only tensorflow.compat.v1 has this print info. Let us just remove it.

deepxde/config.py
```diff
@@ -7,6 +7,23 @@
 from .backend import backend_name, tf, torch, paddle
 from .real import Real
 
+# Import Horovod if a mpirun process is ongoing
+hvd = None
+comm = None
```
lululxvi, line 12: as `rank` is frequently used, add here
```
rank = 0
```
Then update `rank` below.

deepxde/config.py
```diff
@@ -7,6 +7,23 @@
 from .backend import backend_name, tf, torch, paddle
 from .real import Real
 
+# Import Horovod if a mpirun process is ongoing
```
lululxvi, line 10: Change the comment to
```
# Data parallel via Horovod
```

deepxde/config.py
```diff
@@ -7,6 +7,23 @@
...
+hvd = None
+comm = None
+if "OMPI_COMM_WORLD_SIZE" in os.environ:
+    if backend_name == "tensorflow.compat.v1":
+        import horovod.tensorflow as hvd
+        from mpi4py import MPI
+
+        comm = MPI.COMM_WORLD
+        tf.compat.v1.disable_eager_execution()  # Without this line, Horovod broadcasting fails.
+        hvd.init()
```
lululxvi, line 20: `print(f"\nParallel training with {world_size} processes.\n")`

lululxvi, line 20: Check `if world_size > 1`

deepxde/data/pde.py
```diff
@@ -160,7 +160,12 @@ def losses(self, targets, outputs, loss_fn, inputs, model, aux=None):
     @run_if_all_none("train_x", "train_y", "train_aux_vars")
     def train_next_batch(self, batch_size=None):
         self.train_x_all = self.train_points()
-        self.train_x = self.bc_points()
+        if config.hvd is not None:
+            train_x = self.bc_points()[: self.num_boundary, :]
```
lululxvi, line 164: `self.num_boundary` is the number of points sampled on the boundary, not the number of points of all BCs, see the code of `self.bc_points()`. For example, see this case https://deepxde.readthedocs.io/en/latest/demos/pinn_forward/eulerbeam.html

deepxde/data/pde.py
```diff
@@ -160,7 +160,12 @@ def losses(self, targets, outputs, loss_fn, inputs, model, aux=None):
...
+            config.comm.Bcast(train_x, root=0)
+            self.train_x = train_x
+        else:
+            self.train_x = self.bc_points()
         if self.pde is not None:
             self.train_x = np.vstack((self.train_x, self.train_x_all))
```
lululxvi, line 170: `num_domain` and `num_boundary` are the total number of the PINN, not the number in each process. This is what we usually do in parallel computing --- we use more processor to speed up the same simulation. So, the easiest way is still doing `Bcase` first, then split. To reduce the communication cost, we can use `Scatter`.

deepxde/model.py
pescap: https://github.com/lululxvi/deepxde/blob/b87c1c6499b12c6de6641236c3ea7267340d99db/deepxde/model.py#L598

It was the case so far. Where do you suggest to add this print?

lululxvi: Just delete this print of "Initializing variables...".

deepxde/config.py
pescap: If we run the code with `np 1` i.e. `world_size=1` I think that there is no difference with a unacelerated case. What is this condition for?

deepxde/data/pde.py
pescap: I proposed a solution to this issue: 
- Get the number of rows of `train_x`
- Share this number to other ranks 
- Replace `train_x` with different sizes by zeros, to perform the broadcast from rank `0` to other ranks.

deepxde/data/pde.py
pescap: In Horovod for example I understand that the paradigm is definitely a weak scaling one. This is how benchmarks are shocased. We increase the total batch size by adding more ranks. We know that a certain code runs on 1 GPU, so we try to increase the batch size by scaling up over more GPUs.

pescap: Tensordiffeq and Modulus are with weak scaling (constant number of points per GPU):

https://github.com/tensordiffeq/TensorDiffEq/blob/4e63ddf1bb612bbd94e21ff63cd03f42af8af735/examples/AC-dist-new.py#L51

https://docs.nvidia.com/deeplearning/modulus/text/features/parallel_training.html

deepxde/config.py
lululxvi: If `world_size=1`, then no need to use parallel, no need to import mpi4py, etc.

deepxde/data/pde.py
lululxvi: Yes, it works.

deepxde/model.py
```diff
@@ -21,7 +21,7 @@ class Model:
     """A ``Model`` trains a ``NN`` on a ``Data``.
 
     Args:
-        data: ``deepxde.data.Data`` instance.
+        data: ``deepxftende.data.Data`` instance.
```
lululxvi, line 24: `deepxftende`?

deepxde/data/pde.py
lululxvi: In the purely data-driven case, this makes sense to me, as the goal is to train a network by keeping looping the dataset. But in scientific computing, such as finite elements and molecular dynamics, our goal is to solve a pre-defined problem, we use parallel computing to speed up the computation of the same problem.

Another reason, as we know random sampling is not optimal, it is better to use Hammersley sequence etc, which is a deterministic seq. Then in the current code, all processes have exactly the same data.

deepxde/data/pde.py
```diff
@@ -160,7 +160,18 @@ def losses(self, targets, outputs, loss_fn, inputs, model, aux=None):
...
     def train_next_batch(self, batch_size=None):
         self.train_x_all = self.train_points()
-        self.train_x = self.bc_points()
+        if config.hvd is not None:
+            train_x = self.bc_points()
+            n_train = np.array([train_x.shape[0]])
+            config.comm.Bcast(n_train, root=0)
+            
+            if train_x.shape[0] != n_train[0]:
+                train_x = np.zeros((n_train[0], 2))
```
lululxvi, line 169: `2`? How about 3D?

pescap:
work in progress

deepxde/data/pde.py
pescap: Update:
- As you state, `skopt.sampler.Hammersley` is deterministic. One could explore using `scipy.stats.qmc.Hammersley` with `scramble=True`. 
- However, `skopt.sampler.Sobol` with `randomize=True` in:

https://github.com/lululxvi/deepxde/blob/8a644fe265dc77f693853a76295d140c6b21fccc/deepxde/geometry/sampler.py#L55  
does the job for the weak scaling case (tested).

- I agree with your comments. On the other side, I would argue that sampling all the points on one rank and scattering them is not optimal, and can create memory issues. It is much better to work in a decentralized way, by sampling the points directly on each rank.
- I actually believe that using decentralized (on each rank) and dynamic resampling strategies during training would be a very interesting area to explore. 
- This PR is for weak scaling, while what has been implemented in paddle is for strong scaling. I believe that DeepXDE shall have both methods implemented, with strong scaling as default. The changes between both methods are minor, and can be defined on `pde.py`.

lululxvi: I understand your point: weak scaling would have some advantage of speed compared to strong scaling by reducing communication/memory; strong scaling always works, but might not be optimal.

However, the current code only works with `train_distribution="pseudo"`, not other sampling methods. You need to check the `train_distribution` and raise an error if it is not `pseudo`.

deepxde/config.py
```diff
@@ -7,6 +7,32 @@
 from .backend import backend_name, tf, torch, paddle
 from .real import Real
 
+# Data parallel via Horovod
+hvd = None
+comm = None
+rank = 0
+world_size = 1
```
lululxvi, line 14:
```
world_size = 1
rank = 0
```

deepxde/config.py
```diff
@@ -7,6 +7,32 @@
...
+
+            comm = MPI.COMM_WORLD
+            tf.compat.v1.disable_eager_execution()  # Without this line, Horovod broadcasting fails.
+            rank = hvd.rank() # Only single node acceleration supported so far.
+
+        if rank == 0:
+            print(f"\nParallel training with {world_size} processes.\n")
+    else:
+        raise NotImplementedError(
+            "The data parallel acceleration is only implemented in backend tensorflow.compat.v1"
```
lululxvi, line 32: "Parallel training via Horovod is only implemented in backend tensorflow.compat.v1"

deepxde/model.py
```diff
@@ -153,6 +154,11 @@ def _compile_tensorflow_compat_v1(self, lr, loss_fn, decay, loss_weights):
                     tf.OptimizerOptions.ON_2
                 )
                 self.sess = tf.Session(config=cfg)
+            elif config.hvd is not None:
+                cfg = tf.ConfigProto()
+                cfg.gpu_options.visible_device_list = str(config.rank)
+                cfg.gpu_options.allow_growth = True
```
lululxvi, line 160: Is `cfg.gpu_options.allow_growth = True` necessary? as we have
https://github.com/lululxvi/deepxde/blob/8a644fe265dc77f693853a76295d140c6b21fccc/deepxde/backend/tensorflow_compat_v1/__init__.py#L3

deepxde/config.py
```diff
@@ -16,7 +43,7 @@
     jax_random_seed = random.randint(iinfo.min, iinfo.max)
 # XLA
 xla_jit = False
-if backend_name in ["tensorflow.compat.v1", "tensorflow"]:
+if backend_name in ["tensorflow.compat.v1", "tensorflow"] and hvd is None:
```
lululxvi, line 46: Add a comment why `hvd is None`

deepxde/data/pde.py
```diff
@@ -160,7 +160,24 @@ def losses(self, targets, outputs, loss_fn, inputs, model, aux=None):
     @run_if_all_none("train_x", "train_y", "train_aux_vars")
     def train_next_batch(self, batch_size=None):
         self.train_x_all = self.train_points()
-        self.train_x = self.bc_points()
+        if config.hvd is not None:
+            if self.train_distribution not in ["pseudo", "Sobol"]:
```
lululxvi, line 164: Check `self.train_distribution` in `__init__`

deepxde/geometry/sampler.py
```diff
@@ -51,8 +51,9 @@ def quasirandom(n_samples, dimension, sampler):
             sampler = skopt.sampler.Hammersly()
             skip = 1
     elif sampler == "Sobol":
+        randomize = config.hvd is not None
         # 1st point: [0, 0, ...], 2nd point: [0.5, 0.5, ...]
-        sampler = skopt.sampler.Sobol(randomize=False)
+        sampler = skopt.sampler.Sobol(randomize=randomize)
```
lululxvi, line 56: This `randomize` is only random shift, but still the same sequence. Not clear if and how well it works. So remove this, and more future study is needed,

deepxde/model.py
```diff
@@ -625,7 +634,8 @@ def train(
         self.callbacks.on_train_end()
 
         print("")
-        display.training_display.summary(self.train_state)
+        if config.rank == 0:
+            display.training_display.summary(self.train_state)
```
lululxvi, line 638: Add
```
# TODO Now only print the training loss in rank 0. The correct way is to print the average training loss of all ranks.
```

deepxde/model.py
```diff
@@ -595,19 +600,23 @@ def train(
...
             self.restore(model_restore_path, verbose=1)
 
-        print("Training model...\n")
+        if config.rank == 0:
+            print("Training model...\n")
         self.stop_training = False
         self.train_state.set_data_train(*self.data.train_next_batch(self.batch_size))
         self.train_state.set_data_test(*self.data.test())
-        self._test()
+        if config.rank == 0:
```
lululxvi, line 618: `self._test()` is required, as it will update `self.train_state`. Otherwise, some functions such as callbacks would fail.

deepxde/model.py
```diff
@@ -595,19 +600,23 @@ def train(
...
+            self._test()
```
lululxvi, line 619: In `self._test()`, add
```
# TODO Now only print the training loss in rank 0. The correct way is to print the average training loss of all ranks.
```

deepxde/model.py
```diff
@@ -647,7 +657,8 @@ def _train_sgd(self, iterations, display_every):
             self.train_state.epoch += 1
             self.train_state.step += 1
             if self.train_state.step % display_every == 0 or i + 1 == iterations:
-                self._test()
+                if config.rank == 0:
+                    self._test()
```
lululxvi, line 661: `self._test()` is needed

deepxde/data/pde.py
```diff
@@ -160,7 +160,24 @@ def losses(self, targets, outputs, loss_fn, inputs, model, aux=None):
...
+                raise ValueError(
+                "{} train distribution does not support data-parallel acceleration".format(
+                    self.train_distribution
+                ))
+            train_x = self.bc_points()
+            n_train = np.array([train_x.shape])
+            config.comm.Bcast(n_train, root=0)
```
lululxvi, line 171: No need to use `np.array` for int. Directly use
```
config.comm.bcast(n_train, root=0)
```

deepxde/data/pde.py
```diff
@@ -160,7 +160,24 @@ def losses(self, targets, outputs, loss_fn, inputs, model, aux=None):
...
+            
+            if train_x.shape[0] != n_train[0][0]:
+                print(n_train,'ntrain')
```
lululxvi, line 174: This print seems to be used for debug?

deepxde/data/pde.py
```diff
@@ -160,7 +160,24 @@ def losses(self, targets, outputs, loss_fn, inputs, model, aux=None):
...
+                train_x = np.zeros(n_train[0], dtype=train_x.dtype)
+
+            config.comm.Bcast(train_x, root=0)
+            self.train_x = train_x
```
lululxvi, line 178: Also need to update `self.num_bcs` and `self.train_x_bc` which is computed in `bc_points()`

deepxde/model.py
pescap: The loss in rank 0 after each allreduce operation corresponds to the average training loss of all ranks:

https://github.com/horovod/horovod/issues/843

deepxde/data/pde.py
pescap: `n_train` is a tuple of integers.

pescap: `n_train` is a tuple

deepxde/geometry/sampler.py
pescap: Work in progress
---
It sounds to be working fine. The shift uses the Cranly-Patterson rotation:

https://github.com/scikit-optimize/scikit-optimize/blob/273db8f3d223c4d471761e6d9ad1bf183b1ee6e1/skopt/sampler/sobol.py#L402-L408

It consists in shifting the points by a random uniform vector. 

See page 204 in the book:

Screenshot from 2023-04-25 14-31-48

deepxde/model.py
lululxvi: I suspect this. You can print the train loss for each rank in `self._test`, and see if they are the same.

deepxde/data/pde.py
lululxvi: Then you can do `n_train = np.array(train_x.shape)`. Why need the extra "[]"? Also `n_train` is confusing. maybe use `train_x_shape`.

Or you can use `n_train = len(train_x)` as an int, as we know the dim is `self.geom.dim`.

deepxde/geometry/sampler.py
lululxvi: Good reference. But it is still unclear that using multiple shifted Sobol seq would still have good properties/performance. I would suggest removing this from this PR, and opening a new one to investigate this.

deepxde/model.py
pescap: - The loss printed on rank 0 corresponds to to the average training loss of all ranks.
- I did not find out how to get the loss on rank 0 before the allreduce operation.

pescap: > I suspect this. You can print the train loss for each rank in `self._test`, and see if they are the same.
The loss for each rank is different.  An example where I print the loss for each rank at the beginning of `_test()` (l.814)
```
print(self.train_state.loss_train, 'LOSS', config.rank, 'RANK')
```


```
[1,0]<stdout>:Training model...
[1,0]<stdout>:
[1,0]<stdout>:[0.79626596 0.03356912] LOSS 0 RANK
[1,1]<stdout>:[0.7973283  0.03356912] LOSS 1 RANK
[1,3]<stdout>:[0.79765636 0.03356912] LOSS 3 RANK
[1,0]<stdout>:Step      Train loss              Test loss               Test metric
[1,0]<stdout>:0         [7.96e-01, 3.36e-02]    [7.90e-01, 3.36e-02]    []
[1,2]<stdout>:[0.7903435  0.03356912] LOSS 2 RANK
[1,0]<stdout>:[6.3809122e-05 4.1350492e-04] LOSS 0 RANK
[1,1]<stdout>:[6.2816704e-05 4.1350492e-04] LOSS 1 RANK
[1,3]<stdout>:[6.4307082e-05 4.1350492e-04] LOSS 3 RANK
[1,0]<stdout>:1000      [6.38e-05, 4.14e-04]    [4.25e-05, 4.14e-04]    []
[1,2]<stdout>:[6.0273003e-05 4.1350492e-04] LOSS 2 RANK
[1,1]<stdout>:[1.5275291e-05 3.0958428e-04] LOSS 1 RANK
[1,0]<stdout>:[1.5563643e-05 3.0958428e-04] LOSS 0 RANK
[1,3]<stdout>:[1.5553351e-05 3.0958428e-04] LOSS 3 RANK
[1,0]<stdout>:2000      [1.56e-05, 3.10e-04]    [1.18e-05, 3.10e-04]    []
[1,2]<stdout>:[1.5136841e-05 3.0958428e-04] LOSS 2 RANK
[1,0]<stdout>:[1.4009617e-05 3.0715909e-04] LOSS 0 RANK
[1,1]<stdout>:[1.4144935e-05 3.0715909e-04] LOSS 1 RANK
[1,3]<stdout>:[1.4283864e-05 3.0715909e-04] LOSS 3 RANK
[1,0]<stdout>:3000      [1.40e-05, 3.07e-04]    [1.10e-05, 3.07e-04]    []
[1,2]<stdout>:[1.4097668e-05 3.0715909e-04] LOSS 2 RANK
[1,1]<stdout>:[0.00040375 0.00295275] LOSS 1 RANK
[1,3]<stdout>:[0.00040177 0.00295275] LOSS 3 RANK
[1,0]<stdout>:[0.00039194 0.00295275] LOSS 0 RANK
[1,0]<stdout>:4000      [3.92e-04, 2.95e-03]    [4.31e-04, 2.95e-03]    []
[1,2]<stdout>:[0.00040245 0.00295275] LOSS 2 RANK
[1,0]<stdout>:[1.1745713e-05 3.0206933e-04] LOSS 0 RANK
[1,1]<stdout>:[1.1794020e-05 3.0206933e-04] LOSS 1 RANK
[1,3]<stdout>:[1.2182878e-05 3.0206933e-04] LOSS 3 RANK
[1,0]<stdout>:5000      [1.17e-05, 3.02e-04]    [9.39e-06, 3.02e-04]    []
```

pescap: I get the same second term for the loss on each rank (corresponding to bcs). The first term gives different values for each rank, as expected.

deepxde/model.py
```diff
@@ -647,7 +656,7 @@ def _train_sgd(self, iterations, display_every):
             self.train_state.epoch += 1
             self.train_state.step += 1
             if self.train_state.step % display_every == 0 or i + 1 == iterations:
-                self._test()
+                self._test()
```
lululxvi, line 659: Format via black.

deepxde/model.py
lululxvi: Each rank has the same BC points, and thus BC loss is the same for sure. For the PDE loss, different rank has different points, so the loss is different. The loss printed is the loss of each rank, not the average. As far as I know, Hovorod and all other ML frameworks don't compute the average loss. They do the average of the gradients. So add the comment suggested above. This can be easily fixed by using a reduction operator by mpi.

deepxde/data/pde.py
```diff
@@ -92,6 +92,13 @@ def __init__(
         self.num_domain = num_domain
         self.num_boundary = num_boundary
         self.train_distribution = train_distribution
+        if config.hvd is not None and self.train_distribution != 'pseudo':
+            raise ValueError(
```
lululxvi, line 96:
```
raise ValueError("Parallel training via Horovod only supports pseudo train distribution.")
```

deepxde/data/pde.py
```diff
@@ -160,7 +164,22 @@ def losses(self, targets, outputs, loss_fn, inputs, model, aux=None):
...
+            self.num_bcs = list(num_bcs)
+
+            if train_x.shape[0] != train_x_shape[0]:
+                train_x = np.zeros(train_x_shape, dtype=train_x.dtype)
+
+            config.comm.Bcast(train_x, root=0)
+            self.train_x_bc = train_x
+            self.train_x = train_x
+        else:
+            self.train_x = self.bc_points()
```
lululxvi, line 182:
```
self.bc_points()  # Generate self.num_bcs and self.train_x_bc
if self.bcs and config.hvd is not None:
    num_bcs = np.array(self.num_bcs)
    config.comm.Bcast(num_bcs, root=0)
    self.num_bcs = list(num_bcs)

    x_bc_shape = np.array(self.train_x_bc.shape)
    config.comm.Bcast(x_bc_shape, root=0)
    if len(self.train_x_bc) != x_bc_shape[0]:
        self.train_x_bc = np.empty(x_bc_shape, dtype=self.train_x_bc.dtype)
    config.comm.Bcast(self.train_x_bc, root=0)
self.train_x = self.train_x_bc
```

pescap:
Hi, are we ok now with this weak scaling acceleration? Best wishes,

lululxvi:
> Hi, are we ok now with this weak scaling acceleration? Best wishes,

Yes, it is good now. One last thing: in `PDE.__init__`, print an info that `num_domain` and `num_boundary` are the number of points in each process.

deepxde/data/pde.py
```diff
@@ -15,8 +15,8 @@ class PDE(Data):
         pde: A global PDE or a list of PDEs. ``None`` if no global PDE.
         bcs: A boundary condition or a list of boundary conditions. Use ``[]`` if no
             boundary condition.
-        num_domain (int): The number of training points sampled inside the domain.
-        num_boundary (int): The number of training points sampled on the boundary.
+        num_domain (int): The number of training points sampled inside the domain (over each rank for data parallel training).
```
lululxvi, line 18: Move this comment to line 95.

deepxde/data/pde.py
```diff
@@ -92,6 +92,12 @@ def __init__(
         self.num_domain = num_domain
         self.num_boundary = num_boundary
         self.train_distribution = train_distribution
+        if config.hvd is not None:
+            # num_domain and num_boundary are the number of points over each rank.
```
lululxvi, line 96: `print("When parallel training via Horovod, num_domain and num_boundary are the numbers of points over each rank.")`

lululxvi:
Do you want to run your test cases again for double check?

Also, we need to add documentation about this in a new PR.

pescap:
Done, double check performed: 
Tested with 8 GPUs for `elliptic_inverse_field` and `Poisson_Lshape.py`.

pescap:
Several PRs to come:
- Update Docker image with `mpi4py`
- Add documentation for weak scaling
- Add parameter for scaling mode (weak or strong)
- Implement strong scaling for `tensorflow.compat.v1`
- Study Sobol distribution for weak scaling
- Study resampling strategies for weak scaling

examples/pinn_forward/Poisson_Dirichlet_1d.py
```diff
@@ -30,7 +30,7 @@ def func(x):
 
 geom = dde.geometry.Interval(-1, 1)
 bc = dde.icbc.DirichletBC(geom, func, boundary)
-data = dde.data.PDE(geom, pde, bc, 16, 2, solution=func, num_test=100)
```
lululxvi, line 33: Why modify this?

examples/pinn_forward/Poisson_Dirichlet_1d.py
```diff
@@ -30,7 +30,7 @@ def func(x):
...
+data = dde.data.PDE(geom, pde, bc, 0, 2, solution=func, num_test=100)
```
pescap, line 33:
```suggestion
data = dde.data.PDE(geom, pde, bc, 16, 2, solution=func, num_test=100)
```

examples/pinn_forward/Poisson_Dirichlet_1d.py
pescap: Corrected, sorry.

PR #1213, opened 2023-03-28
Change location of setting default backend to file

lijialin03:
Change location to set default backend after all checks for backend.


deepxde/backend/__init__.py
```diff
@@ -33,6 +33,7 @@ def load_backend(mod_name):
         raise NotImplementedError("Unsupported backend: %s" % mod_name)
 
     print("Using backend: %s\n" % mod_name, file=sys.stderr, flush=True)
+    set_default_backend(mod_name)
```
lululxvi, line 36: We only set default backend for user when the user didn't specify a valid one.

PR #1216, opened 2023-03-28
Modify a example to support paddle

lijialin03:


examples/pinn_forward/Helmholtz_Dirichlet_2d_HPO.py
```diff
@@ -11,6 +11,8 @@
 
 if dde.backend.backend_name == "pytorch":
     sin = dde.backend.pytorch.sin
+elif dde.backend.backend_name == "paddle":
+    sin = dde.backend.paddle.sin
```
lijialin03, line 15: Sorry, I do not fully understand the comment. Does that mean code should be like this?

lululxvi: First line:
"""Backend supported: tensorflow.compat.v1, tensorflow, pytorch, paddle"""

PR #1217, opened 2023-03-28
Modify some examples to support paddle with an uniform format

lijialin03:
Try to modify some examples to support paddle with an uniform format if a example could be modified simply.

Uniform format:

    # Define function
    if dde.backend.backend_name == "paddle":
        # Backend paddle
        import paddle
    
        xxx = paddle.xxx
    elif dde.backend.backend_name == "pytorch":
        # Backend pytorch
        import torch
    
        xxx = torch.xxx
    elif dde.backend.backend_name in ["tensorflow.compat.v1", "tensorflow"]:
        # Backend tensorflow.compat.v1 or tensorflow
        
        xxx = tf.xxx
    elif dde.backend.backend_name == "jax":
        # Backend jax
        import jax.numpy as jnp
    
        xxx = jnp.xxx

### DIFF
--- a/examples/pinn_forward/Allen_Cahn.py
+++ b/examples/pinn_forward/Allen_Cahn.py
@@ -5,12 +5,28 @@ Implementation of Allen-Cahn equation example in paper https://arxiv.org/abs/211
 import deepxde as dde
 import numpy as np
 from scipy.io import loadmat
-# Import tf if using backend tensorflow.compat.v1 or tensorflow
-from deepxde.backend import tf
-# Import torch if using backend pytorch
-# import torch
-# Import paddle if using backend paddle
-# import paddle
+
+# Define function
+if dde.backend.backend_name == "paddle":
+    # Backend paddle
+    import paddle
+
+    cos = paddle.cos
+elif dde.backend.backend_name == "pytorch":
+    # Backend pytorch
+    import torch
+
+    cos = torch.cos
+elif dde.backend.backend_name in ["tensorflow.compat.v1", "tensorflow"]:
+    # Backend tensorflow.compat.v1 or tensorflow
+    from deepxde.backend import tf
+
+    cos = tf.cos
+elif dde.backend.backend_name == "jax":
+    # Backend jax
+    import jax.numpy as jnp
+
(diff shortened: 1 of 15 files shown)


examples/pinn_forward/Allen_Cahn.py
```diff
@@ -5,13 +5,23 @@
...
 import numpy as np
 from scipy.io import loadmat
-# Import tf if using backend tensorflow.compat.v1 or tensorflow
-from deepxde.backend import tf
-# Import torch if using backend pytorch
-# import torch
-# Import paddle if using backend paddle
-# import paddle
 
+# Define function
```
lululxvi, line 9: No need to modify the code. The current version works well.

examples/pinn_forward/Laplace_disk.py
```diff
@@ -1,13 +1,29 @@
...
 import deepxde as dde
 import numpy as np
-# Import tf if using backend tensorflow.compat.v1 or tensorflow
-from deepxde.backend import tf
-# Import torch if using backend pytorch
-# import torch
-# Import paddle if using backend paddle
-# import paddle
 
+# Define function
```
lululxvi, line 5: also no need to mofidy. The same for other examples.

lijialin03:
We still wish to discuss the possibility of unifying examples like this commit, although them work well now.
There are two reasons:
1. This form is a more friendly for beginner who barely know about programming and just want to try it. Program will not fail when they run it without a certain backend. As for those skilled coders, they will not be affected because they will can run it in whatever form.
2. This form is more convenient if user want to exchange their backend.
Thanks for your attention.

lululxvi:
Your current design doesn't support jax.

lijialin03:
> Your current design doesn't support jax.

I add jax part.
However, these examples doesn't support jax now(master). I'm not sure if should i change them.

lululxvi:
The way you did doesn't support jax, see example: 
https://github.com/lululxvi/deepxde/blob/5bd9fa49de24ed55412a1b1a30cc1bbea3140d25/examples/pinn_forward/ode_system.py#L15

PR #1224, opened 2023-03-30
Restore model to a specified device

bonneted:
Allow restoring model on another device than the one it was saved from (e. g. saved on GPU and restored on CPU)
Only implemented for PyTorch.


deepxde/model.py
```diff
@@ -1014,21 +1014,29 @@ def save(self, save_path, protocol="backend", verbose=0):
             )
         return save_path
 
-    def restore(self, save_path, verbose=0):
+    def restore(self, save_path, device=None, verbose=0):
         """Restore all variables from a disk file.
 
         Args:
             save_path (string): Path where model was previously saved.
+            device (optional): Device to load the model on. By default, the model is loaded on the device it was saved from.
```
lululxvi, line 1022: What is the device format? string like "cpu", "gpu:0"?

lululxvi: Can torch.device be used in tensorflow/jax?

Using a string is better, so it is independent to the backend.

bonneted: Just update the code so that the device is a string

deepxde/model.py
```diff
@@ -1014,21 +1014,29 @@ def save(self, save_path, protocol="backend", verbose=0):
             )
         return save_path
 
-    def restore(self, save_path, verbose=0):
+    def restore(self, save_path, device=None, verbose=0):
         """Restore all variables from a disk file.
 
         Args:
             save_path (string): Path where model was previously saved.
+            device (string): Optionnal, device to load the model on (e.g. "cpu","gpu:0","cuda:1"...). By default, the model is loaded on the device it was saved from.
```
lululxvi, line 1022: Typo: Optional

lululxvi, line 1022: What is the difference between "gpu:x","cuda:x"? I think you mean the same thing. Then let us just do "gpu:x".

bonneted: In Pytorch you never use "gpu:x", you have to specify the API you're using. Cuda is the most common, but here are all the possibilities: cpu, cuda, xpu, mkldnn, opengl, opencl, ideep, hip, ve, ort, mlc, xla, lazy, vulkan, meta, hpu
I let "gpu" thinking it could be the way to do it for other libraries. But that can be indeed confusing, so as long as only Pytorch is implemented I can maybe remove the "gpu:x"

deepxde/model.py
```diff
@@ -1014,21 +1014,29 @@ def save(self, save_path, protocol="backend", verbose=0):
             )
         return save_path
 
-    def restore(self, save_path, verbose=0):
+    def restore(self, save_path, device=None, verbose=0):
         """Restore all variables from a disk file.
 
         Args:
             save_path (string): Path where model was previously saved.
+            device (string, optionnal): Device to load the model on (e.g. "cpu","cuda:0"...). By default, the model is loaded on the device it was saved from.
```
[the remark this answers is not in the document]
bonneted: fixed

PR #1232, opened 2023-04-03
Update pointcloud.py to accept NeumannBC

mitchelldaneker:
Added `boundary_normal(self,x)` function so NeumannBC can be used. Has been tested on simple 1D case and same solution was found for self defined normals.


deepxde/geometry/pointcloud.py
```diff
@@ -68,6 +68,13 @@ def on_boundary(self, x):
             .any(axis=1)
         )
 
+    def boundary_normal(self, x):
+        if self.boundary_normals is None:
+            raise ValueError(
+                "boundary_normals must be defined for boundary_normal"
+            )
+        return self.boundary_normals
```
lululxvi, line 76: Should only return the normal for those x in the function argument.

lululxvi: @mitchelldaneker

mitchelldaneker: I need to determine a way to find those x in the function argument. So far, the only methods I have been able to create are slow, and even when using matrix alg. are not fast enough to work for when there are many inputs.

mitchelldaneker: I have implemented a method similar to on_boundary. This uses np.where, and allows for repeats in the x input.

deepxde/geometry/pointcloud.py
```diff
@@ -62,6 +62,17 @@ def on_boundary(self, x):
             .any(axis=1)
         )
 
+    def boundary_normal(self, x):
+        if self.boundary_normals is None:
+            raise ValueError(
+                "boundary_normals must be defined for boundary_normal"
+            )
+        boundary_point_matches = np.isclose(
```
lululxvi, line 70: Replace `np.isclose` with `isclose`

lululxvi:
Please fix the Codacy error.

mitchelldaneker:
> Please fix the Codacy error.

The current implementation is correct as per numpy documentation. Suggestions are incorrect as well.

lululxvi:
> > Please fix the Codacy error.
> 
> The current implementation is correct as per numpy documentation. Suggestions by Codacy are incorrect as well.

Are you sure? References?

mitchelldaneker:
> > > Please fix the Codacy error.
> > 
> > 
> > The current implementation is correct as per numpy documentation. Suggestions by Codacy are incorrect as well.
> 
> Are you sure? References?

Removed a redundancy which should fix the error. It seemed codacy did not know what I was trying to do, so gave incorrect suggestions that broke the code. Since it is already a boolean array, do not need `== True`.

PR #1258, opened 2023-04-18
Add approximate distance functions for hard-constraint methods

csuastt:
ADFs are implemented for all geometries except CSGDifference, CSGxxx, Point Cloud, and Polygon. The functions have been tested.

### DIFF
--- a/deepxde/backend/backend.py
+++ b/deepxde/backend/backend.py
@@ -305,6 +305,14 @@ def square(x):
     """Returns the square of the elements of input."""
 
 
+def abs(x):
+    """Computes the absolute value of a tensor."""
+
+
+def minimum(x, y):
+    """Returns the minimum of x and y (i.e. x < y ? x : y) element-wise."""
+
+
 def tanh(x):
     """Computes hyperbolic tangent of x element-wise."""
 
--- a/deepxde/backend/backend.py
+++ b/deepxde/backend/backend.py
@@ -305,7 +305,7 @@ def square(x):
     """Returns the square of the elements of input."""
 
 
-def absolute(x):
+def abs(x):
     """Computes the absolute value element-wise."""
 
 
@@ -366,7 +366,7 @@ def prod(input_tensor, dim, keepdims=False):
     Args:
         input_tensor (Tensor). The input tensor.
         dim (int). The reduce dim.
-        keepdims (bool). Whether to keep the summed dimension.
+        keepdims (bool). Whether to keep the product dimension.
 
     Returns:
         Tensor: A framework-specific tensor.
@@ -384,13 +384,13 @@ def reduce_prod(input_tensor):
     """
 
 
-def amin(input_tensor, dim, keepdims=False):
+def min(input_tensor, dim, keepdims=False):
     """Returns the minimum of the input tensor along the given dim.
 
     Args:
         input_tensor (Tensor). The input tensor.
         dim (int). The reduce dim.
-        keepdims (bool). Whether to keep the summed dimension.
+        keepdims (bool). Whether to keep the dimension.
 
     Returns:
         Tensor: A framework-specific tensor.
@@ -408,13 +408,13 @@ def reduce_min(input_tensor):
     """
 
 
-def amax(input_tensor, dim, keepdims=False):
+def max(input_tensor, dim, keepdims=False):
     """Returns the maximum of the input tensor along the given dim.
 
     Args:
         input_tensor (Tensor). The input tensor.
         dim (int). The reduce dim.
-        keepdims (bool). Whether to keep the summed dimension.
+        keepdims (bool). Whether to keep the dimension.
 
     Returns:
         Tensor: A framework-specific tensor.
--- a/deepxde/backend/backend.py
+++ b/deepxde/backend/backend.py
@@ -360,6 +360,78 @@ def reduce_sum(input_tensor):
     """
 
 
+def prod(input_tensor, dim, keepdims=False):
+    """Returns the product of the input tensor along the given dim.
+
+    Args:
+        input_tensor (Tensor). The input tensor.
+        dim (int). The reduce dim.
+        keepdims (bool). Whether to keep the summed dimension.
+
+    Returns:
+        Tensor: A framework-specific tensor.
+    """
+
+
+def reduce_prod(input_tensor):
+    """Returns the product of all elements in the input tensor.
+
+    Args:
+        input_tensor (Tensor). The input tensor.
+
+    Returns:
+        Tensor.
+    """
+
+
+def min(input_tensor, dim, keepdims=False):
+    """Returns the minimum of the input tensor along the given dim.
+
+    Args:
+        input_tensor (Tensor). The input tensor.
+        dim (int). The reduce dim.
+        keepdims (bool). Whether to keep the summed dimension.
+
+    Returns:
+        Tensor: A framework-specific tensor.
+    """
+
+
+def reduce_min(input_tensor):
+    """Returns the minimum of all elements in the input tensor.
+
+    Args:
+        input_tensor (Tensor). The input tensor.
+
+    Returns:
+        Tensor.
+    """
+
+
+def max(input_tensor, dim, keepdims=False):
+    """Returns the maximum of the input tensor along the given dim.
+
+    Args:
+        input_tensor (Tensor). The input tensor.
+        dim (int). The reduce dim.
+        keepdims (bool). Whether to keep the summed dimension.
+
+    Returns:
+        Tensor: A framework-specific tensor.
+    """
+
+
+def reduce_max(input_tensor):
+    """Returns the maximum of all elements in the input tensor.
+
+    Args:
+        input_tensor (Tensor). The input tensor.
+
+    Returns:
+        Tensor.
+    """
+
+
 def norm(tensor, ord=None, axis=None, keepdims=False):
     """Computes a vector norm.
 
--- a/deepxde/backend/backend.py
+++ b/deepxde/backend/backend.py
@@ -305,8 +305,8 @@ def square(x):
     """Returns the square of the elements of input."""
 
 
-def abs(x):
-    """Computes the absolute value of a tensor."""
+def absolute(x):
+    """Computes the absolute value element-wise."""
 
 
 def minimum(x, y):
@@ -384,7 +384,7 @@ def reduce_prod(input_tensor):
     """
 
 
-def min(input_tensor, dim, keepdims=False):
+def amin(input_tensor, dim, keepdims=False):
     """Returns the minimum of the input tensor along the given dim.
 
     Args:
@@ -408,7 +408,7 @@ def reduce_min(input_tensor):
     """
 
 
-def max(input_tensor, dim, keepdims=False):
+def amax(input_tensor, dim, keepdims=False):
     """Returns the maximum of the input tensor along the given dim.
 
     Args:
--- a/deepxde/backend/paddle/tensor.py
+++ b/deepxde/backend/paddle/tensor.py
@@ -146,6 +146,7 @@ def square(x):
     return paddle.square(x)
 
 
+# pylint: disable=redefined-builtin
 def abs(x):
     return paddle.abs(x)
 
@@ -186,6 +187,7 @@ def reduce_prod(input_tensor):
     return paddle.prod(input_tensor)
 
 
+# pylint: disable=redefined-builtin
 def min(input_tensor, dim, keepdims=False):
     return paddle.min(input_tensor, axis=dim, keepdim=keepdims)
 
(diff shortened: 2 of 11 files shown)


deepxde/backend/backend.py
```diff
@@ -305,6 +305,14 @@ def square(x):
     """Returns the square of the elements of input."""
 
 
+def absolute(x):
```
lululxvi, line 308: `def abs(x):`

deepxde/backend/backend.py
```diff
@@ -352,6 +360,78 @@ def reduce_sum(input_tensor):
     """
 
 
+def prod(input_tensor, dim, keepdims=False):
+    """Returns the product of the input tensor along the given dim.
+
+    Args:
+        input_tensor (Tensor). The input tensor.
+        dim (int). The reduce dim.
+        keepdims (bool). Whether to keep the summed dimension.
```
lululxvi, line 369: `keepdims (bool). Whether to keep the product dimension.`

deepxde/backend/backend.py
```diff
@@ -352,6 +360,78 @@ def reduce_sum(input_tensor):
...
+    """
+
+
+def amin(input_tensor, dim, keepdims=False):
+    """Returns the minimum of the input tensor along the given dim.
+
+    Args:
+        input_tensor (Tensor). The input tensor.
+        dim (int). The reduce dim.
+        keepdims (bool). Whether to keep the summed dimension.
```
lululxvi, line 393: `keepdims (bool). Whether to keep the dimension.`

deepxde/backend/backend.py
```diff
@@ -352,6 +360,78 @@ def reduce_sum(input_tensor):
...
+
+    Args:
+        input_tensor (Tensor). The input tensor.
+
+    Returns:
+        Tensor.
+    """
+
+
+def amin(input_tensor, dim, keepdims=False):
```
lululxvi, line 387: `amin` --> `min`

deepxde/backend/backend.py
```diff
@@ -352,6 +360,78 @@ def reduce_sum(input_tensor):
...
+
+    Args:
+        input_tensor (Tensor). The input tensor.
+
+    Returns:
+        Tensor.
+    """
+
+
+def amax(input_tensor, dim, keepdims=False):
```
lululxvi, line 411: `max`

deepxde/backend/paddle/tensor.py
```diff
@@ -170,6 +178,30 @@ def reduce_sum(input_tensor):
...
 
+def prod(input_tensor, dim, keepdims=False):
+    return paddle.prod(input_tensor, axis=dim, keepdim=keepdims)
+
+
+def reduce_prod(input_tensor):
+    return paddle.prod(input_tensor)
+
+
+def amax(input_tensor, dim, keepdims=False):
```
lululxvi, line 189: `min` should be before `max`.

deepxde/backend/tensorflow/tensor.py
```diff
@@ -131,6 +139,30 @@ def reduce_sum(input_tensor):
...
 
+def prod(input_tensor, dim, keepdims=False):
+    return tf.math.reduce_prod(input_tensor, axis=dim, keepdims=keepdims)
+
+
+def reduce_prod(input_tensor):
+    return tf.math.reduce_prod(input_tensor)
+
+
+def amax(input_tensor, dim, keepdims=False):
```
lululxvi, line 150: `min` should be before `max`.

deepxde/geometry/geometry.py
```diff
@@ -27,6 +27,46 @@ def mindist2boundary(self, x):
         raise NotImplementedError(
             "{}.mindist2boundary to be implemented".format(self.idstr)
         )
+    
+    def approxdist2boundary(self, x, where: Union[
+        None, Literal["left", "right", "top", "bottom"]] = None,
+        smoothness: Literal["L", "M", "H"] = "M"):
+        """Compute the approximate distance at x to the boundary.
```
lululxvi, line 34: Wrong docstring format. Check other examples of the format used in DeepXDE.

deepxde/geometry/geometry.py
```diff
@@ -27,6 +27,46 @@ def mindist2boundary(self, x):
...
+        - This function is used for the hard-constraint methods.
+        - The approximate distance function satisfies the following properties:
```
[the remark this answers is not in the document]
csuastt: Unfortunately, there is no such guarantee and it is easy to draw counterexamples. But a weaker property is bound to hold: the approximate distance approaches infinity if the true distance approaches infinity, and vice versa.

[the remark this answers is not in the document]
csuastt: Another property may also hold (while I'm not sure): let $B(r) = \\{ x | \mathrm{approxdist}(x) < r \\}$, then for any positive number $r$, $B(r)$ is always simply connected.

deepxde/geometry/geometry.py
```diff
@@ -27,6 +27,41 @@ def mindist2boundary(self, x):
         raise NotImplementedError(
             "{}.mindist2boundary to be implemented".format(self.idstr)
         )
+    
+    def approxdist2boundary(self, x, 
+        smoothness: Literal["L", "M", "H"] = "M"):
+        """Compute the approximate distance at x to the boundary.
+        - This function is used for the hard-constraint methods.
+        - The approximate distance function satisfies the following properties:
+            - The function is zero on the boundary and positive elsewhere.
```
lululxvi, line 36: `positive elsewhere`? or `positive elsewhere inside the geometry`?

deepxde/geometry/geometry.py
```diff
@@ -27,6 +27,41 @@ def mindist2boundary(self, x):
         raise NotImplementedError(
             "{}.mindist2boundary to be implemented".format(self.idstr)
         )
+    
+    def approxdist2boundary(self, x, 
+        smoothness: Literal["L", "M", "H"] = "M"):
+        """Compute the approximate distance at x to the boundary.
```
lululxvi, line 33: Some docstring has the wrong format. See the generated doc at https://deepxde--1258.org.readthedocs.build/en/1258/

deepxde/geometry/geometry_1d.py
```diff
@@ -1,8 +1,10 @@
 import numpy as np
 
 from .geometry import Geometry
+from typing import Literal, Union
 from .sampler import sample
 from .. import config
+from .. import backend as bkd
```
lululxvi, line 7:
```
from .. import backend as bkd
from .. import config
```

deepxde/geometry/geometry.py
csuastt: Yes, it's `positive elsewhere`. This is similar to `abs(signed distance function)` or `square(signed distance function)`. Suppose we have a ring, for the inner circle, the distance function is positive outside.

lululxvi: There is no need to restrict the value outside the geometry, which may make the implementation more complex in some cases. We should follow this rule:
`The function is zero on the boundary and positive inside the geometry. The function value outside the geometry has no restriction`.

deepxde/geometry/geometry.py
```diff
@@ -27,6 +27,44 @@ def mindist2boundary(self, x):
...
+        - The function is zero on the boundary and positive elsewhere.
+        - The function is almost differentiable at any order.
+        - The function is not necessarily equal to the exact distance function.
+
+        Args:
+            x: A 2D array of shape (n, dim), where `n` is the number of points and
+                `dim` is the dimension of the geometry. Note that `x` should be a tensor type
+                of backend (e.g., `tf.Tensor` or `torch.Tensor`), not a numpy array.
+            smoothness (string, optional): A string to specify the smoothness of the distance function,
+                e.g., "L", "M", "H". "L" is the least smooth, "H" is the most smooth.
```
lululxvi, line 46: "L", "M", and "H" are not standard math languages. I think of something like C^k smoothness https://en.wikipedia.org/wiki/Smoothness, but this is not exactly what you do here.

Another idea: instead of using string "L", "M", "H", I prefer to use numbers 1, 2, 3 for the smoothness level.

deepxde/geometry/geometry.py
```diff
@@ -27,6 +27,44 @@ def mindist2boundary(self, x):
...
+                of backend (e.g., `tf.Tensor` or `torch.Tensor`), not a numpy array.
+            smoothness (string, optional): A string to specify the smoothness of the distance function,
+                e.g., "C0", "Cinf", "Cinf+". "C0" is the least smooth, "Cinf+" is the most smooth.
+                Default is "Cinf".
+
+                - C0
+                The distance function is continuous but can be non-differentiable on a
+                set of points, which has measure zero.
+
+                - Cinf
```
lululxvi, line 53: It seems to me this is also C0. The difference is that the C0 above is not differentiable on some points inside, but here not differentiable on the points on the boundary. I would call this as C0+, and the Cinf+ as Cinf.

deepxde/geometry/geometry.py
```diff
@@ -27,6 +27,44 @@ def mindist2boundary(self, x):
...
+                The distance function is continuous and differentiable at any order. The
+                non-differentiable points can only appear on boundaries. If the points in `x` are
+                all inside or outside the geometry, the distance function is smooth.
+
+                - Cinf+
+                The distance function is continuous and differentiable at any order on any 
+                points. This option may result in a polynomial of HIGH order.
+
+        Returns:
+            A NumPy array of shape (n, 1). The distance at each point in `x`.
```
lululxvi, line 63: The return type is numpy or tensor?

deepxde/geometry/geometry.py
```diff
@@ -27,6 +27,44 @@ def mindist2boundary(self, x):
         raise NotImplementedError(
             "{}.mindist2boundary to be implemented".format(self.idstr)
         )
+    
+    def approxdist2boundary(self, x, smoothness: Literal["C0", "Cinf", "Cinf+"] = "Cinf"):
+        """Compute the approximate distance at x to the boundary.
```
lululxvi, line 32: `approximate distance` is not clear. Actually, it doesn't approximate the distance in many cases. Need to come up with another name.

csuastt: Is it ok with `pseudo distance`?

lululxvi: It seems better. but still not good enough. I don't have a good name in mind.

csuastt: Here are a few suggestions from the ChatGPT:

1. `hardConstraintFactor`: This name emphasizes the function's role in enforcing hard constraints.

2. `boundaryConditionScaler`: This name suggests that the function scales the boundary condition in the ansatz.

3. `constraintEnforcingMeasure`: This name suggests that the function measures something that helps enforce constraints.

4. `boundaryConditionIndicator`: This name suggests that the function indicates whether the boundary condition should be applied.

5. `constraintActivationFunction`: This name suggests that the function activates the constraint when necessary.

6. `boundaryEnforcingFunction`: This name suggests that the function plays a role in enforcing the boundary condition.

lululxvi: How about `boundary_constraint_factor`?

csuastt: Okay, I'll go with it.

deepxde/geometry/geometry.py
```diff
@@ -1,5 +1,5 @@
 import abc
-
+from typing import Literal
 import numpy as np
```
lululxvi, line 3:
```
import abc
from typing import Literal

import numpy as np
```

deepxde/geometry/geometry_2d.py
```diff
@@ -251,6 +266,139 @@ def random_boundary_points(self, n, random="pseudo"):
                 x.append([self.xmin[0], self.xmax[1] - l + l3])
         return np.vstack(x)
 
+    def _boundary_constraint_factor_inside(self, x, where: Union[
```
lululxvi, line 269: Format all the code using black https://github.com/psf/black

deepxde/geometry/geometry_1d.py
```diff
@@ -22,6 +25,87 @@ def distance2boundary(self, x, dirn):
...
+
+            where (string, optional): A string to specify which part of the boundary to compute the distance,
+                e.g., "left", "right". If `None`, compute the distance to the whole boundary. Default is `None`.
+
+        Returns:
+            A tensor of a type determined by the backend, which will have a shape of (n, 1).
+            Each element in the tensor corresponds to the computed distance value for the respective point in `x`.
+        """
+
+        assert where in [None, "left", "right"], "where must be None, left, or right"
```
lululxvi, line 76: Change `assert` to `if` and `raise`. See https://softwareengineering.stackexchange.com/questions/225956/python-assert-vs-if-return

lululxvi:
The code looks good. Please double check and confirm if everything is good.

PR #1262, opened 2023-04-20
fix very small bug with float16

g-w1:
Sorry for not catching this in #1132

### DIFF
--- a/deepxde/config.py
+++ b/deepxde/config.py
@@ -28,7 +28,10 @@ def default_float():
     """Returns the default float type, as a string."""
     if real.precision == 64:
         return "float64"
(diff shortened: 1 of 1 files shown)


deepxde/config.py
```diff
@@ -28,7 +28,10 @@ def default_float():
     """Returns the default float type, as a string."""
     if real.precision == 64:
         return "float64"
-    return "float32"
+    elif real.precision == 32:
+        return "float32"
+    else:
+        return "float16"
```
lululxvi, line 34:
```
elif real.precision == 16:
    return "float16"
```

PR #1273, opened 2023-04-26
updata function 'as_tensor' in backend 'paddle'

lijialin03:
'paddle.to_tensor()' can not transform int/float to list in paddle develop version after 2023.4.26 so that tensor.shape = [] will appear. 
Update as_tensor() to adapt to this change.


lijialin03:
> Why has transform int/float to list? A tensor of type int/float is also OK.
paddle.to_tensor(int) returns Tensor(shape=[])，which causes failure when send it to paddle.concate() in paddle develop version after 2023.4.26.
To verify this error, you can download newest version of paddle develop and run example fractional_diffusion_1d.py.

PR #1274, opened 2023-04-27
update import of LBFGS in backend 'paddle'

lijialin03:
LBFGS moved from paddle.incubate.optimizer to paddle.optimizer in paddle develop version after 2023.4.27  and also in incoming paddle release 2.5 version. Modify import in code to adapt to it.
What is more, 'lr' param of LBFGS has been change to 'learning_rate' to keep consistent with other optimizers of paddle.

### DIFF
--- a/deepxde/optimizers/paddle/optimizers.py
+++ b/deepxde/optimizers/paddle/optimizers.py
@@ -1,7 +1,7 @@
 __all__ = ["get", "is_external_optimizer"]
 
 import paddle
-from paddle.incubate.optimizer import LBFGS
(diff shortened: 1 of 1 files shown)


lijialin03:
@lululxvi 
This pr can not pass macos ci because I find that this ci installs paddle in an old URL that has been deprecated. 
The old url is https://www.paddlepaddle.org.cn/whl/mac/cpu-noavx/develop.html.
The url now is https://www.paddlepaddle.org.cn/whl/mac/cpu/develop.html which is shown in paddle official website.
So url in ci should be changed now.

PR #1276, opened 2023-04-29
update import of LBFGS in backend 'paddle'

lijialin03:
LBFGS moved from paddle.incubate.optimizer to paddle.optimizer in paddle develop version after 2023.4.27 and also in incoming paddle release 2.5 version. Modify import in code to adapt to it.
What is more, 'lr' param of LBFGS has been change to 'learning_rate' to keep consistent with other optimizers of paddle.

### DIFF
--- a/deepxde/optimizers/paddle/optimizers.py
+++ b/deepxde/optimizers/paddle/optimizers.py
@@ -1,7 +1,6 @@
 __all__ = ["get", "is_external_optimizer"]
 
 import paddle
-from paddle.incubate.optimizer import LBFGS
 
 from ..config import LBFGS_options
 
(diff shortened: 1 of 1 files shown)


deepxde/optimizers/paddle/optimizers.py
```diff
@@ -29,7 +29,7 @@ def get(params, optimizer, learning_rate=None, decay=None):
         if learning_rate is not None or decay is not None:
             print("Warning: learning rate is ignored for {}".format(optimizer))
         optim = LBFGS(
```
lululxvi, line 31: `paddle.optimizer.LBFGS`

PR #1282, opened 2023-05-01
prepare strong scaling

pescap:
I want to prepare everything so that the implementation of strong scaling is straightforward.
In parallel, I will be working on the code for MPI splitting, and will add the code in another PR.

### DIFF
--- a/deepxde/data/pde.py
+++ b/deepxde/data/pde.py
@@ -93,16 +93,17 @@ class PDE(Data):
         self.num_boundary = num_boundary
         self.train_distribution = train_distribution
         if config.hvd is not None:
-            print(
-                "When parallel training via Horovod, num_domain and num_boundary are the numbers of points over each rank, not the total number of points."
-            )
             if self.train_distribution != "pseudo":
                 raise ValueError(
                     "Parallel training via Horovod only supports pseudo train distribution."
                 )
-            if config.parallel_scaling == "strong":
-                raise ValueError(
-                    "Strong scaling is not supported with tensorflow.compat.v1. Please use weak scaling."
+            if config.parallel_scaling == "weak":
+                print(
+                    "For weak scaling, num_domain and num_boundary are the numbers of points over each rank, not the total number of points."
+                )
+            else:
+                print(
+                    "For strong scaling, num_domain and num_boundary are the total number of points."
                 )
         self.anchors = None if anchors is None else anchors.astype(config.real(np))
         self.exclusions = exclusions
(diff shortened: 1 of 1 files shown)


deepxde/data/pde.py
```diff
@@ -93,16 +93,17 @@ def __init__(
...
                     "Parallel training via Horovod only supports pseudo train distribution."
                 )
-            if config.parallel_scaling == "strong":
-                raise ValueError(
-                    "Strong scaling is not supported with tensorflow.compat.v1. Please use weak scaling."
+            if config.parallel_scaling == "weak":
+                print(
+                    "For weak scaling, num_domain and num_boundary are the numbers of points over each rank, not the total number of points."
+                )
+            else:
```
lululxvi, line 104: `elif config.parallel_scaling == "strong":`

lululxvi:
Great. For MPI splitting, you can look at the code https://github.com/lululxvi/deepxde/pull/1094

deepxde/data/pde.py
```diff
@@ -174,15 +175,21 @@ def train_next_batch(self, batch_size=None):
...
+                x_bc_shape = np.array(self.train_x_bc.shape)
+                config.comm.Bcast(x_bc_shape, root=0)
+                if len(self.train_x_bc) != x_bc_shape[0]:
+                    self.train_x_bc = np.empty(x_bc_shape, dtype=self.train_x_bc.dtype)
+                config.comm.Bcast(self.train_x_bc, root=0)
+            else:
+                # Todo: Split the domain training points over rank for strong scaling.
+                raise ValueError(
+                    "Strong scaling is not supported yet with tensorflow.compat.v1. Please use weak scaling."
+                )
```
lululxvi, line 192: BC points are always the same for both strong and weak scaling. PDE points in line 195 below are different.

deepxde/data/pde.py
```diff
@@ -171,6 +172,11 @@ def losses(self, targets, outputs, loss_fn, inputs, model, aux=None):
 
     @run_if_all_none("train_x", "train_y", "train_aux_vars")
     def train_next_batch(self, batch_size=None):
+        if config.parallel_scaling == "strong":
```
lululxvi, line 175: This doesn't seem useful, as in config.py, we have
https://github.com/lululxvi/deepxde/blob/47127ac60496a52ccd01d17fcea9b67cc3afc5ad/deepxde/config.py#L26

pescap: It is, as strong scaling is set at the beginning of any example script with:
```
dde.config.set_parallel_scaling('strong')
```

PR #1284, opened 2023-05-03
Strong scaling data parallel acceleration for tensorflow.compat.v1

pescap:
Hi! At least it works. Some improvements can be done I guess.

### DIFF
--- a/deepxde/data/pde.py
+++ b/deepxde/data/pde.py
@@ -188,14 +188,15 @@ class PDE(Data):
         if config.parallel_scaling == "strong":
             # Split the training points over each rank.
             # We drop last points in order to have the same number of points per rank
-            train_x_all = self.train_x_all
             train_x_all_shape = list(
                 self.train_x_all.shape
             )  # We transform to list to support item assignment
             num_split = train_x_all_shape[0] // config.world_size
             train_x_all_shape[0] = num_split
-            train_x_all_split = np.empty(train_x_all_shape, dtype=train_x_all.dtype)
-            config.comm.Scatter(train_x_all, train_x_all_split, root=0)
+            train_x_all_split = np.empty(
+                train_x_all_shape, dtype=self.train_x_all.dtype
+            )
+            config.comm.Scatter(self.train_x_all, train_x_all_split, root=0)
             self.train_x_all = train_x_all_split
         if self.pde is not None:
             self.train_x = np.vstack((self.train_x, self.train_x_all))
--- a/deepxde/data/pde.py
+++ b/deepxde/data/pde.py
@@ -192,6 +192,10 @@ class PDE(Data):
                 self.train_x_all.shape
             )  # We transform to list to support item assignment
             num_split = train_x_all_shape[0] // config.world_size
+            if num_split < config.world_size:
+                raise ValueError(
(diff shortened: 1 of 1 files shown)


pescap:
`heat.py` tested with weak and strong scaling

deepxde/data/pde.py
```diff
@@ -190,6 +185,18 @@ def train_next_batch(self, batch_size=None):
...
             config.comm.Bcast(self.train_x_bc, root=0)
         self.train_x = self.train_x_bc
+        if config.parallel_scaling == "strong":
+            # Split the training points over each rank.
+            # We drop last points in order to have the same number of points per rank
+            train_x_all = self.train_x_all
+            train_x_all_shape = list(
+                self.train_x_all.shape
+            )  # We transform to list to support item assignment
+            num_split = train_x_all_shape[0] // config.world_size
```
lululxvi, line 195: What if `train_x_all_shape[0] // config.world_size == 0`?

pescap: In this case, all the points are split. For example, if `world_size=2`, `num_domain=10` and `num_boundary=10`, I obtain at `pde line 199`:

```
print(train_x_all_shape[0], num_split, train_x_all.shape[0])
```
as 
```
10 10 20
```

pescap: Hi, let me know if something is not correct.

lululxvi: Yes, if the number of points < world_size, then there is no need to split, i.e., all ranks have all points.

pescap: I raise an error in this case, as the simulation would be pointless with only one point per rank.

deepxde/data/pde.py
```diff
@@ -190,6 +185,23 @@ def train_next_batch(self, batch_size=None):
...
             config.comm.Bcast(self.train_x_bc, root=0)
         self.train_x = self.train_x_bc
+        if config.parallel_scaling == "strong":
+            # Split the training points over each rank.
+            # We drop last points in order to have the same number of points per rank
+            train_x_all_shape = list(
+                self.train_x_all.shape
+            )  # We transform to list to support item assignment
+            num_split = train_x_all_shape[0] // config.world_size
+            if num_split < config.world_size:
```
lululxvi, line 195: `if train_x_all_shape[0] < config.world_size`

pescap: sorry

deepxde/data/pde.py
```diff
@@ -190,6 +185,23 @@ def train_next_batch(self, batch_size=None):
                 self.train_x_bc = np.empty(x_bc_shape, dtype=self.train_x_bc.dtype)
             config.comm.Bcast(self.train_x_bc, root=0)
         self.train_x = self.train_x_bc
+        if config.parallel_scaling == "strong":
+            # Split the training points over each rank.
+            # We drop last points in order to have the same number of points per rank
+            train_x_all_shape = list(
```
lululxvi, line 191:
```
            if len(self.train_x_all) < config.world_size:
                raise ValueError(
                    "The number of training points is smaller than the number of processes. Please use more points."
                )
            train_x_all_shape = list(
               ...
```

PR #1298, opened 2023-05-05
Add split_in_rank

pescap:
As splitting for DeepONet uses the same exact function for strong scaling, I inspired myself from #1094 and added a `split_in_rank` function.

Therefore, I won't need to repeat the funcion in: 
https://github.com/lululxvi/deepxde/blob/a3cfd63e999f3c249f15e85026d191ed69d0bf0e/deepxde/data/triple.py#L79


deepxde/utils/internal.py
```diff
@@ -200,3 +200,27 @@ def get_num_args(func):
     # g = dummy(a.f)
     params = inspect.signature(func).parameters
     return len(params) - ("self" in params)
+
+
+def mpi_split_in_rank(array, drop_last=True):
+    """Split given array into continuous subarray according to world size and rank.
+    Args:
```
lululxvi, line 207: Add an empty line before `Arags`

deepxde/utils/internal.py
```diff
@@ -200,3 +200,27 @@ def get_num_args(func):
...
+        array (array or Tensor): Array to be split.
+        drop_last (bool): Whether to discard the remainder samples
+            not divisible by world_size. Default: True.
+    Returns:
```
lululxvi, line 211: Add an empty line before `Returns`

deepxde/utils/internal.py
```diff
@@ -200,3 +200,29 @@ def get_num_args(func):
...
+        array (array or Tensor): Array to be split.
+        drop_last (bool): Whether to discard the remainder samples
+            not divisible by world_size. Default: True.
+
+    Returns:
+        array or Tensor: Split array or Tensor.
+    """
+    # TODO: support drop_last=False
+    if not drop_last:
+        raise ValueError("Only support drop_last=True now.")
```
lululxvi, line 218: Add
```
    if config.world_size == 1:
        return array
```
so that this function itself is more robust.

deepxde/utils/internal.py
```diff
@@ -200,3 +200,29 @@ def get_num_args(func):
     # g = dummy(a.f)
     params = inspect.signature(func).parameters
     return len(params) - ("self" in params)
+
+
+def mpi_split_in_rank(array, drop_last=True):
+    """Split given array into continuous subarray according to world size and rank.
```
lululxvi, line 206: Scatter the given array into continuous subarrays of equal size from rank 0 to all ranks.

deepxde/utils/internal.py
```diff
@@ -200,3 +200,29 @@ def get_num_args(func):
     # g = dummy(a.f)
     params = inspect.signature(func).parameters
     return len(params) - ("self" in params)
+
+
+def mpi_split_in_rank(array, drop_last=True):
```
lululxvi, line 205:
```
def mpi_scatter_from_rank0(array, drop_last=True):
```

pescap: `def mpi_scatter_from_rank(array, drop_last=True, rank=0):`

deepxde/utils/internal.py
```diff
@@ -200,3 +200,31 @@ def get_num_args(func):
...
+        raise ValueError("Only support drop_last=True now.")    
+    if len(array) < config.world_size:
+        raise ValueError(
+            "The number of training points is smaller than the number of processes. Please use more points."
+        )
+    array_shape = list(array.shape)  # We transform to list to support item assignment
+    num_split = array_shape[0] // config.world_size
+    array_shape[0] = num_split
+    array_split = np.empty(array_shape, dtype=array.dtype)
+    config.comm.Scatter(array, array_split, root=0)
```
lululxvi, line 229: Seems a bug here, see example below. We need to truncate the `array` size to be a multiple of `num_split`.
```
import numpy as np
from mpi4py import MPI


comm = MPI.COMM_WORLD
size = comm.Get_size()
rank = comm.Get_rank()

sendbuf = None
if rank == 0:
    sendbuf = np.arange(7)
    print(sendbuf)

recvbuf = np.empty(2, dtype=int)

comm.Scatter(sendbuf, recvbuf, root=0)
print(recvbuf)
```

deepxde/utils/internal.py
```diff
@@ -200,3 +200,31 @@ def get_num_args(func):
     # g = dummy(a.f)
     params = inspect.signature(func).parameters
     return len(params) - ("self" in params)
+
+
+def mpi_scatter_from_rank0(array, drop_last=True):
+    """Scatter the given array into continuous subarrays of equal size from rank 0 to all ranks.
+
+    Args:
+        array (array or Tensor): Array to be split.
```
lululxvi, line 209: `array: Numpy array to be split.`

deepxde/utils/internal.py
```diff
@@ -200,3 +200,31 @@ def get_num_args(func):
...
+        drop_last (bool): Whether to discard the remainder samples
+            not divisible by world_size. Default: True.
+
+    Returns:
+        array or Tensor: Split array or Tensor.
```
lululxvi, line 214: Cannot be Tensor.

deepxde/utils/internal.py
```diff
@@ -200,3 +200,32 @@ def get_num_args(func):
...
+        raise ValueError("Only support drop_last=True now.")    
+    if len(array) < config.world_size:
+        raise ValueError(
+            "The number of training points is smaller than the number of processes. Please use more points."
+        )
+    array_shape = list(array.shape)  # We transform to list to support item assignment
+    num_split = array_shape[0] // config.world_size
+    array_shape[0] = num_split
+    array_split = np.empty(array_shape, dtype=array.dtype)
+    array = array[:num_split * config.world_size, :] # We truncate array size to be a multiple of num_split to prevent a MPI error.
```
lululxvi, line 229: `array = array[:num_split * config.world_size]`

deepxde/data/pde.py
```diff
@@ -185,23 +185,8 @@ def train_next_batch(self, batch_size=None):
...
-                self.train_x_all.shape
-            )  # We transform to list to support item assignment
-            num_split = train_x_all_shape[0] // config.world_size
-            train_x_all_shape[0] = num_split
-            train_x_all_split = np.empty(
-                train_x_all_shape, dtype=self.train_x_all.dtype
-            )
-            config.comm.Scatter(self.train_x_all, train_x_all_split, root=0)
-            self.train_x_all = train_x_all_split
+        if config.parallel_scaling == "strong" and config.hvd is not None:
```
lululxvi, line 188: `if config.parallel_scaling == "strong":`
This also supports other future parallel framework.

deepxde/utils/internal.py
lululxvi, line 229: Format via black.

Issue #1308, opened 2023-05-10
Physics-informed DeepONet Demo


WenhsiangSung:
Dear Prof. Lu,
I am a beginner of DeepONet. Is there any chance to see a demo of PI-DeepONet written using deepxde?
Thank you very much.

samuelburbulla:
Hi @WenhsiangSung, hi @lululxvi,

I was just working on such an example and would like to contribute it to the documentation!

Please refer to #1311.

The example trains a physics-informed DeepONet for a one-dimensional Poisson equation where the input to the operator is the source term. In my opinion, this is a MWE for such kind of problem.

Happy to hear about your opinion and any suggestions!

Best,
Samuel

WenhsiangSung:
Thx a lot! @samuelburbulla  
I will try this demo and give you some feedback!

PR #1311, opened 2023-05-10
Add PI-DeepONet example for 1D Poisson equation.

samuelburbulla:
This PR proposes an example for training a physics-informed DeepONet with DeepXDE.
The example solves a simple one-dimensional Poisson equation where the source term is the input to the solution operator.

Related to #1308.


samuelburbulla:
I improved this PR a little more:

- Incorporate some fixes for using PyTorch backend
- Add auxiliary_variables in PyTorch.NN
- Use L-BFGS for training the DeepONet

Here's a plot of the result:

pideeponet_poisson1d

deepxde/data/pde_operator.py
```diff
@@ -275,10 +276,12 @@ def losses_test(self, targets, outputs, loss_fn, inputs, model, aux=None):
     def train_next_batch(self, batch_size=None):
         if self.train_x is None:
             func_feats = self.func_space.random(self.num_func)
-            func_vals = self.func_space.eval_batch(func_feats, self.eval_pts)
+            func_vals = self.func_space.eval_batch(
+                func_feats, self.eval_pts
+            ).astype(config.real(np))
```
lululxvi, line 281: Don't change this. `self.func_space.eval_batch` should return the correct `dtype`; if not, then should modify `self.func_space.eval_batch` or the inputs `func_feats` and `self.eval_pts`.

deepxde/data/pde_operator.py
```diff
@@ -275,10 +276,12 @@ def losses_test(self, targets, outputs, loss_fn, inputs, model, aux=None):
...
             vx = self.func_space.eval_batch(
                 func_feats, self.pde.train_x[:, self.func_vars]
-            )
+            ).astype(config.real(np))
```
lululxvi, line 284: The same as above.

deepxde/data/pde_operator.py
```diff
@@ -262,7 +262,8 @@ def _losses(self, outputs, loss_fn, inputs, model, num_func):
             losses.append(losses_i)
 
         losses = zip(*losses)
-        losses = [bkd.reduce_mean(bkd.as_tensor(l)) for l in losses]
+        # Reduce mean without loosing derivatives
+        losses = [sum(l) / len(l) for l in losses]
```
lululxvi, line 266: What is the difference between `reduce_mean` and `sum / len`?

lululxvi:
I suggest splitting this PR into two separate PRs
1. the demo example code;
2. Pytorch backed code.

samuelburbulla:
Thanks, @lululxvi, for your review!

This PR is now split into:
- #1311 (this one) including the example only
- #1318 including the PyTorch backend modifications

I have found better solutions for the points mentioned, see comments above.

deepxde/data/pde_operator.py
samuelburbulla: Okay, I added `@to_tensor` decorators to the function space `eval_batch` functions.

deepxde/data/pde_operator.py
samuelburbulla: The issue is that `bkd.tensor(l)` looses gradients of `l` for PyTorch, due to the use of `torch.to_tensor` for a tuple of tensors.

I proposed an adaptation of `torch.tensor()` that uses `torch.stack` for list/tuple of tensors. Then, this line can stay as it is.

sibyjackgrove:
@samuelburbulla Thanks for making this very useful example. @lululxvi would you be merging this soon?

deepxde/data/function_spaces.py
```diff
@@ -90,6 +91,7 @@ def random(self, size):
     def eval_one(self, feature, x):
         return np.dot(feature, x ** np.arange(self.N))
 
+    @return_tensor
```
lululxvi, line 94: Don't return tensor. This file is designed to be pure Python/Numpy.

lululxvi:
Which backends does the new code support? The existing code supports tensorflow 1.x.

samuelburbulla:
> Which backends does the new code support? The existing code supports tensorflow 1.x.

This example is tested with tensorflow.compat.v1, tensorflow, and pytorch.

deepxde/model.py
```diff
@@ -281,7 +281,8 @@ def outputs(training, inputs):
             return self.net(inputs)
 
         def outputs_losses(training, inputs, targets, auxiliary_vars, losses_fn):
-            self.net.auxiliary_vars = auxiliary_vars
+            if auxiliary_vars is not None:
+                self.net.auxiliary_vars = torch.as_tensor(auxiliary_vars).requires_grad_()
```
lululxvi, line 285: Why you need the grad of auxiliary_vars?

lululxvi:
I have added a few PI-DeepONet examples, see https://deepxde.readthedocs.io/en/latest/demos/operator.html#pi-deeponet

Could you split this PR into two PRs:
1. only modify the source code;
2. add new example

samuelburbulla:
> I have added a few PI-DeepONet examples

Nice!

> Could you split this PR into two Pls

Done, see #1379.

samuelburbulla:
Rebased on master.

Together with #1403, this example works with all mentioned backends.

Is there anything else that needs to be done here?

examples/operator/pideeponet_poisson1d.py
```diff
@@ -0,0 +1,102 @@
+"""Backend supported: tensorflow.compat.v1, tensorflow, pytorch"""
+import numpy as np
+import deepxde as dde
+import matplotlib.pyplot as plt
```
lululxvi, line 4:
```
import deepxde as dde
import numpy as np
```

examples/operator/pideeponet_poisson1d.py
```diff
@@ -0,0 +1,102 @@
...
+    [num_eval_points, 32, p],
+    [dim_x, 32, p],
+    activation="tanh",
+    kernel_initializer="Glorot normal",
+)
+
+# Define and train model
+model = dde.Model(pde_op, net)
+
+if dde.backend.backend_name == "pytorch":
```
lululxvi, line 66: Why different optimizers for different backends?

lululxvi:
Reformat via black https://pypi.org/project/black/

docs/demos/operator/poisson_1d_pideeponet.rst
```diff
@@ -0,0 +1,114 @@
+Physics-informed DeepONet for Poisson equation in 1D
```
lululxvi, line 1: Just "Poisson equation in 1D"

PR #1315, opened 2023-05-12
update backend.paddle.tensor.Variable

lijialin03:
update backend.paddle.tensor.Variable to make sure shape of the method's output is [] because PaddlePaddle-develop can support 0D Tensor after 2023.05.10 version.

### DIFF
--- a/deepxde/backend/paddle/tensor.py
+++ b/deepxde/backend/paddle/tensor.py
@@ -59,6 +59,8 @@ def reshape(tensor, shape):
 
 
 def Variable(initial_value, dtype=None):
+    if paddle.in_dynamic_mode():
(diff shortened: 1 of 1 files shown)


deepxde/backend/paddle/tensor.py
```diff
@@ -59,6 +59,8 @@ def reshape(tensor, shape):
 
 
 def Variable(initial_value, dtype=None):
+    if paddle.in_dynamic_mode():
```
lululxvi, line 62: This is only supported in dynamic mode,  not static mode?

lijialin03: Yes, static mode and dynamic-to-static mode are being optimized now and will be supported in the future.

PR #1318, opened 2023-05-15
Add auxiliary_variables for pytorch.

samuelburbulla:
Adds auxiliary_variables to `pytorch.nn`.

~~Includes a fix in `backend.pytorch.tensor.to_tensor`:
Use `torch.stack` instead of `torch.to_tensor` for list/tuple of `torch.Tensors` to keep track of gradients.~~

Related to #1311.

### DIFF
--- a/deepxde/model.py
+++ b/deepxde/model.py
@@ -280,7 +280,8 @@ class Model:
             grad.clear()
             return self.net(inputs)
 
-        def outputs_losses(training, inputs, targets, losses_fn):
+        def outputs_losses(training, inputs, targets, auxiliary_vars, losses_fn):
+            self.net.auxiliary_vars = auxiliary_vars
             self.net.train(mode=training)
             if isinstance(inputs, tuple):
                 inputs = tuple(
@@ -304,11 +305,15 @@ class Model:
             grad.clear()
             return outputs_, losses
 
-        def outputs_losses_train(inputs, targets):
-            return outputs_losses(True, inputs, targets, self.data.losses_train)
+        def outputs_losses_train(inputs, targets, auxiliary_vars):
+            return outputs_losses(
+                True, inputs, targets, auxiliary_vars, self.data.losses_train
+            )
 
-        def outputs_losses_test(inputs, targets):
-            return outputs_losses(False, inputs, targets, self.data.losses_test)
+        def outputs_losses_test(inputs, targets, auxiliary_vars):
+            return outputs_losses(
+                False, inputs, targets, auxiliary_vars, self.data.losses_test
+            )
 
         # Another way is using per-parameter options
         # https://pytorch.org/docs/stable/optim.html#per-parameter-options,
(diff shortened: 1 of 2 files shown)


deepxde/backend/pytorch/tensor.py
```diff
@@ -71,6 +71,8 @@ def Variable(initial_value, dtype=None):
 
 
 def as_tensor(data, dtype=None):
+    if isinstance(data, (list, tuple)) and all(isinstance(d, torch.Tensor) for d in data):
```
lululxvi, line 74: Why do we need this? According to https://pytorch.org/docs/stable/generated/torch.as_tensor.html, `torch.as_tensor` supports list and tuple.

samuelburbulla: It supports it, but it looses the gradient information.

Consider this minimal example:

````
import torch

a = torch.tensor([1.], requires_grad=True)
b = torch.tensor([2.], requires_grad=True)

# as_tensor
ab = torch.as_tensor((a, b))
ab.requires_grad = True
ab.sum().backward()
print(f"a.grad = {a.grad}, b.grad = {b.grad}")

# stack
ab = torch.stack((a, b))
ab.sum().backward()
print(f"a.grad = {a.grad}, b.grad = {b.grad}")
````

Output:
````
a.grad = None, b.grad = None
a.grad = tensor([1.]), b.grad = tensor([1.])
````

lululxvi: Good point.  That is why we use `torch.stack` here
https://github.com/lululxvi/deepxde/blob/63a0e1ba345f992fbe7653a4e190a771d09f2659/deepxde/model.py#L299

But `as_tensor` is designed to convert the data to a Tensor, see the docstring below. In which case, do we need to support gradient information?

https://github.com/lululxvi/deepxde/blob/63a0e1ba345f992fbe7653a4e190a771d09f2659/deepxde/backend/backend.py#L143

samuelburbulla: The line emerged when I built the example in #1311. As I considered this a general pitfall (TensorFlow behaves differently, for instance) I introduced this here.

Of course, I see this might imply undesirable side effects - and I do not oversee all consequences as you do!
If you don’t like to introduce this as a general fix, I propose to remove this line from the PR and discuss a possible solution in #1311.


Some details about the issue that occured: There was a dtype mismatch in `deepxde/deepxde/nn/pytorch/fnn.py:43`. Therefore, I added `@return_tensor` decorators to the `eval_batch` functions in `function_spaces.py`.
When adding these decorators, the gradient information is lost as the wrapper calls `bkd.as_tensor`. That was supposed to be fixed (in general) by adding the stack line in `as_tensor` here.

lululxvi: Let us remove this line from the PR and discuss a possible solution in https://github.com/lululxvi/deepxde/pull/1311.

samuelburbulla: Removed the line.

Issue #1244, opened 2023-04-13
Enhancement: add tests, pre-commit, and coverage support


Saransh-cpp:
This is a placeholder issue for testing related developments of DeepXDE. Things I am planning to work on in the next couple of months -

- [ ] Add pre-commit support for better linting, formatting, and readability of the codebase
- [x] Add integration tests (test all examples)
- [x] ~Add basic testing infrastructure using `pytest`~
- [x] ~Add basic coverage infrastructure using `pytest-cov`~
- [x] ~Add [Codecov](https://about.codecov.io) integration for better coverage statistics on GitHub~
- [x] ~Add more and more tests to bring the coverage value up~
- [x] ~Add docs about testing and encourage developers to add a test with every new feature added to DeepXDE~

pescap:
Happy to help! This would greatly improve the package!

Saransh-cpp:
Thanks @pescap! Apologies for abandoning this. I'll start some discussions here next week, along with finishing up the open PR.

Saransh-cpp:
I think this can now be closed. The unchecked point aimed at adding automatic formatters, static code checkers, and linters using [pre-commit](https://pre-commit.com) and [pre-commit.ci](https://pre-commit.ci) in DeepXDE's CI, but I would wait for everyone's opinion on that. I'll be happy to add them to the CI if they are needed!

lululxvi:
Sounds good. I don't have experience on [pre-commit](https://pre-commit.com/) and [pre-commit.ci](https://pre-commit.ci/). So feel free to close.

PR #1324, opened 2023-05-22
Create an integration testing suite

Saransh-cpp:
Xref: #1244

~This PR adds a minimal testing suite to DeepXDE. There are a few options available for testing libraries in Python and the most widely used is `pytest`. `pytest` will run the files in the `tests/` directory (starting with `test_`) and validate that every `assert` statement passes.~

~The tests check that the code works as intended, including the returned values, datatypes, and if the warnings and errors are raised correctly. This way, if someone now modifies one of the loss functions and ends up messing up with them, the tests will fail indicating that something wrong has gone with the particular loss function.~

~I have currently only implemented tests for the loss function for every backend. Some of the functions are only supported on particular backends, so they are skipped while other backends are tested. I've tried my best to include edge cases and make the tests reliable, but please let me know if I can add more tests for these functions.~

~These tests now also execute in the CI, which means every they will be executed on every PR. DeepXDE is already imported within the test module; hence, no need to explicitly test the import in the CI.~

~I have also added some minimal configuration for `pytest` in `pyproject.toml` and added it as a `dev` dependency, so that developers can install `pytest` with `deepxde` using `pip install "deepxde[dev]"`. More `dev` dependencies will be added as we move further (including `pytest-cov` and `pre-commit`).~

~Finally, looking at https://github.com/lululxvi/deepxde/discussions/1313#discussioncomment-5879758, all the tests of a particular backend only run with the Tensor or vector of that particular backend.~

---

~Once this is merged, the tests can be populated for other DeepXDE modules. I will add support for coverage after this so that we can visualise which modules are not tested DeepXDE (all of them right now, but that will change and coverage will catch that).~

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ concurrency:
   cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
 
 jobs:
-  backend_tensorflow_v1:
+  build:
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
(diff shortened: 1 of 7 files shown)


Saransh-cpp:
Sorry for sitting on this for too long, this should be ready for a review!

lululxvi:
Thank you for this PR. It definitely makes important contributions to the testing. However, my concern is DeepXDE now has so many codes and functionalities, so it is really difficult to cover all the tests. It is just too much work.

My current thought is that we don't need to test detailed functions such as losses, and instead, we only test those example codes. If the examples work OK, then it should be fine. Any suggestions?

Saransh-cpp:
Oh yes, that sounds much better. Thanks for the review and the suggestion! I'll add tests for the notebooks and revamp this PR.

examples/sample_to_test.py
```diff
@@ -45,5 +47,39 @@ def transform(line, file_name):
...
+            "fractional_Poisson_1d.py",
+            "fractional_Poisson_2d.py",
+            "fractional_Poisson_3d.py",
+            "ide.py",
+            "diffusion_1d_inverse.py",
+            "fractional_Poisson_1d_inverse.py",
+            "fractional_Poisson_2d_inverse.py",
+            "Poisson_Lshape.py",
+            "ode_system.py",
+            "Lorenz_inverse.py",
```
Saransh-cpp, line 76: How should I handle these examples? They are not being tested at the moment as backend-specific code is written as comments in these files. One way to handle this would be how some examples like `operator/advection_aligned_pideeponet_2d.py` do -

https://github.com/lululxvi/deepxde/blob/c807b4bff2e218fb094574c2b21d8d5f15c5292b/examples/operator/advection_aligned_pideeponet_2d.py#L6-L26

but I think it would be best to work on them in further PRs, given that this PR has become huge.

Saransh-cpp (commented):
This should be ready for review now.

This PR now updates and fixes the already existing `Makefile` in `examples/` to run all the example files using `make run_all_examples`. I've added the same command in the CI and have parallelized the builds on different backends to save a ton of CI time. I also had to make some minor fixes in the examples to get everything working, but I have ensured that major example changes are not present in this PR.

examples/dataset/antiderivative_aligned_test.npz
Saransh-cpp, line 1: I am not sure if I should add these datasets in or if I should simply exclude these examples from testing?

lululxvi:
Yes, this PR is huge. Could you first make a PR that changes `epochs` to `iterations` in `model.train()`? So I can merge it immediately.

Saransh-cpp:
Oh yes, I can split this PR.

examples/pinn_forward/Allen_Cahn.py
```diff
@@ -2,6 +2,7 @@
 
 Implementation of Allen-Cahn equation example in paper https://arxiv.org/abs/2111.02801.
 """
+import os
```
Saransh-cpp, line 5:
```suggestion
```

examples/dataset/antiderivative_unaligned_test.npz
lululxvi, line 1: I prefer not to add these large data files in GitHub repo. It is OK not to test all examples.

Saransh-cpp: Thanks for the review! I have removed the data files from every commit and squashed the changes into 1. Everything should work now.

deepxde/config.py
```diff
@@ -1,4 +1,5 @@
 import os
+import sys
 import random
```
lululxvi, line 3:
```
import os
import random
import sys
```

examples/sample_to_test.py
```diff
@@ -9,9 +9,10 @@
 
 import re
 import sys
+import deepxde as dde
```
lululxvi, line 12:
```
import re
import sys

import deepxde as dde
```

.github/workflows/build.yml
```diff
@@ -14,12 +14,12 @@ concurrency:
   cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
 
 jobs:
-  build:
+  build_with_tensorflow_v1:
```
lululxvi, line 17: `build` is already in the name. so use `backend_tensorflow_v1`, etc.

lululxvi:
- The current build tests many examples, which takes a long time to test. Also, more examples will be added in the future.
- On another hand, when we add more examples or revise existing examples, we may change the sample_to_test.py, which is OK but requires more work.

So, I would suggest still only testing import. If we need to test the examples, we can run the Makefile manually. What do you think?

Saransh-cpp:
Yes, the extra time required to run all the examples is definitely not good, and I have been worried about that since the start of my work on this (which is why I decided to run all jobs in parallel, but there are only a limited number of machines available from GitHub, making this not "completely" parallel).

Testing locally sounds great. I can periodically check if the Makefile is working locally and fix it if it goes stale.

I will revert the changes done to the workflow file and test import on all backends sequentially (parallelly on OS and Python versions).

.github/workflows/build.yml
```diff
@@ -14,12 +14,12 @@ concurrency:
   cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
 
 jobs:
-  build:
+  deepxde:
```
lululxvi, line 17: Use `import` so it is more clear.

PR #1350, opened 2023-06-19
some changes about readthedocs

lijialin03:
1. add paddle part in Makefile
2. change needed paddle version to 'paddlepaddle==2.4.2' temporarily.


lululxvi:
I prefer to use Ubuntu 22.04. Let us wait.

lijialin03:
> I prefer to use Ubuntu 22.04. Let us wait.
The reason for the problem is openssl 3.x which is used in Ubuntu22 is not supported by the newest paddle develop now and will be supported in the future.
Since the openssl used in readthedocs cannot be downgraded, installing paddle 2.4.2 for generating docs now is an available solutions. Thanks.

Issue #1376, opened 2023-07-08
Fractional Poisson equation in 2D produces errors


xuelanghanbao:
I tried to run the Fractional Poisson equation in 2D demo, but it produces the following error. 
```
Traceback (most recent call last):
  File "deepxde\fractional_Poisson_2d.py", line 35, in <module>
    data = dde.data.FPDE(
  File "deepxde\deepxde\data\fpde.py", line 88, in __init__
    super().__init__(
  File "deepxde\deepxde\data\pde.py", line 127, in __init__
    self.train_next_batch()
  File "deepxde\deepxde\data\fpde.py", line 163, in train_next_batch
    X = self.frac_train.get_x()
  File "deepxde\deepxde\data\fpde.py", line 402, in get_x
    else self.get_x_dynamic()
  File "deepxde\deepxde\data\fpde.py", line 474, in get_x_dynamic
    self.w.append(array_ops_compat.hstack(wi))
  File "deepxde\deepxde\utils\array_ops_compat.py", line 25, in hstack
    if not is_tensor(tup[0]) and tup[0] == []:
ValueError: operands could not be broadcast together with shapes (172,) (0,)
```

I just changed the tf backend to deepxde.backend, as shown below:

```
import deepxde as dde
import numpy as np
import deepxde.backend as bkd
from scipy.special import gamma

alpha = 1.8

# Backend tensorflow.compat.v1
def fpde(x, y, int_mat):
    """\int_theta D_theta^alpha u(x)"""
    if isinstance(int_mat, (list, tuple)) and len(int_mat) == 3:
        int_mat = bkd.sparse_tensor(*int_mat)
        lhs = bkd.sparse_dense_matmul(int_mat, y)
    else:
        lhs = bkd.matmul(bkd.from_numpy(int_mat), bkd.from_numpy(y))
    lhs = lhs[:, 0]
    lhs *= gamma((1 - alpha) / 2) * gamma((2 + alpha) / 2) / (2 * np.pi ** 1.5)
    x = x[: bkd.size(lhs)]
    rhs = (
        2 ** alpha
        * gamma(2 + alpha / 2)
        * gamma(1 + alpha / 2)
        * (1 - (1 + alpha / 2) * bkd.from_numpy(np.sum( bkd.to_numpy(x ** 2),axis=1)))
    )
    return lhs - rhs

def func(x):
    return (np.abs(1 - np.linalg.norm(x, axis=1, keepdims=True) ** 2)) ** (
[...]

PR #1377, opened 2023-07-08
use isinstance() to determine type rather than []

xuelanghanbao:
#1376 
Use `isinstance()` to determine `tup[0]` type rather than `tup[0]==[]`. When `tup` is passed in from `self.w.append(array_ops_compat.hstack(wi))` in `deepxde\data\fpde.py, line 474`, `tup[0]==[]` can cause the following errors: `ValueError: operands could not be broadcast together with shapes (172,) (0,)`


lululxvi:
Good point. But here we want to check if it is a empty list, not only checking if it is a list.

xuelanghanbao:
Thank you for your explanation. I have a better understanding of the purpose of this code. So changing `not is_tensor(tup[0])` to `isinstance(tup[0],list)` would be better, right?

lululxvi:
> changing `not is_tensor(tup[0])` to `isinstance(tup[0],list)` would be better, right?

Seems OK, but I am not 100% sure. It would be safer to use both.

deepxde/utils/array_ops_compat.py
```diff
@@ -22,7 +22,7 @@ def convert_to_array(value):
 
 
 def hstack(tup):
-    if not is_tensor(tup[0]) and tup[0] == []:
+    if not is_tensor(tup[0]) and isinstance(tup[0],list) and tup[0] == []:
```
lululxvi, line 25: Format via black https://pypi.org/project/black/

PR #1379, opened 2023-07-11
Feature/pde operator backend

samuelburbulla:
Backend changes of #1311.

PR #1388, opened 2023-07-18
Fixed a bug in `pde_operator.py` that bkd.as_tensor() would make the gradient disappear.

supercgor:
Using `PyTorch` as the backend, running some examples codes would raise errors:
`RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn`

- The problem is `bkd.as_tensor(array_like)` would create a new tensor, and the loss would detach from the graph, thus unable to do gradient descent. 

- Not sure if this change would affect `TensorFlow` or other backends.

### DIFF
--- a/deepxde/backend/tensorflow/tensor.py
+++ b/deepxde/backend/tensorflow/tensor.py
@@ -75,6 +75,30 @@ def to_numpy(input_tensor):
     return input_tensor.numpy()
 
 
+def concat(values, axis):
+    return tf.concat(values, axis)
+
+
+def stack(values, axis):
+    return tf.stack(values, axis)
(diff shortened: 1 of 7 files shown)


samuelburbulla:
Already tackled in #1379. Additional backend changes for TensorFlow necessary.

lululxvi:
Please resolve the conflict. If now pytorch works for PI-DeepONet, then update the supported backends (the first line of each file):

https://github.com/lululxvi/deepxde/blob/23bba4f983d57eb971759da47a9658d4f6238312/examples/operator/antiderivative_aligned_pideeponet.py#L1

PR #1391, opened 2023-07-20
Bug fix, add gelu for torch, add lambda lr scheduler for torch, allow `predict` to handle CartesianProd` data for torch, automatically change the dataset's datatype for torch.

supercgor:
1. In torch, the `predict` doesn't accept tuple input, that is a bug.
2. The second thing is for `CartesianProd`, `predict` can not deal with the differential equation `operator`. So added some code for torch to resolve it.
3. automatically change the data type for torch training.

* add torch gelu

* add lambda lr scheduler for torch

* repair missing deeponet for pytorch

### DIFF
--- a/deepxde/backend/backend.py
+++ b/deepxde/backend/backend.py
@@ -274,6 +274,8 @@ def elu(x):
 def relu(x):
     """Applies the rectified linear unit activation function."""
 
+def gelu(x):
+    """Computes Gaussian Error Linear Unit."""
 
 def selu(x):
     """Computes scaled exponential linear."""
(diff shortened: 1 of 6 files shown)


deepxde/backend/backend.py
```diff
@@ -274,6 +274,8 @@ def elu(x):
 def relu(x):
     """Applies the rectified linear unit activation function."""
```
lululxvi, line 276: Two empty lines

lululxvi:
Split the PR into a few independent ones.

PR #1395, opened 2023-07-21
fix bug for tuple inputs in `predict` for torch

supercgor:
For tuple inputs (DeepONet), this function is not work.

### DIFF
--- a/deepxde/model.py
+++ b/deepxde/model.py
@@ -911,8 +911,10 @@ class Model:
             y = utils.to_numpy(y)
         elif backend_name == "pytorch":
             self.net.eval()
-            inputs = torch.as_tensor(x)
-            inputs.requires_grad_()
+            if isinstance(x, tuple):
+                inputs = tuple(map(lambda x: torch.as_tensor(x).requires_grad_(), x))
+            else:
+                inputs = torch.as_tensor(x).requires_grad_()
             outputs = self.net(inputs)
             if utils.get_num_args(operator) == 2:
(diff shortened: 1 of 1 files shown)


supercgor:
Two more problems after this merge. The first is do we need to automatically convert the numpy dtype to the same as the network use? Because numpy default floating point is `np.float64`. But for torch the default floating point is `torch.float32`.
The second is `predict` is not support pde operator if the data is formatted as `CartesianProd`. The autograd is not working. I will then try to fix it.

lululxvi:
> The first is do we need to automatically convert the numpy dtype to the same as the network use?

It is already supported.

https://github.com/lululxvi/deepxde/blob/98697ffeaa1351f81448ea30bbca0e66b86dee41/deepxde/model.py#L870

> The second is `predict` is not support pde operator if the data is formatted as `CartesianProd`. The autograd is not working. I will then try to fix it.

Good.

supercgor:
The `train` part didn't support it now. If the train data is `np.float64`, for torch, it can't convert to `torch.float32` and throw an error. I wonder where should I change the dtype? For each backend in the `model`, in the `TrainState`, or for each `data` class.

lululxvi:
All the data should be converted to the correct dtype at the first place when they are generated.

https://github.com/lululxvi/deepxde/blob/411ca5c4c77728181a2668af1aa11247b02f4abe/deepxde/config.py#L41

Trace back to the first place where the dtype is not correct and fix it.

PR #1396, opened 2023-07-21
repair missing deeponet for pytorch

supercgor:


deepxde/nn/pytorch/__init__.py
```diff
@@ -8,9 +8,10 @@
     "PFNN",
     "PODDeepONet",
     "PODMIONet",
+    "DeepONet",
```
lululxvi, line 11: Reorder. see https://github.com/lululxvi/deepxde/blob/master/deepxde/nn/tensorflow_compat_v1/__init__.py

PR #1403, opened 2023-07-25
Return correct real type in eval_batch.

samuelburbulla:
Return numpy arrays with the real type defined by `config.real(np)` in function spaces.

Apply the approach which was already taken in `GRF` to all `eval_batch` methods.

Fixes the type mismatches mentioned in #1311.

### DIFF
--- a/deepxde/data/function_spaces.py
+++ b/deepxde/data/function_spaces.py
@@ -85,16 +85,16 @@ class PowerSeries(FunctionSpace):
         self.M = M
 
     def random(self, size):
-        return 2 * self.M * np.random.rand(size, self.N) - self.M
+        return 2 * self.M * np.random.rand(size, self.N).astype(config.real(np)) - self.M
 
     def eval_one(self, feature, x):
         return np.dot(feature, x ** np.arange(self.N))
 
     def eval_batch(self, features, xs):
-        mat = np.ones((self.N, len(xs)))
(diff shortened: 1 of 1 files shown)


deepxde/data/function_spaces.py
```diff
@@ -94,7 +94,7 @@ def eval_batch(self, features, xs):
         mat = np.ones((self.N, len(xs)))
         for i in range(1, self.N):
             mat[i] = np.ravel(xs**i)
-        return np.dot(features, mat)
+        return np.dot(features, mat).astype(config.real(np))
```
lululxvi, line 97: Don't change this line. Change line 94 to
```
mat = np.ones((self.N, len(xs)), dtype=config.real(np))
```

deepxde/data/function_spaces.py
```diff
@@ -226,7 +226,7 @@ def eval_one(self, feature, x):
 
     def eval_batch(self, features, xs):
         eigfun = np.array([np.ravel(f(xs)) for f in self.eigfun])
-        return np.dot(features, eigfun)
+        return np.dot(features, eigfun).astype(config.real(np))
```
lululxvi, line 229: Same here. Use `dtype` in line 228.

deepxde/data/function_spaces.py
samuelburbulla: Okay, I also need to adjust the dtype of `features` then.

PR #1424, opened 2023-08-09
Backend paddle: Support batch_size

lijialin03:
1. 'batch_size' also can be supported by paddlepaddle. So update boundary_conditions.py to support elliptic_inverse_field_batch example. The training results are as follows, attention, time may not be the shortest time becuase of GPU usage.
paddle backend
image
pytorch backend
image

### DIFF
--- a/deepxde/icbc/boundary_conditions.py
+++ b/deepxde/icbc/boundary_conditions.py
@@ -175,13 +175,13 @@ class PointSetBC:
         component: Integer or a list of integers. The output components satisfying this BC.
             List of integers only supported for the backend PyTorch.
         batch_size: The number of points per minibatch, or `None` to return all points.
-            This is only supported for the backend PyTorch.
+            This is only supported for the backend PyTorch and PaddlePaddle.
+            Note, If you want to use batch size here, you should also set callback
+            'dde.callbacks.PDEPointResampler(bc_points=True)' in training.
         shuffle: Randomize the order on each pass through the data when batching.
     """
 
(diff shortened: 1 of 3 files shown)


lululxvi:
Just to confirm, you have checked that the collation points change every iteration, right?

lijialin03:
> Just to confirm, you have checked that the collation points change every iteration, right?

I found that collocation_points() cannot be called during training in original code even in pytorch backend, which means collation points didn't change. I checked code and found that should send 'bc_points=True' to dde.callbacks.PDEPointResampler() in elliptic_inverse_field_batch.py to call collocation_points()
But I think it is a little bit strange for me that using 'batch size' by setting this kind of special callback. Maybe other ways would be better？

lululxvi:
Good point.

deepxde/icbc/boundary_conditions.py
```diff
@@ -175,13 +175,11 @@ class PointSetBC:
         component: Integer or a list of integers. The output components satisfying this BC.
             List of integers only supported for the backend PyTorch.
         batch_size: The number of points per minibatch, or `None` to return all points.
-            This is only supported for the backend PyTorch.
+            This is only supported for the backend PyTorch and PaddlePaddle.
```
lululxvi, line 178: Add a comment here for dde.callbacks.PDEPointResampler()

lululxvi:
Also modify the docs of `PDEPointResampler` to add the support of paddle

PR #1443, opened 2023-08-17
Support example torch

supercgor:


supercgor:
support torch for operator examples

Issue #1446, opened 2023-08-18
An error of example/pinn_forward/Helmholtz_Dirichlet_2d_HPO.py


lijialin03:
### Discussed in https://github.com/lululxvi/deepxde/discussions/1445

<div type='discussions-op-text'>

<sup>Originally posted by **lijialin03** August 18, 2023</sup>
Function 'gp_minimize' of package 'skopt (scikit-optimize)' is used in this example and an error has been reported like,
![16e8051a98747fcc62553d09c11b873b](https://github.com/lululxvi/deepxde/assets/124568209/a4c21d4a-4170-4fd1-a7f0-20c038522b20)
The reason is that 'np.int' used in skopt was deprecated in NumPy 1.20. The newest NumPy is version 1.25.
My skopt version is 0.9.0. I think it is the latest version in https://pypi.org/search/?q=scikit-optimize. So this example may require NumPy 1.20 or lower while there is no version requirements of NumPy in requirements.txt.</div>

lululxvi:
How about fixing this using the solutions at https://github.com/scikit-optimize/scikit-optimize/issues/1138

lijialin03:
> How about fixing this using the solutions at [scikit-optimize/scikit-optimize#1138](https://github.com/scikit-optimize/scikit-optimize/issues/1138)

Yes, the solution do works. I submit a PR here https://github.com/lululxvi/deepxde/pull/1450 to apply it.

PR #1450, opened 2023-08-21
Fix version mismatch between scikit-optimize and Numpy

lijialin03:
To fix version mismatch between scikit-optimize and Numpy of example Helmholtz_Dirichlet_2d_HPO.py https://github.com/lululxvi/deepxde/issues/1446 accroding to https://github.com/scikit-optimize/scikit-optimize/issues/1138.


examples/pinn_forward/Helmholtz_Dirichlet_2d_HPO.py
```diff
@@ -9,6 +9,8 @@
 from skopt.space import Real, Categorical, Integer
 from skopt.utils import use_named_args
 
+np.int = int
```
lululxvi, line 12: Add a comment here for the reason of adding this line.

PR #1453, opened 2023-08-21
Add typing and type annotations for icbc 

hheei:
creating an extra file called `types.py`, referred from torch.

### DIFF
--- a/deepxde/backend/backend.py
+++ b/deepxde/backend/backend.py
@@ -1,3 +1,9 @@
+from __future__ import annotations
+
+from numbers import Number
+from typing import Sequence, overload
+from ..types import Tensor, dtype, SparseTensor, TensorOrTensors
+from numpy.typing import NDArray, ArrayLike
 """This file defines the unified tensor framework interface required by DeepXDE.
 
 The principles of this interface:
@@ -26,7 +32,7 @@ paddle = None
 # Tensor, data type and context interfaces
 
 
-def data_type_dict():
+def data_type_dict() -> dict[str, object]:
     """Returns a dictionary from data type string to the data type.
 
     The dictionary should include at least:
@@ -58,7 +64,7 @@ def data_type_dict():
     """
 
 
-def is_gpu_available():
+def is_gpu_available() -> bool:
     """Returns a bool indicating if GPU is currently available.
 
     Returns:
@@ -66,11 +72,11 @@ def is_gpu_available():
     """
 
 
-def is_tensor(obj):
+def is_tensor(obj: object) -> bool:
     """Returns True if `obj` is a backend-native type tensor."""
 
 
-def shape(input_tensor):
+def shape(input_tensor: Tensor) -> Sequence[int]:
     """Return the shape of the tensor.
 
     Args:
@@ -81,7 +87,7 @@ def shape(input_tensor):
     """
 
 
-def size(input_tensor):
+def size(input_tensor: Tensor) -> int:
     """Return the total number of elements in the input tensor.
 
     Args:
@@ -92,7 +98,7 @@ def size(input_tensor):
     """
 
 
-def ndim(input_tensor):
+def ndim(input_tensor: Tensor) -> int:
     """Returns the number of dimensions of the tensor.
 
     Args:
@@ -103,7 +109,7 @@ def ndim(input_tensor):
     """
 
 
-def transpose(tensor, axes=None):
+def transpose(tensor: Tensor, axes: Sequence[int] | int | None = None) -> Tensor:
     """Reverse or permute the axes of a tensor; returns the modified array.
 
     For a tensor with two axes, transpose gives the matrix transpose.
@@ -117,7 +123,7 @@ def transpose(tensor, axes=None):
     """
 
 
-def reshape(tensor, shape):
+def reshape(tensor: Tensor, shape: Sequence[int]) -> Tensor:
     """Gives a new shape to a tensor without changing its data.
 
     Args:
@@ -130,7 +136,7 @@ def reshape(tensor, shape):
     """
 
 
-def Variable(initial_value, dtype=None):
+def Variable(initial_value: Number, dtype: dtype = None) -> Tensor:
     """Return a trainable variable.
 
     Args:
@@ -140,7 +146,7 @@ def Variable(initial_value, dtype=None):
     """
 
 
-def as_tensor(data, dtype=None):
+def as_tensor(data: ArrayLike, dtype: dtype = None) -> Tensor:
     """Convert the data to a Tensor.
 
     If the data is already a tensor and has the same dtype, directly return.
@@ -155,7 +161,7 @@ def as_tensor(data, dtype=None):
     """
 
 
-def sparse_tensor(indices, values, shape):
+def sparse_tensor(indices: Sequence[Sequence[Number, Number]], values: Tensor, shape: Sequence[int]) -> SparseTensor:
     """Construct a sparse tensor based on given indices, values and shape.
 
     Args:
@@ -170,7 +176,7 @@ def sparse_tensor(indices, values, shape):
     """
 
 
-def from_numpy(np_array):
+def from_numpy(np_array: NDArray) -> Tensor:
     """Create a tensor that shares the underlying numpy array memory, if possible.
 
     Args:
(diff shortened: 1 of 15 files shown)


deepxde/icbc/boundary_conditions.py
```diff
@@ -14,27 +14,37 @@
 import numbers
 from abc import ABC, abstractmethod
 from functools import wraps
+from typing import Any, Callable, List, Optional, overload, Tuple, Union
```
Saransh-cpp, line 17: Given that `DeepXDE` supports Python 3.8+, you should use -
```py
from __future__ import annotations
from typing import Any, Callable, Optional, overload
```

deepxde/icbc/boundary_conditions.py
```diff
@@ -45,28 +55,57 @@ def __init__(self, geom, on_boundary, component):
...
     @abstractmethod
-    def error(self, X, inputs, outputs, beg, end, aux_var=None):
+    def error(
+        self,
+        X: NDArray[Any],
+        inputs: TensorOrTensors,
+        outputs: Tensor,
+        beg: int,
+        end: int,
+        aux_var: Union[NDArray[np.float_], None] = None,
```
Saransh-cpp, line 84: This should work once the suggestion above is applied.
```suggestion
        aux_var: NDArray[np.float_] | None = None,
```

deepxde/icbc/boundary_conditions.py
```diff
@@ -45,28 +55,57 @@ def __init__(self, geom, on_boundary, component):
...
+            if ``BC.error`` involves derivatives and the output has multiple components.
+    """
 
-    def __init__(self, geom, func, on_boundary, component=0):
+    def __init__(
+        self,
+        geom: Geometry,
+        func: Callable[[NDArray[np.float_]], NDArray[np.float_]],
+        on_boundary: Callable[[NDArray[Any], NDArray[Any]], NDArray[np.bool_]],
+        component: Union[List[int], int] = 0,
```
Saransh-cpp, line 107:
```suggestion
        component: list[int] | int = 0,
```

deepxde/types.py
```diff
@@ -0,0 +1,7 @@
+import numpy as np
+from typing import Callable, Dict, List, Optional, Sequence, Tuple, Union, TypeVar
+from numpy.typing import NDArray, ArrayLike
+
+# Tensor from any backend
+Tensor = TypeVar("Tensor")
+TensorOrTensors = Union[Tensor, Sequence[Tensor]]
```
Saransh-cpp, line 7:
```suggestion
TensorOrTensors = Tensor | Sequence[Tensor]

```

deepxde/types.py
```diff
@@ -0,0 +1,7 @@
+import numpy as np
+from typing import Callable, Dict, List, Optional, Sequence, Tuple, Union, TypeVar
```
Saransh-cpp, line 2:
```suggestion
from __future__ import annotations
from typing import Sequence, TypeVar
```

Saransh-cpp (changes requested):
Some suggestions 🙂 

Given that DeepXDE supports only Python 3.8+, we can use -
```py
from __future__ import annotations
```
to stay up to date!

I have made a few suggestions regarding the comment above - the rest of the types can be revamped accordingly too!

Also, [`mypy`](https://mypy.readthedocs.io/en/stable/) should be added in the [CI](https://github.com/lululxvi/deepxde/blob/master/.github/workflows/build.yml) to make sure that the types are correct and are being respected throughout the codebase. You can add mypy's configuration in [`pyproject.toml`](https://github.com/lululxvi/deepxde/blob/master/pyproject.toml). This can maybe go in another PR.

(PS: these are just suggestions, the final call would be of DeepXDE's maintainers.)

deepxde/types.py
supercgor: Thank you for your suggestions, I will change them later 👍

lululxvi:
@supercgor @Saransh-cpp Any idea about this PR?

deepxde/backend/backend.py
```diff
@@ -1,3 +1,9 @@
+from __future__ import annotations
+
+from numbers import Number
+from typing import Sequence, overload
+from ..types import Tensor, dtype, SparseTensor, TensorOrTensors
+from numpy.typing import NDArray, ArrayLike
```
Saransh-cpp, line 6: These imports (the heavy ones) should be under an [`if typing.TYPE_CHECKING`](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING).

deepxde/backend/backend.py
```diff
@@ -58,19 +64,19 @@ def data_type_dict():
...
+def is_gpu_available() -> bool:
     """Returns a bool indicating if GPU is currently available.
 
     Returns:
         True if a GPU device is available.
     """
 
 
-def is_tensor(obj):
+def is_tensor(obj: object) -> bool:
```
Saransh-cpp, line 75: Using `object` in type hints is not useful + not a very common practice. You would want to narrow this down ideally (because it is as helpful as not having a type hint at the moment).

deepxde/geometry/geometry.py
```diff
@@ -1,37 +1,42 @@
+from __future__ import annotations
+
 import abc
-from typing import Literal
+from typing import Callable, Literal
+from numbers import Number
 
 import numpy as np
+from numpy.typing import NDArray
```
Saransh-cpp, line 8: Should be moved in an `if` condition.

deepxde/geometry/geometry_2d.py
```diff
@@ -1,19 +1,25 @@
-__all__ = ["Disk", "Ellipse", "Polygon", "Rectangle", "StarShaped", "Triangle"]
+from __future__ import annotations
 
-from typing import Union, Literal
+__all__ = ["Disk", "Ellipse", "Polygon", "Rectangle", "StarShaped", "Triangle"]
+from numbers import Number
+from typing import Any, Literal
 
 import numpy as np
+from numpy.typing import ArrayLike, NDArray
```
Saransh-cpp, line 8: Ditto

deepxde/geometry/geometry_2d.py
```diff
@@ -1,19 +1,25 @@
...
 from scipy import spatial
 
 from .geometry import Geometry
 from .geometry_nd import Hypercube, Hypersphere
 from .sampler import sample
 from .. import backend as bkd
 from .. import config
+from ..types import Tensor
```
Saransh-cpp, line 16: Ditto

deepxde/geometry/geometry_2d.py
```diff
@@ -739,7 +745,7 @@ def boundary_constraint_factor(
             self.x2_tensor = bkd.as_tensor(self.x2)
             self.x3_tensor = bkd.as_tensor(self.x3)
 
-        diff_x1_x2 = diff_x1_x3 = diff_x2_x3 = None
+        # diff_x1_x2 = diff_x1_x3 = diff_x2_x3 = 0.0
```
Saransh-cpp, line 748: There are a couple of code (logic) changes in this PR, which should not be the case.

deepxde/geometry/timedomain.py
```diff
@@ -11,17 +14,17 @@
...
         self.t1 = t1
 
-    def on_initial(self, t):
+    def on_initial(self, t: NDArray[np.float_]) -> NDArray[np.bool_]:
         return isclose(t, self.t0).flatten()
 
 
-class GeometryXTime:
-    def __init__(self, geometry, timedomain):
+class GeometryXTime():
```
Saransh-cpp, line 26:
```suggestion
class GeometryXTime:
```

deepxde/types.py
Saransh-cpp, line 1: We should not call this file as `types` as [types is a builtin Python module](https://docs.python.org/3/library/types.html#module-types).

pyproject.toml
```diff
@@ -54,3 +54,8 @@ exclude = ["docker", "docs*", "examples*"]
 
 [tool.setuptools_scm]
 write_to = "deepxde/_version.py"
+
+[mypy]
+python_version = 3.8
+strict = false
+ignore_missing_imports = true
```
Saransh-cpp, line 61:
```suggestion
ignore_missing_imports = true

```

Saransh-cpp (changes requested):
@lululxvi this looks almost ready, but needs to be polished for merging. It would be great if @supercgor can pick this up again :)

echen5503:
@hheei any updates here?

PR #1463, opened 2023-08-30
Add possibility to enable line search in pytorch.

samuelburbulla:
This PR adds the ability to enable/disable line search in pytorch.
It activates line search by default, as this makes L-BFGS converge much more stable.

Unfortunately, there is no way to set the `maxls` parameter in pytorch (it always defaults to 25).
To give the possibility to disable the line search, I suggest to disable if `maxls` is non-positive.


lululxvi:
Good point. Need to also update the docstring 
https://github.com/lululxvi/deepxde/blob/683682c9baf876d58048160733d9897ad3090af3/deepxde/optimizers/config.py#L45

PR #1472, opened 2023-09-11
epochs -> iterations in examples

Saransh-cpp:
See https://github.com/lululxvi/deepxde/pull/1324#issuecomment-1713010997

### DIFF
--- a/examples/pinn_inverse/elliptic_inverse_field_batch.py
+++ b/examples/pinn_inverse/elliptic_inverse_field_batch.py
@@ -1,4 +1,4 @@
-"""Backend supported: tensorflow.compat.v1, pytorch, paddle"""
+"""Backend supported: pytorch, paddle"""
 import deepxde as dde
 import matplotlib.pyplot as plt
 import numpy as np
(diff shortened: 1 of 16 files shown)


examples/pinn_inverse/elliptic_inverse_field_batch.py
```diff
@@ -1,4 +1,4 @@
-"""Backend supported: tensorflow.compat.v1, pytorch, paddle"""
```
lululxvi, line 1: Why removing `tensorflow.compat.v1`?

Saransh-cpp: The example uses the `batch_size` argument in `PointSetBC`-

https://github.com/lululxvi/deepxde/blob/c807b4bff2e218fb094574c2b21d8d5f15c5292b/examples/pinn_inverse/elliptic_inverse_field_batch.py#L29

which is not implemented for `tensorflow.compat.v1` - 

https://github.com/lululxvi/deepxde/blob/c807b4bff2e218fb094574c2b21d8d5f15c5292b/deepxde/icbc/boundary_conditions.py#L195-L199

PR #1490, opened 2023-09-24
Enhanced JAX compatibility

tttc3:
This provides JAX compatibility with most ics and bcs. In addition, it provides checkpointing via Orbax.

To avoid requiring Orbax as a package wide dependency it is conditionally imported. While this provides the desired result, it is not PEP8 compliant; any better suggestions would be greatly appreciated.

### DIFF
--- a/deepxde/data/pde.py
+++ b/deepxde/data/pde.py
@@ -3,7 +3,6 @@ import numpy as np
 from .data import Data
 from .. import backend as bkd
 from .. import config
-from ..backend import backend_name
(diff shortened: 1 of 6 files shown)


lululxvi:
@ZongrenZou Any suggestions?

lululxvi:
You can split this PR into two PRs, one about ics/bcs, another one about Orbax

tttc3:
This pull has been split into #1491 and #1492.

PR #1491, opened 2023-09-25
JAX checkpointing via Orbax

tttc3:
Split from #1490 . Provides a very basic checkpointing functionality for flax based JAX models via Orbax. Essentially follows the simple example from [flax documentation](https://flax.readthedocs.io/en/latest/guides/use_checkpointing.html), without the use of the context manager.

As noted in #1490 , Orbax is currently imported lazily (only imported once first called) in a non PEP8 compliant manner, to avoid requiring Orbax as a package dependency. I feel that there should be a better way to do this, but I can't currently think of one. 
As such, I've marked this as a draft for now. 

Any thought or suggestions are greatly appreciated!

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,5 +1,6 @@
 matplotlib
 numpy
+orbax-checkpoint
 scikit-learn
 scikit-optimize>=0.9.0
 scipy
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,6 +37,7 @@ classifiers = [
 dependencies = [
     "matplotlib",
     "numpy",
+    "orbax-checkpoint",
     "scikit-learn",
     "scikit-optimize>=0.9.0",
     "scipy",
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 matplotlib
 numpy
+orbax-checkpoint
 scikit-learn
 scikit-optimize>=0.9.0
 scipy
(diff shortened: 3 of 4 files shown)


lululxvi:
From the information here https://orbax.readthedocs.io/en/latest/, Orbax is not a package, but rather a separate package for each functionality provided by the Orbax namespace. For checkpoint, we need orbax-checkpoint

tttc3:
I've updated the import to only require `orbax.checkpointing`

lululxvi:
As checkpoint is a very basic tool, I think it is ok to add it to the required package dependency.

tttc3:
Added orbax-checkpoint as a top level import and a package wide dependency. Still have `flax.training` as a non top level import though.

docs/requirements.txt
```diff
@@ -1,5 +1,6 @@
 matplotlib
 numpy
+orbax-checkpoint
```
lululxvi, line 3: Need to move this to the JAX part below.

pyproject.toml
```diff
@@ -37,6 +37,7 @@ classifiers = [
 dependencies = [
     "matplotlib",
     "numpy",
+    "orbax-checkpoint",
```
lululxvi, line 40: Not add here.

requirements.txt
```diff
@@ -1,5 +1,6 @@
 matplotlib
 numpy
+orbax-checkpoint
```
lululxvi, line 3: Not add here.

echen5503:
@tttc3 still working on this one?

tttc3:
I'm no longer working on this.

PR #1492, opened 2023-09-25
Improve JAX compatibility with ics and bcs.

tttc3:
Split out from #1490 . This pull provides JAX compatibility with most ics and bcs.

This may not be the nicest code, but it is a simple approach to providing the compatibility. Hence, I have marked this as a draft for now.

### DIFF
--- a/deepxde/data/fpde.py
+++ b/deepxde/data/fpde.py
@@ -7,7 +7,6 @@ import numpy as np
 from .pde import PDE
 from .. import backend as bkd
 from .. import config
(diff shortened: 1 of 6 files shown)


lululxvi:
Many new codes are added. I would suggest not to revise fPDE, IDE and PDE operator cases in this PR. First focus on pde case. Once this is done, then add others.

echen5503:
@tttc3  any new updates for this?

tttc3:
I'm no longer working on this.

PR #1511, opened 2023-10-05
loss_weights as a Model attribute

pescap:
Passing loss_weights as a Model attribute will allow for easier access to the weights and their update during training.

### DIFF
--- a/deepxde/model.py
+++ b/deepxde/model.py
@@ -31,7 +31,7 @@ class Model:
 
         self.opt_name = None
         self.batch_size = None
-        self.loss_weights = None  
+        self.loss_weights = None
         self.callbacks = None
         self.metrics = None
         self.external_trainable_variables = []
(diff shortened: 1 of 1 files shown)


deepxde/model.py
```diff
@@ -118,7 +118,8 @@ def compile(
             print("Compiling model...")
         self.opt_name = optimizer
         loss_fn = losses_module.get(loss)
-        self.losshistory.set_loss_weights(loss_weights)
+        self.losshistory.set_loss_weights(loss_weights)  # Check what it is for.
```
pescap, line 121: I'd be happy to understand the role of `set_loss_weights`. Is it useful? Shall we remove it?

lululxvi: This is used to record the current loss weights in `self.losshistory`.

deepxde/model.py
```diff
@@ -118,7 +118,7 @@ def compile(
             print("Compiling model...")
         self.opt_name = optimizer
         loss_fn = losses_module.get(loss)
-        self.losshistory.set_loss_weights(loss_weights)
+        self.loss_weights = loss_weights
```
lululxvi, line 121: Add `self.loss_weights = None` in line 33

deepxde/model.py
```diff
@@ -364,7 +365,7 @@ def closure():
         self.outputs_losses_test = outputs_losses_test
         self.train_step = train_step
 
-    def _compile_jax(self, lr, loss_fn, decay, loss_weights):
+    def _compile_jax(self, lr, loss_fn, decay):
```
lululxvi, line 368: If loss_wieght is not None, add a error here that loss_weights is not supported.

PR #1516, opened 2023-10-09
Added sphinx-rtd-theme to allow ReadTheDocs compilation

pescap:
Just to check whether readthedocs compiles with this line

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -18,7 +18,6 @@ jax
 flax
 optax
 
-docutils<0.18  # https://github.com/readthedocs/readthedocs.org/issues/8616
 sphinx-copybutton
 protobuf==3.20.*
 sphinx-rtd-theme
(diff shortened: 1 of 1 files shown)


docs/requirements.txt
```diff
@@ -21,3 +21,4 @@ optax
 docutils<0.18  # https://github.com/readthedocs/readthedocs.org/issues/8616
```
lululxvi, line 21: Do you want to double-check if this is needed anymore?

docs/requirements.txt
```diff
@@ -18,6 +18,6 @@ jax
 flax
 optax
 
-docutils<0.18  # https://github.com/readthedocs/readthedocs.org/issues/8616
 sphinx-copybutton
 protobuf==3.20.*
+sphinx-rtd-theme
```
lululxvi, line 23: Reorder
```
sphinx-copybutton
sphinx-rtd-theme
protobuf==3.20.*
```
Also, how about `protobuf==3.20.*`?

PR #1525, opened 2023-10-17
Backend paddle: Support elasticity_plate example

lijialin03:
Support paddle backend for elasticity_plate example
result:
with pytorch backend:
image
with paddle backend:
image


examples/pinn_forward/elasticity_plate.py
```diff
@@ -6,12 +6,22 @@
 """
```
lululxvi, line 6: Add paddle in the first line.

PR #1529, opened 2023-10-19
Update research.rst

Kostap7:
This is our newest publication that utilizes the DeepXDE library: https://www.nature.com/articles/s41598-023-43325-1

Best, 
Konstantinos

### DIFF
--- a/docs/user/research.rst
+++ b/docs/user/research.rst
@@ -120,6 +120,7 @@ Here is a list of research papers that used DeepXDE. If you would like your pape
 PINN
 ----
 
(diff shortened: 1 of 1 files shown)


docs/user/research.rst
```diff
@@ -120,6 +120,7 @@ Here is a list of research papers that used DeepXDE. If you would like your pape
 PINN
 ----
 
+#. K\. Prantikos, L.H. Tsoukalas, & A. Heifetz. `Physics-informed neural network with transfer learning (TL-PINN) based on domain similarity measure for prediction of nuclear reactor transients <https://www.nature.com/articles/s41598-023-43325-1>`_. *Scientific Reports*,  13, 16840 (2023).
```
lululxvi, line 123: `,  13` --- there are two empty spaces in between. Use only one.

lululxvi, line 123: the format of year is not correct. Check other examples.

PR #1532, opened 2023-10-23
Tensorflow 1.x backend: multiple outputs extension of DeepONet

vl-dud:
This pull request suggests an enhancement to DeepONet, addressing its current limitation of producing only a scalar output, which restricts its applicability in various tasks. The functionality of handling multiple outputs has been added to both the DeepONet and DeepONetCartesianProd classes.
This is an implementation for TensorFlow 1.x backend.

PR #1549, opened 2023-11-07
Add an example of Poisson-like 2D problem for multioutput DeepONet Add the example

vl-dud:
The separate PR for [https://github.com/lululxvi/deepxde/pull/1532](https://github.com/lululxvi/deepxde/pull/1532)


lululxvi:
As the code of multi-output is merged, please double check this PR.

vl-dud:
The example works correctly for all strategies, both with TF 1.x and 2.x.

examples/operator/poisson_aligned_multioutput_pideeponet_2d.py
```diff
@@ -0,0 +1,65 @@
+"""
+Poisson-like 2D problem
+Supported backend: tensorflow.compat.v1, tensorflow
+"""
```
lululxvi, line 4:
```
"""Supported backend: tensorflow.compat.v1, tensorflow

Poisson-like 2D problem ... More explanation...
"""
```

vl-dud: I'm afraid this is also a fake problem used for demonstration. The equations are of the form `dA_xx = f` and `dB_tt = f`, where `dA_xx` and `dB_tt` are the second-order derivatives of some A and B with respect to spatial and temporal dimensions, respectively.

lululxvi: Umm. Is it possible to have some real example?

vl-dud: Sorry, but the suitable real problem does not come to mind. Do you have any idea?

lululxvi: A 2D fluid problem with outputs of u and v?

echen5503:
@vl-dud still working on this?

vl-dud:
@echen5503 No, we can close this PR.

PR #1533, opened 2023-10-23
Tensorflow 2.x backend: multiple outputs extension of DeepONet

vl-dud:
This pull request suggests an enhancement to DeepONet, addressing its current limitation of producing only a scalar output, which restricts its applicability in various tasks. The functionality of handling multiple outputs has been added to both the DeepONet and DeepONetCartesianProd classes.
This is an implementation for TensorFlow 2.x backend.

PR #1550, opened 2023-11-07
Add the antiderivative example for multioutput DeepONet 

vl-dud:
The separate PR for [https://github.com/lululxvi/deepxde/pull/1533](https://github.com/lululxvi/deepxde/pull/1533)


lululxvi:
Double check this PR.

vl-dud:
I checked it. The example works correctly for all strategies, both with TF 1.x and 2.x.

lululxvi:
Then this example may be redundant and would make the reader confused. We already have this https://github.com/lululxvi/deepxde/pull/1549/files , so it is not necessary to have this example.

vl-dud:
Ok, I'm closing this request.

Issue #54, opened 2020-05-30
How can we save model so for multiple predictions


jhssyb:
Hi Lu, just wondering how can we save network structure and weights after lengthy training so that next time we can only use the pretrained model for predictions rather than starting from scratch (first training then predictions)?

lululxvi:
- To save model, there are two ways:
1. Save the model only in the last step. Use `model_save_path` in [`Mode.train`](https://deepxde.readthedocs.io/en/latest/modules/deepxde.html#deepxde.model.Model.train).
2. Save the model every certain steps. Use [ModelCheckpoint](https://deepxde.readthedocs.io/en/latest/modules/deepxde.html#deepxde.callbacks.ModelCheckpoint). See Poisson_Dirichlet_1d.py.

```python
checker = dde.callbacks.ModelCheckpoint(
    "model/model.ckpt", save_better_only=True, period=1000
)
model.train(epochs=epochs, callbacks=[checker])
```

- To restore a model, use [`Model.restore`](https://deepxde.readthedocs.io/en/latest/modules/deepxde.html#deepxde.model.Model.restore).

```python
model.restore("model/model.ckpt-?", verbose=1)  # Replace ? with the exact filename
```

- To continue training from a saved model, use `model_restore_path` in [`Mode.train`](https://deepxde.readthedocs.io/en/latest/modules/deepxde.html#deepxde.model.Model.train).

jhssyb:
Thanks for your detailed answer. I can save the model now

tanielfranklin:
Hi, what about inverse problems with trainable variables? When I restore a checkpoint a "Key Variable/Adam not found in checkpoint" error is thrown. Any tip?

lululxvi:
You can open a new issue with more details.

SongPapers:
Hi Dr.Lu:
I want to save model by the second way:  using ModelCheckpoint. But I meet a problem. For briefly, I put my code.

```
data = dde.data.TimePDE(
    geomtime,
    pde,
    [IC_A, IC_u, ],
    num_domain = 100,
    num_boundary = 10,
    num_initial  = 10,
)

net = dde.maps.FNN(
    layer_sizes = [dim_input] + hidden_layers_num + [dim_output],
    activation         = activation_select,
    kernel_initializer = "Glorot normal",
)

model2 = dde.Model(data, net)
model2.compile('adam', 
              lr = 0.0005,
)

checkpointer = dde.callbacks.ModelCheckpoint(
[...]

PR #1559, opened 2023-11-14
Tensorflow 1.x backend: define a model multiple times in the same code

vl-dud:
When you define `deepxde.Model` multiple times in the same code it leads to errors. This was briefly mentioned in [https://github.com/lululxvi/deepxde/issues/54](https://github.com/lululxvi/deepxde/issues/54). So this code:
```
import os

os.environ["DDEBACKEND"] = "tensorflow.compat.v1"
import deepxde as dde
import numpy as np


def func(x):
    return x * np.sin(5 * x)


num_train = 16
num_test = 100
data = dde.data.Function(geom, func, num_train, num_test)

net = dde.nn.FNN([1] + [20] * 3 + [1], "tanh", "Glorot uniform")
model = dde.Model(data, net)
model.train(iterations=1000)
model.save("dump/model")

net_2 = dde.nn.FNN([1] + [20] * 3 + [1], "tanh", "Glorot uniform")
model_2 = dde.Model(data, net_2)
model_2.compile("adam", lr=0.001)
model_2.restore("dump/model-1000.ckpt", verbose=1)
```
will not be executed. I suggest resetting the global default graph to fix this.

### DIFF
--- a/deepxde/nn/tensorflow_compat_v1/nn.py
+++ b/deepxde/nn/tensorflow_compat_v1/nn.py
@@ -9,6 +9,14 @@ class NN:
     """Base class for all neural network modules."""
 
     def __init__(self):
+        try:
+            seed = tf.random.get_seed(None)[0]
+        except RuntimeError:
+            seed = None
+        tf.reset_default_graph()
+        if seed is not None:
+            tf.set_random_seed(seed)
+
         self.training = tf.placeholder(tf.bool)
         self.regularizer = None
 


deepxde/nn/tensorflow_compat_v1/nn.py
```diff
@@ -9,6 +9,14 @@ class NN:
     """Base class for all neural network modules."""
 
     def __init__(self):
+        try:
+            seed = tf.random.get_seed(None)[0]
```
lululxvi, line 13: Why setting the seed?

vl-dud: To handle the situation when a user has set a seed before (e.g. using `deepxde.config.set_random_seed`)

lululxvi:
Actually DeepXDE is not indented to work for defining Model for several times. I am not sure if doing this modification would cause some issues at somewhere else.

There is an utility function for this purpose: https://deepxde.readthedocs.io/en/latest/modules/deepxde.utils.html#deepxde.utils.external.apply

vl-dud:
If you think it might cause any issues, then let's close this PR. But I spent a lot of time trying to figure out what the problem was with my code and this feature is quite unobvious. It might make sense to pay more attention to this in the documentation and examples to save time for other users.

lululxvi:
You are right. Let us think more about this. A note in the documentation could be helpful.

PR #1579, opened 2023-12-03
Tensorflow 1.x backend: add dropout to DeepONet

vl-dud:
Now DeepONet supports dropout technique.

### DIFF
--- a/deepxde/nn/tensorflow_compat_v1/deeponet.py
+++ b/deepxde/nn/tensorflow_compat_v1/deeponet.py
@@ -305,7 +305,7 @@ class DeepONet(NN):
                 )
                 if self.dropout_rate > 0:
                     y_func = tf.layers.dropout(
-                        y_func, rate=self.dropout_rate, training=self.trainable_branch
+                        y_func, rate=self.dropout_rate, training=self.training
                     )
             y_func = self._stacked_dense(
                 y_func,
@@ -326,7 +326,7 @@ class DeepONet(NN):
                 )
                 if self.dropout_rate > 0:
                     y_func = tf.layers.dropout(
-                        y_func, rate=self.dropout_rate, training=self.trainable_branch
+                        y_func, rate=self.dropout_rate, training=self.training
                     )
             y_func = self._dense(
                 y_func,
@@ -356,7 +356,7 @@ class DeepONet(NN):
             )
             if self.dropout_rate > 0:
                 y_loc = tf.layers.dropout(
-                    y_loc, rate=self.dropout_rate, training=trainable
+                    y_loc, rate=self.dropout_rate, training=self.training
                 )
         return y_loc
 
--- a/deepxde/nn/tensorflow_compat_v1/deeponet.py
+++ b/deepxde/nn/tensorflow_compat_v1/deeponet.py
@@ -155,6 +155,7 @@ class DeepONet(NN):
             both trunk and branch nets. If `activation` is a ``dict``, then the trunk
             net uses the activation `activation["trunk"]`, and the branch net uses
             `activation["branch"]`.
+        dropout_rate (float): The dropout rate, between 0 and 1.
         trainable_branch: Boolean.
         trainable_trunk: Boolean or a list of booleans.
         num_outputs (integer): Number of outputs. In case of multiple outputs, i.e., `num_outputs` > 1,
@@ -192,6 +193,7 @@ class DeepONet(NN):
         activation,
         kernel_initializer,
         regularization=None,
+        dropout_rate=0,
         use_bias=True,
         stacked=False,
         trainable_branch=True,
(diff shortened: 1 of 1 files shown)


vl-dud:
> I am curious how useful it is. In general, I found dropout is not that useful, and L1/L2 regularization seems good enough.

Hyperparameter tuning showed that in my case neither dropout nor regularization is required. However I will use dropout anyway for UQ.

lululxvi:
Yes, dropout is useful for UQ. How do you implement DeepONet UQ? Using the callback?

vl-dud:
I just run `model.predict` many times to get the final prediction with CI:

```
def predict_with_uncertainty(model, x, trial_num=100):
    predictions = []
    for _ in range(trial_num):
        predictions.append(model.predict(x))
    return np.mean(predictions, axis=0), np.std(predictions, axis=0)
```

lululxvi:
In fact, we have this callback https://deepxde.readthedocs.io/en/latest/modules/deepxde.html#deepxde.callbacks.DropoutUncertainty . Does this work for your case?

vl-dud:
Yes, I used `DropoutUncertainty`. The above code snippet is more convenient in my case, since I use it with already trained models.
In addition, it allows you to set `trial_num`, which is a constant in `DropoutUncertainty`.

deepxde/nn/tensorflow_compat_v1/deeponet.py
```diff
@@ -300,6 +303,10 @@ def build_branch_net(self):
                     activation=self.activation_branch,
                     trainable=self.trainable_branch,
                 )
+                if self.dropout_rate > 0:
+                    y_func = tf.layers.dropout(
+                        y_func, rate=self.dropout_rate, training=self.trainable_branch
```
lululxvi, line 308: It should not be `self.trainable_branch`, which is used to indicate whether to train the branch during the training. It should be `self.training`, which indicates whether it is training or testing. See https://github.com/lululxvi/deepxde/blob/master/deepxde/nn/tensorflow_compat_v1/nn.py 
and
https://github.com/lululxvi/deepxde/blob/ad6399b27c3a0702abd24d45f889ca55497476c2/deepxde/callbacks.py#L307

vl-dud: I fixed it. I suppose this applies to `self.trainable_trunk` as well.

deepxde/nn/tensorflow_compat_v1/deeponet.py
```diff
@@ -553,6 +574,8 @@ def build_branch_net(self):
                     kernel_initializer=self.kernel_initializer,
                     kernel_regularizer=self.regularizer,
                 )
+                if self.dropout_rate > 0:
+                    y_func = tf.layers.dropout(y_func, rate=self.dropout_rate)
```
lululxvi, line 578: Miss `training=self.training`

deepxde/nn/tensorflow_compat_v1/deeponet.py
```diff
@@ -331,15 +342,22 @@ def build_trunk_net(self):
...
+            )
             y_loc = self._dense(
                 y_loc,
                 self.layer_size_loc[i],
                 activation=self.activation_trunk,
                 regularizer=self.regularizer,
-                trainable=self.trainable_trunk[i - 1]
-                if isinstance(self.trainable_trunk, (list, tuple))
-                else self.trainable_trunk,
+                trainable=trainable,
```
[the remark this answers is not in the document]
vl-dud: I assumed that the same logic applies here as in the branch network. I am wrong?

deepxde/nn/tensorflow_compat_v1/deeponet.py
vl-dud: For class `DeepONetCartesianProd`, `self.training` is not used anywhere. Should I add it only on line 578?

deepxde/nn/tensorflow_compat_v1/deeponet.py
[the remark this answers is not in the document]
lululxvi: I mean the new code (line 355, line 345--349) works identically as the original code.

deepxde/nn/tensorflow_compat_v1/deeponet.py
lululxvi: I guess so. In `DeepONet`, `self.training` is also only used a few times, right? But it is better to test the code.

vl-dud: Fixed

deepxde/nn/tensorflow_compat_v1/deeponet.py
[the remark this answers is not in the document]
vl-dud: In the previous [commit](https://github.com/lululxvi/deepxde/pull/1579/commits/71f1a460662b904eedf116d4c2d5db3cea57e8fb), the `trainable` variable is used twice. Now this is unnecessary. I fixed it.

PR #1593, opened 2023-12-18
Backend JAX supports 1D diffusion equation in PINN forward

ZongrenZou:
Now both the forward and reverse modes support 1D diffusion equation in PINN forward.

### DIFF
--- a/examples/pinn_forward/diffusion_1d.py
+++ b/examples/pinn_forward/diffusion_1d.py
@@ -32,19 +32,19 @@ def pde(x, y):
     #     + torch.exp(-x[:, 1:])
     #     * (torch.sin(np.pi * x[:, 0:1]) - np.pi ** 2 * torch.sin(np.pi * x[:, 0:1]))
     # )
-    # Backend paddle
-    # return (
-    #     dy_t
-    #     - dy_xx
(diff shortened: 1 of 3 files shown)


examples/pinn_forward/diffusion_1d.py
```diff
@@ -7,11 +7,17 @@
 # import torch
 # Backend paddle
 # import paddle
+# Backend jax
```
lululxvi, line 10: Move jax before paddle.

examples/pinn_forward/diffusion_1d.py
```diff
@@ -33,6 +39,13 @@ def pde(x, y):
     #     + paddle.exp(-x[:, 1:])
     #     * (paddle.sin(np.pi * x[:, 0:1]) - np.pi ** 2 * paddle.sin(np.pi * x[:, 0:1]))
     # )
+    # Backend jax
```
lululxvi, line 42: Move jax before paddle.

ZongrenZou: Done.

examples/pinn_forward/diffusion_1d.py
ZongrenZou: Done.

PR #1594, opened 2023-12-18
Backend JAX supports model.predict

ZongrenZou:


deepxde/model.py
```diff
@@ -935,6 +935,24 @@ def op(inputs):
             # Clear cached Jacobians and Hessians.
             grad.clear()
             y = utils.to_numpy(y)
+        elif backend_name == "jax":
+            if utils.get_num_args(operator) == 2:
+
+                @jax.jit
+                def op(inputs):
+                    y_fn = lambda _x: self.net.apply(self.params[0], _x)
```
lululxvi, line 943: `self.params[0]` --> `self.net.params`

deepxde/model.py
```diff
@@ -935,6 +935,24 @@ def op(inputs):
...
+                    return operator(inputs, (y_fn(inputs), y_fn))
```
lululxvi, line 944: Do you need vmap for y_fn(inputs)?

deepxde/model.py
```diff
@@ -935,6 +935,24 @@ def op(inputs):
...
+
+            elif utils.get_num_args(operator) == 3:
+                # TODO: JAX backend Implementation of Auxiliary variables.
+                raise NotImplementedError(
+                    "Model.predict() with auxiliary variable hasn't been implemented "
+                    "for backend jax."
+                )
+            y = op(x)
+            # Clear cached Jacobians and Hessians.
+            grad.clear()
```
lululxvi, line 954: JAX doesn't need grad.clear

ZongrenZou: Done.

deepxde/model.py
ZongrenZou: No. flax supports vector-valued input. As far as I know, current deepxde only has vmap in "gradients".

deepxde/model.py
ZongrenZou: Yes this is better. Done.

PR #1595, opened 2023-12-19
Backend JAX supports more examples on forward problems with PINNs

ZongrenZou:


ZongrenZou:
Tested these three examples on both the forward and reverse modes for computing gradients.

PR #1602, opened 2023-12-24
Add regularization to DeepONet

vl-dud:
`regularization` parameter was missing in `DeepONet` class (TF 2.x)

### DIFF
--- a/deepxde/nn/tensorflow/deeponet.py
+++ b/deepxde/nn/tensorflow/deeponet.py
@@ -235,6 +235,7 @@ class DeepONet(NN):
         kernel_initializer,
         num_outputs=1,
         multi_output_strategy=None,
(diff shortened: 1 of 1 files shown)


deepxde/nn/tensorflow/deeponet.py
```diff
@@ -235,6 +235,7 @@ def __init__(
         kernel_initializer,
         num_outputs=1,
         multi_output_strategy=None,
+        regularization=None,
```
lululxvi, line 238: Update the docstring above.

vl-dud: If I'm not mistaken, there is no docstring for `regularization` in any deepxde class. Should I add the docstring only here or everywhere?

PR #1607, opened 2024-01-01
Backend PyTorch supports forward-mode automatic differentiation: Step 1

ZongrenZou:


deepxde/model.py
```diff
@@ -301,7 +301,12 @@ def outputs_losses(training, inputs, targets, auxiliary_vars, losses_fn):
             # Data losses
             if targets is not None:
                 targets = torch.as_tensor(targets)
-            losses = losses_fn(targets, outputs_, loss_fn, inputs, self)
+            if config.autodiff == "reverse":
+                aux = None
+            elif config.autodiff == "forward":
+                # if forward-mode AD is used, then a forward call needs to be passed
+                aux = [lambda _x: self.net(_x)]
```
lululxvi, line 308:
```
aux = None
if config.autodiff == "forward":
    # if forward-mode AD is used, then a forward call needs to be passed
    aux = [lambda _x: self.net(_x)]
```

ZongrenZou: Done.

deepxde/model.py
```diff
@@ -301,7 +301,9 @@ def outputs_losses(training, inputs, targets, auxiliary_vars, losses_fn):
             # Data losses
             if targets is not None:
                 targets = torch.as_tensor(targets)
-            losses = losses_fn(targets, outputs_, loss_fn, inputs, self)
+            # if forward-mode AD is used, then a forward call needs to be passed
+            aux = [lambda _x: self.net(_x)] if config.autodiff == "forward" else None
```
lululxvi, line 305:
```
aux = [self.net] if config.autodiff == "forward" else None
```

ZongrenZou: Done.

PR #1608, opened 2024-01-02
Backend PyTorch supports forward-mode automatic differentiation: Step 2

ZongrenZou:


deepxde/gradients/gradients_forward.py
```diff
@@ -3,7 +3,7 @@
 __all__ = ["hessian", "jacobian"]
 
 from .jacobian import Jacobian, Jacobians
-from ..backend import backend_name, jax
+from ..backend import backend_name, torch, jax
```
lululxvi, line 6: `from ..backend import backend_name, jax, torch`

deepxde/gradients/jacobian.py
```diff
@@ -1,6 +1,7 @@
 from abc import ABC, abstractmethod
 
 from ..backend import backend_name
+from .. import config
```
lululxvi, line 4:
```
from .. import config
from ..backend import backend_name
```

deepxde/gradients/gradients_forward.py
ZongrenZou: Done.

deepxde/gradients/jacobian.py
ZongrenZou: Done.

PR #1614, opened 2024-01-03
Backend tensorflow.compat.v1 supports forward-mode automatic differentiation via double backwards trick

ZongrenZou:


deepxde/gradients/gradients_forward.py
```diff
@@ -20,19 +20,27 @@ def __call__(self, i=None, j=None):
 
         # Compute J[:, j]
         if j not in self.J:
-            if backend_name in [
-                "tensorflow.compat.v1",
-                "paddle",
-            ]:
+            if backend_name == "paddle":
```
lululxvi, line 23: Reorder as tf1, tf, pytorch, jax, paddle

deepxde/gradients/gradients_forward.py
```diff
@@ -87,6 +95,8 @@ def grad_fn(x):
                     self.J[j][0][:, i : i + 1],
                     lambda x: self.J[j][1](x)[i : i + 1],
                 )
+            elif backend_name == "tensorflow.compat.v1":
```
lululxvi, line 98: Put tf1 as the first `if`

ZongrenZou: Done

deepxde/gradients/gradients_forward.py
ZongrenZou: Done

PR #1631, opened 2024-01-20
Tentative: multi-dimensional outputs for pytorch and paddle

kuangdai:
# How this was done

I made it in two steps:
1) moving the strategy classes from `tensorflow`'s `deeponet.py` to a separate file `deeponet_strategy.py` so that all backends can share it
2) adding strategeis to `pytorch` and `paddle` following the same way (with few synatx adaptions)

# Remaining Issue
I noted that in `call(self, x_func, x_loc, training=False)` of `DeepONetStrategy`, the argument `training` is completely unused. It is kept in this PR because I don't know if it serves some other purpose. I suggest removal of it.

# Tests
A file `test_strat.py` (renamed to `test_start.txt` for uploading) is attached in this description. It tests all the model classes (`DeepONet`, `DeepONetCartesianProd`, `PODDeepONet`) and strategy settings (single, independent, split both, split branch, split trunk), making sure that the forward pass yields the correct output shape. To reproduce the tests, put this file under the top directory and run

```bash
DDE_BACKEND=tensorflow python test_strat.py 
DDE_BACKEND=pytorch python test_strat.py 
DDE_BACKEND=paddle python test_strat.py
```

And you should see no errors.

[test_strat.txt](https://github.com/lululxvi/deepxde/files/13999388/test_strat.txt)

### DIFF
--- a/deepxde/nn/deeponet_strategy.py
+++ b/deepxde/nn/deeponet_strategy.py
@@ -1,0 +1,172 @@
+from abc import ABC, abstractmethod
+
+
+class DeepONetStrategy(ABC):
+    """DeepONet building strategy.
+
+    See the section 3.1.6. in
+    L. Lu, X. Meng, S. Cai, Z. Mao, S. Goswami, Z. Zhang, & G. Karniadakis.
+    A comprehensive and fair comparison of two neural operators
+    (with practical extensions) based on FAIR data.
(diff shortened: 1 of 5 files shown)


lululxvi:
Great.

Could you make paddle a separate PR? As I need someone else to check the paddle code.

PR #1636, opened 2024-01-24
Callback that restores the best model

vl-dud:
It is often necessary to restore the best model at the end of training. `BestModelRestoring` allows you to do it without working directly with files and `model.restore` method.

### DIFF
--- a/deepxde/callbacks.py
+++ b/deepxde/callbacks.py
@@ -1,5 +1,7 @@
+import os
 import sys
 import time
+from tempfile import TemporaryDirectory
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


lululxvi:
Callbacks are designed to be called frequently during training. If it is only called once at the end, it is better just to implement a function in [utils](https://github.com/lululxvi/deepxde/tree/master/deepxde/utils)

vl-dud:
You are right, but saving of the best model occurs after each training epoch. Callbacks seem to be better suited for this. If I'm wrong, please tell me how to reproduce the behavior similar to `on_epoch_end` in utils?

lululxvi:
There is a callback https://deepxde.readthedocs.io/en/latest/modules/deepxde.html#deepxde.callbacks.ModelCheckpoint that saves the model. I suggest to use this one to save the model and write a util function to load.

vl-dud:
I don’t see any particular advantages of such a util function over `model.restore`. PR probably doesn't make sense.

PR #1643, opened 2024-02-03
Multi-dimentional output strategy for DeepONet in PyTorch

kuangdai:
# How this was done

I made it in two steps:
1) moving the strategy classes from `tensorflow`'s `deeponet.py` to a separate file `deeponet_strategy.py` so that all backends can share it
2) adding strategeis to `pytorch` and `paddle` following the same way (with few synatx adaptions). Changes for `paddle` comes only after this PR is merged.

# Remaining Issue
I noted that in `call(self, x_func, x_loc, training=False)` of `DeepONetStrategy`, the argument `training` is completely unused. It is kept in this PR because I don't know if it serves some other purpose. I suggest removal of it.

# Tests
A file `test_strat.py` (renamed to `test_start.txt` for uploading) is attached in this description. It tests all the model classes (`DeepONet`, `DeepONetCartesianProd`, `PODDeepONet`) and strategy settings (single, independent, split both, split branch, split trunk), making sure that the forward pass yields the correct output shape. To reproduce the tests, put this file under the top directory and run

```bash
DDE_BACKEND=tensorflow python test_strat.py 
DDE_BACKEND=pytorch python test_strat.py 
DDE_BACKEND=paddle python test_strat.py
```

And you should see no errors.

[test_strat.txt](https://github.com/lululxvi/deepxde/files/13999388/test_strat.txt)

### DIFF
--- a/deepxde/nn/pytorch/deeponet.py
+++ b/deepxde/nn/pytorch/deeponet.py
@@ -102,13 +102,12 @@ class DeepONet(NN):
         self.branch, self.trunk = self.multi_output_strategy.build(
             layer_sizes_branch, layer_sizes_trunk
         )
-
-        self.b = cycle(
-            [
-                torch.nn.parameter.Parameter(torch.tensor(0.0))
-                for _ in range(self.num_outputs)
-            ]
+        self.branch = torch.nn.ModuleList(self.branch)
+        self.trunk = torch.nn.ModuleList(self.trunk)
+        self.biases = torch.nn.ParameterList(
+            [torch.nn.Parameter(torch.tensor(0.0)) for _ in range(self.num_outputs)]
         )
+        self.b = cycle(range(self.num_outputs))
 
     def build_branch_net(self, layer_sizes_branch):
         # User-defined network
@@ -123,7 +122,7 @@ class DeepONet(NN):
     def merge_branch_trunk(self, x_func, x_loc):
         y = torch.einsum("bi,bi->b", x_func, x_loc)
         y = torch.unsqueeze(y, dim=1)
-        y += next(self.b)
+        y += self.biases[next(self.b)]
         return y
 
     @staticmethod
@@ -224,13 +223,12 @@ class DeepONetCartesianProd(NN):
         self.branch, self.trunk = self.multi_output_strategy.build(
             layer_sizes_branch, layer_sizes_trunk
         )
-
-        self.b = cycle(
-            [
-                torch.nn.parameter.Parameter(torch.tensor(0.0))
-                for _ in range(self.num_outputs)
-            ]
+        self.branch = torch.nn.ModuleList(self.branch)
+        self.trunk = torch.nn.ModuleList(self.trunk)
+        self.biases = torch.nn.ParameterList(
+            [torch.nn.Parameter(torch.tensor(0.0)) for _ in range(self.num_outputs)]
         )
+        self.b = cycle(range(self.num_outputs))
 
     def build_branch_net(self, layer_sizes_branch):
         # User-defined network
@@ -244,7 +242,7 @@ class DeepONetCartesianProd(NN):
 
     def merge_branch_trunk(self, x_func, x_loc):
         y = torch.einsum("bi,ni->bn", x_func, x_loc)
-        y += next(self.b)
+        y += self.biases[next(self.b)]
         return y
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


deepxde/nn/tensorflow/deeponet.py
```diff
@@ -1,184 +1,19 @@
...
-            xs.append(x)
-            shift += size
-        return self.net.concatenate_outputs(xs)
+from ..deeponet_strategy import (
+    SingleOutputStrategy,
+    IndependentStrategy,
+    SplitBothStrategy,
+    SplitBranchStrategy,
+    SplitTrunkStrategy,
+)
```
lululxvi, line 16:
```python
from .. import activations
from ..deeponet_strategy import (
    SingleOutputStrategy,
    IndependentStrategy,
    SplitBothStrategy,
    SplitBranchStrategy,
    SplitTrunkStrategy,
)
from ... import config
from ...backend import tf
```

lululxvi: @kuangdai Reorder the import.

lululxvi:
If you make pytorch code as a separate PR, then I can merge this Tensorflow one right away.

kuangdai:
> If you make pytorch code as a separate PR, then I can merge this Tensorflow one right away.

I would suggest not rewinding the commits, as it may cause other issues. The change on tensorflow is simply separating the code into two, which does not serve a standalone purpose for a PR. Please take your time to check the pytorch implementation. No rush.

deepxde/nn/pytorch/deeponet.py
```diff
@@ -1,24 +1,65 @@
...
 from .. import activations
+from ..deeponet_strategy import (
+    SingleOutputStrategy,
+    IndependentStrategy,
+    SplitBothStrategy,
+    SplitBranchStrategy,
+    SplitTrunkStrategy,
+)
+from ... import config
+from ...backend import torch
```
lululxvi, line 16:
```python
from itertools import cycle

import torch

from .fnn import FNN
...
```

No need to import torch from backend. The reason of importing tf from backend is that tf has v1 and v2, which are managed by backend.

deepxde/nn/pytorch/deeponet.py
```diff
@@ -27,40 +68,76 @@ def __init__(
...
+        y = torch.unsqueeze(y, dim=1)
+        y += self.biases[next(self.b)]
+        return y
+
+    @staticmethod
+    def concatenate_outputs(ys):
+        return torch.concat(ys, dim=1)
 
-    def forward(self, inputs):
+    def forward(self, inputs, training=False):
```
lululxvi, line 134: pytorch doesn't use `training=False`, which is only needed in tf.

kuangdai: Good spot. As I wrote in PR description, `training` is unused in `DeepONetStrategy.call()`. Shall we remove it from strategies?

lululxvi: Yes, we can remove.

deepxde/nn/pytorch/deeponet.py
```diff
@@ -1,24 +1,65 @@
-import torch
+__all__ = ["DeepONet", "DeepONetCartesianProd", "PODDeepONet"]
+
+from itertools import cycle
```
lululxvi, line 3:
```
from itertools import cycle

import torch

from .fnn import FNN
```

kuangdai:
I removed argument `training`. 

Also, I realised that using `cycle` to iterate the biases inside the class is thread-unsafe, that is, if the external loop over `num_outputs` from strategies is multi-threaded, the cycle won't keep synchronised with `i` because multi-threading is unordered. Although no one will do this, it is better to have right logic. I changed to explicitly passing the index. This also makes the code more readable.

lululxvi:
> Also, I realised that using `cycle` to iterate the biases inside the class is thread-unsafe, that is, if the external loop over `num_outputs` from strategies is multi-threaded, the cycle won't keep synchronised with `i` because multi-threading is unordered. Although no one will do this, it is better to have right logic. I changed to explicitly passing the index. This also makes the code more readable.

Good point.

deepxde/nn/pytorch/deeponet.py
```diff
@@ -160,42 +292,46 @@ def __init__(
         activation,
         kernel_initializer,
         layer_sizes_trunk=None,
-        regularization=None,
```
lululxvi, line 163: Keep this `regularization`, which may be needed in some cases.

lululxvi: I was not clear. Only keep the `regularization` in POD-DeepONet.

PR #1650, opened 2024-02-11
add mixed precision support to deepxde

g-w1:


deepxde/config.py
```diff
@@ -15,6 +15,7 @@
 comm = None
 world_size = 1
 rank = 0
+mixed = False
```
lululxvi, line 18: Move `mixed` to `Real` https://github.com/lululxvi/deepxde/blob/master/deepxde/real.py

deepxde/config.py
```diff
@@ -79,6 +79,19 @@ def set_default_float(value):
     if value == "float16":
         print("Set the default float type to float16")
         real.set_float16()
+    elif value == "mixed":
```
lululxvi, line 82: Move this after `elif value == "float64":`

lululxvi, line 82: Update the docstring

deepxde/config.py
```diff
@@ -74,7 +74,7 @@ def set_default_float(value):
     The default floating point type is 'float32'.
 
     Args:
-        value (String): 'float16', 'float32', or 'float64'.
+        value (String): 'float16', 'float32', 'float64', or 'mixed'.
```
lululxvi, line 77:
```
value (String): 'float16', 'float32', 'float64', or 'mixed' (mixed precision in https://arxiv.org/abs/2401.16645).
```

deepxde/config.py
```diff
@@ -85,6 +85,19 @@ def set_default_float(value):
     elif value == "float64":
         print("Set the default float type to float64")
         real.set_float64()
+    elif value == "mixed":
+        print("Set training policy to mixed")
```
lululxvi, line 89:
```
print("Set the float type to mixed precision of float16 and float32")
```

deepxde/config.py
```diff
@@ -85,6 +85,19 @@ def set_default_float(value):
...
+        real.set_mixed()
+        if backend_name == "tensorflow":
+            real.set_float16()
+            tf.keras.mixed_precision.set_global_policy("mixed_float16")
+        elif backend_name == "pytorch":
+            # we cast to float16 during the passes in the training loop, but store in float32
```
lululxvi, line 95:
```
# Use float16 during the forward and backward passes, but store in float32
```

deepxde/config.py
```diff
@@ -85,6 +85,19 @@ def set_default_float(value):
...
+            real.set_float32()
+        else:
+            raise ValueError(
+                f"{backend_name} backend does not currently support mixed precision in deepXDE"
```
lululxvi, line 99:
```
f"{backend_name} backend does not currently support mixed precision."
```

deepxde/model.py
```diff
@@ -353,10 +353,20 @@ def outputs_losses_test(inputs, targets, auxiliary_vars):
 
         def train_step(inputs, targets, auxiliary_vars):
             def closure():
-                losses = outputs_losses_train(inputs, targets, auxiliary_vars)[1]
-                total_loss = torch.sum(losses)
-                self.opt.zero_grad()
-                total_loss.backward()
+                if config.real.mixed:
```
lululxvi, line 356: If we need to modify the code, then how about https://github.com/lu-group/mixed-precision-sciml?

deepxde/real.py
```diff
@@ -17,6 +18,9 @@ def __init__(self, precision):
     def __call__(self, package):
         return self.reals[package]
 
+    def set_mixed(self):
```
lululxvi, line 21: Move this to the end.

deepxde/model.py
g-w1: This is only for pytorch and Eric just implemented a custom training loop in https://github.com/lu-group/mixed-precision-sciml/blob/main/lipschitz-estimation-bounds/confirmation-of-theoretical-bounds.ipynb

examples/pinn_forward/Burgers_mixed.py
```diff
@@ -0,0 +1,55 @@
+"""Backend supported: tensorflow, pytorch
```
lululxvi, line 1: No need to add this example.

deepxde/model.py
```diff
@@ -353,10 +353,20 @@ def outputs_losses_test(inputs, targets, auxiliary_vars):
...
+                        ]
+                        total_loss = torch.sum(losses)
+                        # we do the backprop in float16
+                        self.opt.zero_grad()
+                        total_loss.backward()
+                else:
+                    losses = outputs_losses_train(inputs, targets, auxiliary_vars)[1]
+                    total_loss = torch.sum(losses)
+                    self.opt.zero_grad()
+                    total_loss.backward()
```
lululxvi, line 369:
```python
def closure():
    losses = outputs_losses_train(inputs, targets, auxiliary_vars)[1]
    total_loss = torch.sum(losses)
    self.opt.zero_grad()
    total_loss.backward()

def closure_mixed():
    with torch.autocast(device_type="cuda", dtype=torch.float16):
        closure()

self.opt.step(closure if not config.real.mixed else closure_mixed)
```

lululxvi:
If I use the API here to set mixed precision, then all the demo code can run in mixed precision?

g-w1:
Not all. The L-BFGS optimizer doesn't work in mixed precision. But if you add the line `dde.config.set_default_float("mixed")` to the top of your file, it should work unless it uses a feature (like L-BFGS) that is not supported with mixed precision. I have tested this on the Burgers.py example and it works.

deepxde/config.py
```diff
@@ -74,7 +74,7 @@ def set_default_float(value):
     The default floating point type is 'float32'.
 
     Args:
-        value (String): 'float16', 'float32', or 'float64'.
+        value (String): 'float16', 'float32', 'float64', or 'mixed' (mixed precision in https://arxiv.org/abs/2401.16645).
```
lululxvi, line 77:
```
value (String): 'float16', 'float32', 'float64', or 'mixed' (mixed precision).
```

deepxde/config.py
```diff
@@ -74,7 +74,7 @@ def set_default_float(value):
     The default floating point type is 'float32'.
```
lululxvi, line 74:
```
The default floating point type is 'float32'. Mixed precision uses the method in the paper: `J. Hayford, J. Goldman-Wetzler, E. Wang, & L. Lu. Speeding up and reducing memory usage for scientific machine learning via mixed precision. Computer Methods in Applied Mechanics and Engineering, 428, 117093, 2024 <https://doi.org/10.1016/j.cma.2024.117093>`_.
```

deepxde/config.py
```diff
@@ -85,6 +85,20 @@ def set_default_float(value):
     elif value == "float64":
         print("Set the default float type to float64")
         real.set_float64()
+    elif value == "mixed":
+        print("Set the float type to mixed precision of float16 and float32")
+        real.set_mixed()
```
lululxvi, line 90: This code is confusing. Here you do `real.set_mixed()`, but later you do either `real.set_float16()` or `real.set_float32()`. It seems you only need a flag `mixed`. You can do this flag after line 42.

deepxde/model.py
```diff
@@ -357,9 +357,12 @@ def closure():
                 total_loss = torch.sum(losses)
                 self.opt.zero_grad()
                 total_loss.backward()
-                return total_loss
 
-            self.opt.step(closure)
+            def closure_mixed():
```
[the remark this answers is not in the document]
g-w1: it creates a new closure that uses the mixed precision algorithm and then does `opt.step` with this below.

deepxde/config.py
g-w1: The `real.set_mixed()` flag is independent of the datatype that the real uses. For example, we use the `mixed` flag in torch when calculating the loss when conditionally using a closure that does autocast:
```
deepxde/model.py
380:            self.opt.step(closure if not config.real.mixed else closure_mixed)
```

lululxvi: It is not necessary to define `mixed` in `real`. You can simply define it in this file after line 42.

deepxde/config.py
```diff
@@ -71,20 +73,40 @@ def default_float():
...
+    `J. Hayford, J. Goldman-Wetzler, E. Wang, & L. Lu. Speeding up and reducing memory usage for scientific machine learning via mixed precision.
+    Computer Methods in Applied Mechanics and Engineering, 428, 117093, 2024 <https://doi.org/10.1016/j.cma.2024.117093>`_.
 
     Args:
-        value (String): 'float16', 'float32', or 'float64'.
+        value (String): 'float16', 'float32', 'float64', or 'mixed' (mixed precision).
     """
+    global mixed
     if value == "float16":
+        mixed = False
```
lululxvi, line 85: The default `mixed` is False in line 44. There is no need to set it to False again.

g-w1: I put this here in case someone wanted to choose mixed and then go back to float16. But I guess changing float type multiple times is quite rare, so I can remove it.

deepxde/model.py
```diff
@@ -374,7 +374,11 @@ def closure():
                 total_loss.backward()
                 return total_loss
 
-            self.opt.step(closure)
+            def closure_mixed():
+                with torch.autocast(device_type="cuda", dtype=torch.float16):
```
lululxvi, line 378: Why only cuda?

g-w1: Good point, I have changed it to use the device type of the network.

deepxde/model.py
```diff
@@ -374,7 +374,11 @@ def closure():
                 total_loss.backward()
                 return total_loss
 
-            self.opt.step(closure)
+            def closure_mixed():
+                with torch.autocast(device_type=next(self.net.parameters()).device.type, dtype=torch.float16):
```
lululxvi, line 378: `next(self.net.parameters()).device.type` seems complicated. Is there a simpler function for torch to get the current device?

g-w1: In general no since a network could have different parameters on different devices (see https://github.com/pytorch/pytorch/issues/7460). But in this case since we can since we are only looking for the device type and deepxde doesn't have different parameters on different devices. I've changed the code to use `get_default_device`.

PR #1654, opened 2024-02-16
ZCS Examples

kuangdai:
This PR creates a few new examples in DeepXDE.

Files added:

* `examples/operator/diff_rec_zcs_pideeponet.py`: diffusion reaction with ZCS.
* `examples/operator/stokes_aligned_pideeponet.py`: Stokes flow with alighned points. Note that this is the first example with multiple outputs $(u, v, p)$, regardless of ZCS.
* `examples/operator/stokes_zcs_pideeponet.py`: Stokes flow with ZCS. 
* `examples/dataset/stokes.npz`: one true solution by FEM for evaluation and visualisation of the Stokes example.

A compelte document has been prepared, which is not included here because docs should better be submitted in another PR. However, please find the document [**here**](https://github.com/stfc-sciml/DeepXDE-ZCS/blob/main/zcs.md), which will help you to quickly understand and assess the new examples.


kuangdai:
I don't why this PR include previous commits. Re-submitting another one later.

PR #1655, opened 2024-02-16
New examples to demo ZCS and multiple outputs

kuangdai:
This PR creates a few new examples in DeepXDE.

Files added:

* `examples/operator/diff_rec_zcs_pideeponet.py`: diffusion reaction with ZCS.
* `examples/operator/stokes_aligned_pideeponet.py`: Stokes flow with alighned points. Note that this is the first example with multiple outputs $(u, v, p)$, regardless of ZCS.
* `examples/operator/stokes_zcs_pideeponet.py`: Stokes flow with ZCS. 
* `examples/dataset/stokes.npz`: one true solution by FEM for evaluation and visualisation of the Stokes example.

A compelte document has been prepared, which is not included here because docs should better be submitted in another PR. However, please find the document [**here**](https://github.com/stfc-sciml/DeepXDE-ZCS/blob/main/zcs.md), which will help you to quickly understand and assess the new examples.


examples/operator/diff_rec_zcs_pideeponet.py
```diff
@@ -0,0 +1,89 @@
+"""Backend supported: tensorflow, pytorch, paddle"""
```
lululxvi, line 1: Change the filename to diff_rec_pideeponet_zcs.py

examples/operator/stokes_zcs_pideeponet.py
```diff
@@ -0,0 +1,123 @@
+"""Backend supported: tensorflow, pytorch"""
```
lululxvi, line 1: Change the filename to stokes_pideeponet_zcs.py

examples/operator/diff_rec_zcs_pideeponet.py
lululxvi: Good point. I prefer `diff_rec_aligned_zcs_pideeponet`,

examples/operator/stokes_aligned_pideeponet.py
```diff
@@ -0,0 +1,119 @@
+"""Backend supported: tensorflow, pytorch"""
+import matplotlib.pyplot as plt
+import numpy as np
+
+import deepxde as dde
```
lululxvi, line 5:
```
import deepxde as dde
import numpy as np
```

PR #1661, opened 2024-02-22
Add NNCG to optimizers submodule

pratikrathore8:


lululxvi:
Format the code using black https://github.com/psf/black

pratikrathore8:
I've formatted the code with black in the latest commit!

pratikrathore8:
Moved to the right folder!

deepxde/optimizers/pytorch/nys_newton_cg.py
```diff
@@ -0,0 +1,312 @@
+import torch
+from torch.optim import Optimizer
+from torch.func import vmap
+from functools import reduce
```
lululxvi, line 4: Reorder
```
from functools import reduce

import torch
from torch.func import vmap
from torch.optim import Optimizer
```

deepxde/optimizers/pytorch/nys_newton_cg.py
```diff
@@ -0,0 +1,312 @@
...
+    z = (lambd_r + mu) * (U @ (S_mu_inv * z)) + (x - U @ z)
+    return z
+
+
+def _nystrom_pcg(hess, b, x, mu, U, S, r, tol, max_iters):
+    """Solves a positive-definite linear system using NyströmPCG.
+
+    `Frangella et al. Randomized Nyström Preconditioning.
+    SIAM Journal on Matrix Analysis and Applications, 2023.
+    <https://epubs.siam.org/doi/10.1137/21M1466244>`"""
```
lululxvi, line 29:
```
    <https://epubs.siam.org/doi/10.1137/21M1466244>`
    """
```

deepxde/optimizers/pytorch/nys_newton_cg.py
```diff
@@ -0,0 +1,312 @@
...
+            parameter groups
+        lr (float, optional): learning rate (default: 1.0)
+        rank (int, optional): rank of the Nyström approximation (default: 10)
+        mu (float, optional): damping parameter (default: 1e-4)
+        chunk_size (int, optional): number of Hessian-vector products to be computed in parallel (default: 1)
+        cg_tol (float, optional): tolerance for PCG (default: 1e-16)
+        cg_max_iters (int, optional): maximum number of PCG iterations (default: 1000)
+        line_search_fn (str, optional): either 'armijo' or None (default: None)
+        verbose (bool, optional): verbosity (default: False)
+
```
lululxvi, line 95: Delete this empty line.

deepxde/optimizers/pytorch/nys_newton_cg.py
```diff
@@ -0,0 +1,312 @@
...
+        self._params = self.param_groups[0]["params"]
+        self._params_list = list(self._params)
+        self._numel_cache = None
+
+    def step(self, closure=None):
+        """Perform a single optimization step.
+
+        Args:
+            closure (callable, optional): A closure that reevaluates the model and returns (i) the loss and (ii) gradient w.r.t. the parameters.
+            The closure can compute the gradient w.r.t. the parameters by calling torch.autograd.grad on the loss with create_graph=True.
```
lululxvi, line 148: These two lines are too long. max 88 characters per line

pratikrathore8:
Made the formatting changes. I noticed there are other lines in the code with > 88 chars per line. Would you like for me to fix those?

lululxvi:
> Made the formatting changes. I noticed there are other lines in the code with > 88 chars per line. Would you like for me to fix those?

Yes.

Could you also fix the Codacy issues?

pratikrathore8:
The formatting has been fixed.

There are still some Codacy issues, but these are mainly due to "too many local variables" and the __init__ function in the optimizer, which is just following PyTorch conventions.

What do we need do next to further integrate NNCG into the codebase?

lululxvi:
Need to add the NNCG to https://github.com/lululxvi/deepxde/blob/master/deepxde/optimizers/pytorch/optimizers.py

pratikrathore8:
I've added NNCG to optimizers.py and NNCG_options to config.py

deepxde/optimizers/pytorch/optimizers.py
```diff
@@ -2,33 +2,49 @@
...
+    return optimizer in ["L-BFGS", "L-BFGS-B", "NNCG"]
 
 
 def get(params, optimizer, learning_rate=None, decay=None, weight_decay=0):
     """Retrieves an Optimizer instance."""
     # Custom Optimizer
     if isinstance(optimizer, torch.optim.Optimizer):
         optim = optimizer
-    elif optimizer in ["L-BFGS", "L-BFGS-B"]:
+    elif optimizer in ["L-BFGS", "L-BFGS-B", "NNCG"]:
```
lululxvi, line 18:
```
elif optimizer in ["L-BFGS", "L-BFGS-B"]:
    ...
elif optimizer == "NNCG":
    ...
```

pratikrathore8:
Fixed!

deepxde/optimizers/pytorch/optimizers.py
```diff
@@ -2,11 +2,12 @@
 
 import torch
 
-from ..config import LBFGS_options
+from ..config import LBFGS_options, NNCG_options
+from .nncg import NNCG
```
lululxvi, line 6:
```
from .nncg import NNCG
from ..config import LBFGS_options, NNCG_options
```

pratikrathore8:
Fixed!

lululxvi:
Have you tried using NNCG in some demo examples?

pratikrathore8:
A quick update:

I've started thinking about how to make a demo example that uses NNCG. The plan would be to modify some of the existing examples (e.g. Burgers.py) and add on NNCG after Adam and L-BFGS.

I'll have to make some modifications to model.py to full integrate NNCG into the training pipeline. I'm aiming to have an example fully working by the end of next week!

pratikrathore8:
I've added a demo `Burgers_NNCG.py` in `examples` that uses NNCG. NNCG consistently lowers the loss after Adam+L-BFGS, although there is variance in the amount of improvement (which is probably due to not setting the random seed).

To fully integrate NNCG into the codebase, I added a function called `_train_pytorch_nncg` to `model.py`. I've left comments in `_train_pytorch_nncg` to describe my thought process.

Let me know what changes need to be made! I'm also happy to add more demos if you have suggestions.

pratikrathore8:
@lululxvi Any updates on the pull request?

deepxde/model.py
```diff
@@ -363,11 +363,27 @@ def closure():
...
+            def closure():
+                return get_loss_grad_nncg(inputs, targets, auxiliary_vars)
+
+            self.opt.step(closure)
+
+        def get_loss_grad_nncg(inputs, targets, auxiliary_vars):
+            losses = outputs_losses_train(inputs, targets, auxiliary_vars)[1]
+            total_loss = torch.sum(losses)
+            self.opt.zero_grad()
+            grad_tuple = torch.autograd.grad(total_loss, trainable_variables,
```
lululxvi, line 376: Is this necessary? In the code of L-BFGS https://pytorch.org/docs/stable/_modules/torch/optim/lbfgs.html#LBFGS, this seems to be computed in `step()` by `self._gather_flat_grad()`.

lululxvi:
> @lululxvi Any updates on the pull request?

Sorry for the late response. There are a lot of changes. Any ways to simplify? Or move more code to nncg.py.

pratikrathore8:
@lululxvi Sorry for not responding -- I was away from Stanford during the summer. My plan is to take a look at the code next week!

pratikrathore8:
@lululxvi My apologies for taking so long. I've updated the code -- I moved most of the complexity from `model.py` into `nncg.py`. 

The main thing to note is that I added an argument `perform_backward` to `train_step` for the PyTorch backend -- this argument is `True` by default and is set to `False` only in `_train_pytorch_nncg`. The reason for adding this argument is because NNCG requires Hessian-vector products (hvps), which necessitates a slightly differently implementation from other optimizers, as I have explained below:

> We can obtain hvps by either using `.backward()` with `create_graph=True` followed by `torch.autograd.grad` or by simply using `torch.autograd.grad` twice. 
>
> However, I find that the approach using `.backward()` runs out of memory; the memory of NNCG increases with the number of iterations instead of remaining constant, which is undesirable. I believe this is because using `create_graph=True` with `.backward()` leads to the computation graph being retained across iterations instead of being deleted at every iteration.
>
> On the other hand using `torch.autograd.grad` twice to compute hvps leads to constant memory across iterations.
>
> The memory issues with `.backward()` **might** be fixable with a lot of debugging, but I don't think it's worth it -- even PyTorch throws a warning about possible memory leaks when using the `.backward()` approach to calculate hvps.

Let me know if you'd like for me to make more changes -- I'd be interested in knowing if the tutorial `Burgers_NNCG.py` in `examples` works on your end. 

The errors in the Codacy analysis are from `nncg.py`, but these are minor issues related to PyTorch optimizer conventions and too many local variables (which is difficult to avoid in the implementation).

lululxvi:
Could you fix the other Codacy issues of dict and super-with-arguments?

deepxde/model.py
```diff
@@ -643,12 +644,24 @@ def train(
         self._test()
         self.callbacks.on_train_begin()
         if optimizers.is_external_optimizer(self.opt_name):
+            if self.opt_name == "NNCG" and backend_name != "pytorch":
```
lululxvi, line 647: I believe this error checking is not needed.

deepxde/model.py
```diff
@@ -643,12 +644,24 @@ def train(
...
             elif backend_name == "tensorflow":
                 self._train_tensorflow_tfp()
             elif backend_name == "pytorch":
-                self._train_pytorch_lbfgs()
+                if self.opt_name == "L-BFGS":
+                    self._train_pytorch_lbfgs()
+                elif self.opt_name == "NNCG":
+                    self._train_pytorch_nncg(iterations, display_every)
+                else:
+                    raise ValueError(
```
lululxvi, line 661: This is also not needed.

deepxde/model.py
pratikrathore8: Done!

deepxde/model.py
pratikrathore8: Done!

pratikrathore8:
@lululxvi I've removed the unnecessary error checking and fixed the Codacy issues of dict and super-with-arguments.

examples/pinn_forward/Burgers_NNCG.py
```diff
@@ -0,0 +1,73 @@
+"""Backend supported: tensorflow.compat.v1, tensorflow, pytorch, paddle"""
```
lululxvi, line 1: """Backend supported: pytorch"""

deepxde/model.py
```diff
@@ -792,6 +796,30 @@ def _train_pytorch_lbfgs(self):
             if self.stop_training:
                 break
 
+    def _train_pytorch_nncg(self, iterations, display_every):
+        for i in range(iterations):
+            self.callbacks.on_epoch_begin()
+            self.callbacks.on_batch_begin()
+
+            # The train step should only use full gradients, so we do not use self.train_state.set_data_train()
```
lululxvi, line 804: Should still add `self.train_state.set_data_train(*self.data.train_next_batch(self.batch_size))`, which uses full data.

pratikrathore8:
@lululxvi Made the changes. Should we consider merging `_train_pytorch_nncg` into `_train_sgd` by adding an argument to `_train_sgd` that decides whether or not to perform a backward pass on the loss? Or would you prefer we leave the functions separate, despite the code duplication?

deepxde/model.py
```diff
@@ -353,12 +353,13 @@ def outputs_losses_test(inputs, targets, auxiliary_vars):
                     "backend pytorch."
                 )
 
-        def train_step(inputs, targets, auxiliary_vars):
+        def train_step(inputs, targets, auxiliary_vars, perform_backward=True):
```
lululxvi, line 356: Instead of adding an argument here, add a new `train_step_nncg`, similar to 
https://github.com/lululxvi/deepxde/blob/40cd7e5af35b6b538cc1b9b15cc7a603975df397/deepxde/model.py#L253

pratikrathore8:
@lululxvi I added a `train_step_nncg` function as you suggested. NNCG now runs using `_train_sgd`. 

Thanks for your patience with the refactoring steps so far!

deepxde/optimizers/config.py
```diff
@@ -60,6 +61,57 @@ def set_LBFGS_options(
...
+    rank=10,
+    mu=1e-4,
+    updatefreq=20,
+    chunksz=1,
+    cgtol=1e-16,
+    cgmaxiter=1000,
+    lsfun="armijo",
+    verbose=False,
+):
+    """Sets the hyperparameters of NysNewtonCG (NNCG).
```
lululxvi, line 75: Add a comment here that only supports PyTorch.

deepxde/optimizers/config.py
```diff
@@ -60,6 +61,57 @@ def set_LBFGS_options(
...
+
+    Args:
+        lr (float): `lr` (torch).
```
lululxvi, line 78: You don't need
```
`lr` (torch).
```
For L-BFGS, the reason of this line
```
maxcor (int): `maxcor` (scipy), `num_correction_pairs` (tfp), `history_size` (torch), `history_size` (paddle).
```
is that we directly use the BFGS from different backends. But different backends use different names, so we indicate the corresponding name in the backend API. However, NNCG is not from any backend package.

pratikrathore8:
@lululxvi Done!

lululxvi:
Add a document for the example similar to this https://deepxde.readthedocs.io/en/latest/demos/pinn_forward/burgers.html

pratikrathore8:
@lululxvi I added the demo. It's basically the same as the one you linked, except I added the part with NNCG to the end.

lululxvi:
Please resolve the conflict.

You can look at the generated doc by clicking "Details" at the right side of "docs/readthedocs.org:deepxde" in in the check list.

docs/demos/pinn_forward.rst
```diff
@@ -52,6 +52,7 @@ Time-dependent PDEs
    pinn_forward/burgers.rar
    pinn_forward/allen.cahn
    pinn_forward/klein.gordon
+   pinn_forward/burgers.nncg
```
lululxvi, line 55: Move this after pinn_forward/burgers
```
pinn_forward/burgers
pinn_forward/burgers.nncg
pinn_forward/heat
```

examples/pinn_forward/Burgers_NNCG.py
```diff
@@ -0,0 +1,73 @@
+"""Backend supported: pytorch"""
```
lululxvi, line 1: I am thinking if it is possible/better to combine this file with the existing Buerger file. Would this be better?

docs/demos/pinn_forward.rst
pratikrathore8: Done!

examples/pinn_forward/Burgers_NNCG.py
pratikrathore8: I agree with you that there is a lot of similarity between `Burgers_NNCG.py` and `Burgers.py`. 

However, if we combine `Burgers_NNCG.py` with `Burgers.py`, we would limit `Burgers.py` to only the PyTorch backend, no?

Also, NNCG is a bit slower than the other optimizers, so this could be frustrating for a practictioner who wants to run a simple demo.

Let me know what you think!

lululxvi: We can comment out the NNCG part, and add comments that this NNCG is optional, and it will further reduce the error, and it only supports pytorch, etc.

pratikrathore8: Ah, I see. I made a new commit that merges NNCG into the original Burgers demo. Let me know if you'd like me to make further improvements!

examples/pinn_forward/Burgers.py
```diff
@@ -1,4 +1,5 @@
 """Backend supported: tensorflow.compat.v1, tensorflow, pytorch, paddle"""
+
```
lululxvi, line 2: Delete this line

examples/pinn_forward/Burgers.py
```diff
@@ -46,3 +47,20 @@ def pde(x, y):
 print("Mean residual:", np.mean(np.absolute(f)))
 print("L2 relative error:", dde.metrics.l2_relative_error(y_true, y_pred))
 np.savetxt("test.dat", np.hstack((X, y_true, y_pred)))
+
+# """Backend supported: pytorch"""
+# # Run NNCG after Adam and L-BFGS
+# dde.optimizers.set_NNCG_options(rank=50, mu=1e-1)
```
lululxvi, line 53: If `rank=50, mu=1e-1` is needed, then should we set them as the default in config.py?

examples/pinn_forward/Burgers.py
```diff
@@ -46,3 +47,20 @@ def pde(x, y):
...
+# model.compile("NNCG")
```
lululxvi, line 54: Delete this code from here, and move the new code before line 44.

pratikrathore8:
These are all great suggestions! I've incorporated them into my most recent commit.

examples/pinn_forward/Burgers.py
```diff
@@ -39,6 +39,12 @@ def pde(x, y):
 model.compile("L-BFGS")
 losshistory, train_state = model.train()
 dde.saveplot(losshistory, train_state, issave=True, isplot=True)
+# """Backend supported: pytorch"""
+# # Run NNCG after Adam and L-BFGS
+# dde.optimizers.set_NNCG_options(rank=50, mu=1e-1)
+# model.compile("NNCG")
+# losshistory, train_state = model.train(iterations=1000, display_every=100)
+# dde.saveplot(losshistory, train_state, issave=True, isplot=True)
```
lululxvi, line 47:
```
model.compile("L-BFGS")
losshistory, train_state = model.train()
# """Backend supported: pytorch"""
# # Run NNCG after Adam and L-BFGS
# dde.optimizers.set_NNCG_options(rank=50, mu=1e-1)
# model.compile("NNCG")
# losshistory, train_state = model.train(iterations=1000, display_every=100)
dde.saveplot(losshistory, train_state, issave=True, isplot=True)
```

pratikrathore8:
Good catch! I made the fix in my most recent commit.

PR #1670, opened 2024-03-06
Add loss weights feature for jax backend

bonneted:


deepxde/model.py
```diff
@@ -402,6 +400,8 @@ def outputs_fn(inputs):
             if not isinstance(losses, list):
                 losses = [losses]
             losses = jax.numpy.asarray(losses)
+            if self.loss_weights is not None:
```
lululxvi, line 403: Update the TODO comment in line 399

Issue #1558, opened 2023-11-14
Type mismatch when trying to use L-BFGS


jdellag:
I've been working on a Navier Stokes problem and would like to try to further optimize my model after training with Adam like many of the examples I have seen. However, when trying to use this optimizer after setting the default precision to float64, I am receiving the following error:

```
Compiling model...

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[18], line 18
     15 for beta in beta_values:
     16     print(f"INTIALIZING RUN ******* {frame} ** Beta = {beta} ** Re = {Re} *******")
---> 18     losshistory, train_state, model = train_for_Re(Re, beta)
     19     # Update the progress count
     20     progress_count += 1

Cell In[17], line 38, in train_for_Re(Re, beta)
     32 variable = dde.callbacks.VariableValue(U, period = 500, filename = fnamevar, precision = 32)
     34 #losshistory, train_state = model.train(iterations = num_iter, callbacks = [variable], display_every = 500)
     35 #dde.utils.external.saveplot(losshistory, train_state, 
     36                             #issave = True, isplot = True, train_fname = train_filename, 
     37                             #    test_fname = test_filename, loss_fname = loss_filename, output_dir = "run_data")
---> 38 model.compile('L-BFGS-B')
     39 losshistory, train_state = model.train(iterations = num_iter, callbacks = [variable])
     41 return losshistory, train_state, model

File ~/python_environments/deepxde2/lib/python3.11/site-packages/deepxde/utils/internal.py:22, in timing.<locals>.wrapper(*args, **kwargs)
     19 @wraps(f)
     20 def wrapper(*args, **kwargs):
     21     ts = timeit.default_timer()
---> 22     result = f(*args, **kwargs)
     23     te = timeit.default_timer()
     24     if config.rank == 0:

File ~/python_environments/deepxde2/lib/python3.11/site-packages/deepxde/model.py:137, in Model.compile(self, optimizer, lr, loss, metrics, decay, loss_weights, external_trainable_variables)
    134     self.external_trainable_variables = external_trainable_variables
    136 if backend_name == "tensorflow.compat.v1":
--> 137     self._compile_tensorflow_compat_v1(lr, loss_fn, decay)
    138 elif backend_name == "tensorflow":
    139     self._compile_tensorflow(lr, loss_fn, decay)

[...]

PR #1684, opened 2024-03-25
Issue #1558: Refactor Variable function in tensor.py

jdellag:
Changed the Variable function to inherit its datatype from config.set_default_float by examining tf.keras.backend.floatx() for the tensorflow_v1_compat backend.


deepxde/backend/tensorflow_compat_v1/tensor.py
```diff
@@ -3,7 +3,6 @@
 
 import tensorflow.compat.v1 as tf
 
-
```
lululxvi, line 6: Keep this line.

PR #1689, opened 2024-03-27
Jax VariableValue callback

bonneted:
This is not the cleanest since it has to reset the model at each iteration to access the variables through self.model.external_trainable_variables. 
Maybe there is a better way using a similar approach as tf v1 ?

### DIFF
--- a/deepxde/callbacks.py
+++ b/deepxde/callbacks.py
@@ -330,11 +330,8 @@ class VariableValue(Callback):
             self.value = [var.numpy() for var in self.var_list]
         elif backend_name in ["pytorch", "paddle"]:
             self.value = [var.detach().item() for var in self.var_list]
-        if backend_name == "jax" and any(
-                type(callback).__name__ == "VariableValue"
-                for callback in self.callbacks.callbacks
-            ):
-                self.callbacks.set_model(self)
+        elif backend_name == "jax":
+            self.value = [var.item() for var in self.model.external_trainable_variables]
         print(
             self.model.train_state.epoch,
             utils.list_to_str(self.value, precision=self.precision),
--- a/deepxde/callbacks.py
+++ b/deepxde/callbacks.py
@@ -330,6 +330,11 @@ class VariableValue(Callback):
             self.value = [var.numpy() for var in self.var_list]
         elif backend_name in ["pytorch", "paddle"]:
             self.value = [var.detach().item() for var in self.var_list]
+        if backend_name == "jax" and any(
+                type(callback).__name__ == "VariableValue"
+                for callback in self.callbacks.callbacks
+            ):
+                self.callbacks.set_model(self)
         print(
             self.model.train_state.epoch,
             utils.list_to_str(self.value, precision=self.precision),
--- a/deepxde/model.py
+++ b/deepxde/model.py
@@ -661,6 +661,11 @@ class Model:
         for i in range(iterations):
             self.callbacks.on_epoch_begin()
             self.callbacks.on_batch_begin()
+            if backend_name == "jax" and any(
+                type(callback).__name__ == "VariableValue"
+                for callback in self.callbacks.callbacks
+            ):
+                self.callbacks.set_model(self)
 
             self.train_state.set_data_train(
                 *self.data.train_next_batch(self.batch_size)


deepxde/callbacks.py
```diff
@@ -330,6 +330,8 @@ def on_train_begin(self):
             self.value = [var.numpy() for var in self.var_list]
         elif backend_name in ["pytorch", "paddle"]:
             self.value = [var.detach().item() for var in self.var_list]
+        elif backend_name == "jax":
+            self.value = [var.item() for var in self.model.external_trainable_variables]
```
lululxvi, line 334: Why not using `var_list`?

lululxvi: um, this is strange, as in https://github.com/lululxvi/deepxde/discussions/760#discussioncomment-8926115, `model.external_trainable_variables` and `var_list` both point to `external_trainable_variables`.

In this callback, it is designed to only show `var_list`. Need to think about this. If this is not doable, then need to show a warning here.

bonneted: I tried with var_list in the callback, it writes always the same (=initial) value.

lululxvi: Do you have any idea why?

bonneted: Found it : jax arrays are immutable, so a new instance of self.external_trainable_variables was created on every update of the parameters :
```
#model.py line 563
  elif backend_name == "jax":
      # TODO: auxiliary_vars
      self.params, self.opt_state = self.train_step(
          self.params, self.opt_state, inputs, targets
      )
      self.net.params, self.external_trainable_variables = self.params
```
To solve it I created an arrayWrapper class for variableValue in jax.
The array can be accessed using variable.arr

It works fine on the Lorenz example.

deepxde/backend/jax/tensor.py
```diff
@@ -1,4 +1,5 @@
 """jax backend implementation"""
+
```
lululxvi, line 2: Remove this empty line.

deepxde/model.py
```diff
@@ -661,7 +666,7 @@ def _train_sgd(self, iterations, display_every):
         for i in range(iterations):
             self.callbacks.on_epoch_begin()
             self.callbacks.on_batch_begin()
-
+
```
lululxvi, line 669: Remove this line.

deepxde/backend/jax/tensor.py
```diff
@@ -41,8 +42,14 @@ def reshape(tensor, shape):
...
 
+class ArrayWrapper:
+    def __init__(self, arr):
+        self.arr = arr
+
+
 def Variable(initial_value, dtype=None):
-    return jnp.array(initial_value, dtype=dtype)
+    arr = jnp.array(initial_value, dtype=dtype)
+    return ArrayWrapper(arr)
```
lululxvi, line 52: How about something like this?
```
class Variable:
    def __init__(self, initial_value, dtype=None):
        self._value = jnp.array(initial_value, dtype=dtype)

    @property
    def value(self):
        return self._value

    @value.setter(self, value):
        self._value = value
```

bonneted: I've refactored the code, it works fine on the Lorenz inverse example.

deepxde/model.py
lululxvi: I mean change this line to the original one.

bonneted: yes sorry, should be like the original now

deepxde/backend/jax/tensor.py
```diff
@@ -41,8 +41,17 @@ def reshape(tensor, shape):
     return jnp.reshape(tensor, shape)
 
 
-def Variable(initial_value, dtype=None):
-    return jnp.array(initial_value, dtype=dtype)
+class Variable:
```
lululxvi, line 44: Would this break the existing code?

For example,
https://github.com/lululxvi/deepxde/blob/862b67f671fcdb661a97feb0b087eba906137da9/examples/pinn_inverse/Lorenz_inverse.py#L27
Now, should use `C1.value`?

lululxvi: If it works, then it is great and no need to make it explicit. But I want to understand why it works to ensure everything works as expected.

bonneted: Ok, I got it. 
The way inverse quantification is implemented in JAX, you have to pass the unknown variables as an extra `unknowns` input in your pde : 
https://github.com/lululxvi/deepxde/blob/63de873c4096ddcacc7941cafdeabe4b44525e09/deepxde/data/pde.py#L144-L153

I think it's because auxiliary functions are not implemented in jax.

Then what is used as unknowns input for the pde (`self.params`) is already jax array and not the `Variable` instance. This is done in `model.py` : 
https://github.com/lululxvi/deepxde/blob/63de873c4096ddcacc7941cafdeabe4b44525e09/deepxde/model.py#L378-L381

And then the `Variable` instances are updated at each step to be accessible for the callback :
https://github.com/lululxvi/deepxde/blob/63de873c4096ddcacc7941cafdeabe4b44525e09/deepxde/model.py#L571-L573

So the net and the pde only see the jax array and not the "Variable" instance, which is only updated in the `model.py` file.
That's why we don't need to use `C1.value`

bonneted:
I just added the L-BFGS is currently not implemented for JAX.
But hopefully it will be the case soon as jaxopt (which support L-BFGS) is currenlty being merged into optax, the optimization library used by DeepXDE:
https://github.com/google/jaxopt

PR #1717, opened 2024-04-19
Jax transform bug

bonneted:
The previous update to handle transform with Jax was not working to compute the Hessian.
In that case, the output of the Jacobian is reused to compute the Hessian. There was an index error because squeeze make a one element array 0D but we actually want to keep one dimension. Changing the reshape(-1) solved it :
```
function       |  x.shape
.squeeze:        (1,1) --> ()
.reshape(-1):    (1,1) --> (1,)
```
I also implemented Jax on Helmotz equation example

### DIFF
--- a/examples/pinn_forward/Helmholtz_Dirichlet_2d.py
+++ b/examples/pinn_forward/Helmholtz_Dirichlet_2d.py
@@ -1,6 +1,7 @@
-"""Backend supported: tensorflow.compat.v1, tensorflow, pytorch, paddle"""
+"""Backend supported: tensorflow.compat.v1, tensorflow, pytorch, paddle, jax"""
 import deepxde as dde
 import numpy as np
+import deepxde.backend as bkd
 
 # General parameters
 n = 2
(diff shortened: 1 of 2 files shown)


examples/pinn_forward/Helmholtz_Dirichlet_2d.py
```diff
@@ -1,6 +1,7 @@
-"""Backend supported: tensorflow.compat.v1, tensorflow, pytorch, paddle"""
+"""Backend supported: tensorflow.compat.v1, tensorflow, pytorch, paddle, jax"""
 import deepxde as dde
 import numpy as np
+import deepxde.backend as bkd
```
lululxvi, line 4: No need to import bkd. Use `dde.bkd`

examples/pinn_forward/Helmholtz_Dirichlet_2d.py
```diff
@@ -1,6 +1,7 @@
-"""Backend supported: tensorflow.compat.v1, tensorflow, pytorch, paddle"""
+"""Backend supported: tensorflow.compat.v1, tensorflow, pytorch, paddle, jax"""
```
lululxvi, line 1: """Backend supported: tensorflow.compat.v1, tensorflow, pytorch, jax, paddle"""

Issue #1754, opened 2024-05-24
ImportError check_pandas_support


aoguedao:
Hi Dr. Lu and DeepXDE team - 

I just created a new environment for working with deepxde using JAX as a backend and I got the following error after changing the backend with `python -m deepxde.backend.set_default_backend jax`

```
ImportError: cannot import name 'check_pandas_support' from 'sklearn.utils' (/mnt/c/Users/aogueda/Documents/git/misc/sciml/lib/python3.12/site-packages/sklearn/utils/__init__.py)
```

I think it is because 'check_pandas_support' has a new path according to [this commit](https://github.com/scikit-learn/scikit-learn/commit/e125e67bce7811eefb6f449918dc6aba596678ee) in scikit-learn.

I am using WSL2, miniforge for virtual environments and I installed scikit-learn using conda/mamba but JAX and DeepXDE with pip as the final step of setting my environment. 

Thanks in advance.

aoguedao:
> Hi aoguedao,
> 
> interestingly I ran into the exact same problem today. You have to downgrade scikit-learn to Version 1.3.0.
> 
> I also opened an issue today but closed it. There you can lookt it up.

Thank you! I'll downgrade scikit-learn for now then. Appreciate it!

entropy2333:
scikit-learn removes the `check_pandas_support` according to [here](https://github.com/scikit-learn/scikit-learn/pull/28482), downgrade to `scikit-learn <=1.4.2 ` can solve the problem

AWBroscius:
@entropy2333 Thank you very much, changing the path clears the issue right up! I was having this same issue trying to start working with the Deepxde package yesterday.

PR #1760, opened 2024-05-30
Fix requirements.txt

vl-dud:
See [issue](https://github.com/lululxvi/deepxde/issues/1754)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 matplotlib
 numpy
-scikit-learn
+scikit-learn<=1.4.2
 scikit-optimize>=0.9.0
 scipy


requirements.txt
```diff
@@ -1,5 +1,5 @@
 matplotlib
 numpy
-scikit-learn
+scikit-learn<=1.4.2
```
lululxvi, line 3: Add a comment why this is needed.

vl-dud: This is needed to remove errors like `ImportError: cannot import name 'check_pandas_support' from 'sklearn.utils'` from CI checks

lululxvi:
@vl-dud The issue still exists...

vl-dud:
> @vl-dud The issue still exists...

@lululxvi As far as I can see, my new [PR](https://github.com/lululxvi/deepxde/pull/1769) went through correctly

PR #1772, opened 2024-06-14
Update scale factor for variable value

bonneted:
Sometimes the dynamic for learning the system parameters is not the same as the weight and biases of the NN.
This introduces a scale factor to multiply the step calculated by the optimizer to update the variables.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,7 +46,7 @@ jobs:
         env:
           DDE_BACKEND: tensorflow
         run: |
-          python -m pip install tensorflow-probability
+          python -m pip install tensorflow-probability[tf]
           python -c "import deepxde"
 
       - name: Test build with PyTorch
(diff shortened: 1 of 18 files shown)


lululxvi:
It is not only scaling. In many cases, it requires a transformation. Does this code work in JAX?
https://github.com/lu-group/sbinn/blob/b2c1c94d6564732189722f6e6772af0f63cb0d8c/sbinn/sbinn_tf.py#L8

bonneted:
I implemented jax for sbinn, it works fine. This allowed me to spot two bugs for external variables with jax, I opened a new PR to fix them.
I've also  opened a PR in the sbinn repo for the jax implementation if you're interested.

lululxvi:
Is this PR necessary? As the user can easily implement a linear transformation.

bonneted:
Yes you're right, linear transformation works fine. Thank you for letting me know.

For your information, here is the damping effect I was facing because of the too high learning rate : (inverse quantification for linear elasticity)
image

Improved with a scale factor of 0.1 :
image

Same results doing a linear transformation :
image

PR #1775, opened 2024-06-17
fix external variable initialization

bonneted:
I've faced two bugs when trying to implement :
https://github.com/lu-group/sbinn/blob/b2c1c94d6564732189722f6e6772af0f63cb0d8c/sbinn/sbinn_tf.py#L8

- in `model.py` it's because the external variables were not initialized on the second compile as the parameters of the net were already

- in `pde.py` if I don't compile with external variables I still want the code to work with the default values of unknowns
I think this code can be safely modified only for jax because the line after was already only for jax


deepxde/data/pde.py
```diff
@@ -147,8 +147,9 @@ def losses(self, targets, outputs, loss_fn, inputs, model, aux=None):
             elif get_num_args(self.pde) == 3:
                 if self.auxiliary_var_fn is None:
                     if aux is None or len(aux) == 1:
-                        raise ValueError("Auxiliary variable function not defined.")
-                    f = self.pde(inputs, outputs_pde, unknowns=aux[1])
+                        f = self.pde(inputs, outputs_pde)
```
lululxvi, line 150: Why modify here? This is an error checking.
When changing code for JAX, it needs to ensure that it doesn't affect other backends.

bonneted: The error is raised but shouldn't in that case :
`self.pde` is : `def ODE(t, y, unknowns=[var.value for var in var_list_])` with 3 arguments
but we compiled with no external variables for the first train :
```
model.compile("adam", lr=1e-3, loss_weights=[0, 0, 0, 0, 0, 0, 1e-2])
model.train(epochs=firsttrain, display_every=1000)
```
So `len(aux) == 1` and the error is raised.
But we want the code to run with the default unknown values for the ODE, that's why I've changed to `f = self.pde(inputs, outputs_pde)`

I think this code could be changed safely, because it's only for JAX.
The `unknowns` extra parameter was introduced specifically for JAX, I think that's the only way to have 3 parameters for the pde. #760 
Moreover if the error is not raised, ` f = self.pde(inputs, outputs_pde, unknowns=aux[1])` is called, which is JAX specific

bonneted: Sorry, I didn't know about this other implementation. 
I've updated the code to check if the 3rd argument has a default value, if so the function is called with this default value.

bonneted:
Here : https://github.com/bonneted/sbinn/blob/main/sbinn/sbinn_jax.py

The implementation of sbinn using JAX.
We first train the model without the external variables :
```
    def ODE(t, y, unknowns=[var.value for var in var_list_]):
    ...

    model.compile("adam", lr=1e-3, loss_weights=[0, 0, 0, 0, 0, 0, 1e-2])
    model.train(epochs=firsttrain, display_every=1000)
    model.compile(
        "adam",
        lr=1e-3,
        loss_weights=[1, 1, 1e-2, 1, 1, 1, 1e-2],
        external_trainable_variables=var_list_,
    )
    variablefilename = "variables.csv"
    variable = dde.callbacks.VariableValue(
        var_list_, period=callbackperiod, filename=variablefilename
    )
    losshistory, train_state = model.train(
        epochs=maxepochs, display_every=1000, callbacks=[variable]
    )
```
For this first train, we want to use the default `unknowns` argument for the ODE

lululxvi:
The code modification seems necessary. But there is another example https://github.com/lululxvi/deepxde/blob/master/examples/pinn_inverse/Lorenz_inverse.py , which works well (at least worked earlier).

bonneted:
This one was already working well because there is no pertaining without the external variables.
The model is only compiled with the external trainable variables :
```
model.compile(
    "adam", lr=0.001, external_trainable_variables=external_trainable_variables
)
losshistory, train_state = model.train(iterations=20000, callbacks=[variable])
```

The problem occurs when we compile without the external trainable variables, which is when we want the PDE to use the default `unknowns` argument.

lululxvi:
The code seems OK. But the underlying logic becomes extremely complicated now.

In fact, you can simply add `external_trainable_variables` in the first compile. As the PDE loss weight is 0, those variables won't get updated any way.

bonneted:
That's true in that case, but it can be interesting to start training the model with frozen parameters (for example https://doi.org/10.1126/sciadv.abk0644) 
Moreover, it would mean that putting default `unknowns` values for the `PDE` is useless and misleading as they can never be used.

lululxvi:
Please resolve the conflicts.

deepxde/data/pde.py
```diff
@@ -150,9 +155,14 @@ def losses(self, targets, outputs, loss_fn, inputs, model, aux=None):
...
                 if self.auxiliary_var_fn is not None:
                     f = self.pde(inputs, outputs_pde, model.net.auxiliary_vars)
-                elif backend_name == "jax" and len(aux) == 2:
+                elif backend_name == "jax":
                     # JAX inverse problem requires unknowns as the input.
-                    f = self.pde(inputs, outputs_pde, unknowns=aux[1])
+                    if len(aux) == 2:
+                        # External trainable variables in aux[1] are used for unknowns
+                        f = self.pde(inputs, outputs_pde, unknowns=aux[1])
+                    if len(aux) == 1 and has_default_values(self.pde)[-1]:
```
lululxvi, line 163: - `if` --> `elif`
- How about the case `has_default_values(self.pde)[-1]` is False

bonneted: I've added the case, it means that no trainable variables are provided and no defaults unknowns values are set, leading to an error.

deepxde/utils/internal.py
```diff
@@ -202,6 +202,9 @@ def get_num_args(func):
     params = inspect.signature(func).parameters
     return len(params) - ("self" in params)
 
+def has_default_values(func):
```
lululxvi, line 205: Add a docstring to this function

deepxde/model.py
```diff
@@ -401,10 +401,10 @@ def _compile_jax(self, lr, loss_fn, decay):
...
             key = jax.random.PRNGKey(config.jax_random_seed)
             self.net.params = self.net.init(key, self.data.test()[0])
-            external_trainable_variables_arr = [
-                var.value for var in self.external_trainable_variables
-            ]
-            self.params = [self.net.params, external_trainable_variables_arr]
+        external_trainable_variables_val = [
+            var.value for var in self.external_trainable_variables
+        ]
+        self.params = [self.net.params, external_trainable_variables_val]
```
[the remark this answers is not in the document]
bonneted: it's working fine, var.value is never evaluated and the output is []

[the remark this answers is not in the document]
lululxvi: In this case,
```
self.params = [self.net.params, []]
```
which is strange.

deepxde/utils/internal.py
```diff
@@ -202,6 +202,18 @@ def get_num_args(func):
...
+    """
+    Check if the given function has default values for its parameters.
```
lululxvi, line 207:
```
"""Check if the given function has default values for its parameters.
```

deepxde/model.py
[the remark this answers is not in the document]
bonneted: It was already like that before the introduction `external_trainable_variables_val ` in a previous commit.

It used to be: (see https://github.com/bonneted/deepxde/commit/3810a9888d13e04cecf114e93a00dfbf90021472) 
`self.params = [self.net.params, self.external_trainable_variables]`
with `self.external_trainable_variables = []` by default

We can see how it works in the original commit by @ZongrenZou : https://github.com/bonneted/deepxde/commit/09c74ab6a5c3269e248b09f27518a05d506b0327
`self.params` is split in `nn_params `and `ext_params` in `outputs_losses` function:
```
  def outputs_losses(params, training, inputs, targets, losses_fn):
      nn_params, ext_params = params
      # TODO: Add auxiliary vars
      def outputs_fn(inputs):
          return self.net.apply(nn_params, inputs, training=training)
      outputs_ = self.net.apply(nn_params, inputs, training=training)
      # Data losses
      # We use aux so that self.data.losses is a pure function.
      aux = [outputs_fn, ext_params] if ext_params else [outputs_fn]
      losses = losses_fn(targets, outputs_, loss_fn, inputs, self, aux=aux)
```

deepxde/utils/internal.py
```diff
@@ -202,6 +202,17 @@ def get_num_args(func):
     params = inspect.signature(func).parameters
     return len(params) - ("self" in params)
 
+def has_default_values(func):
```
lululxvi, line 205: Use two empty lines between two functions, so need to add one more line before 205 and one more line after 215.

deepxde/data/pde.py
```diff
@@ -4,7 +4,12 @@
 from .. import backend as bkd
 from .. import config
 from ..backend import backend_name
-from ..utils import get_num_args, run_if_all_none, mpi_scatter_from_rank0
+from ..utils import (
+    get_num_args,
+    has_default_values,
+    run_if_all_none,
+    mpi_scatter_from_rank0,
+)
```
lululxvi, line 12:
```
from ..utils import (
    get_num_args,
    has_default_values,
    mpi_scatter_from_rank0,
    run_if_all_none,
)
```

deepxde/model.py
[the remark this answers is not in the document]
lululxvi: Although the current code works, it is better to make it clear when `self.external_trainable_variables` is []. We can add a `if` here for this case to make the logic better.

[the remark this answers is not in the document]
bonneted: I've added an `if`statement for the `[]` case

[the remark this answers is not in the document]
lululxvi: Sorry, you are right: there is no need to check `self.external_trainable_variables`. Your previous code is good, and you can revert it back.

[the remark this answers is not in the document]
lululxvi: > Sorry, you are right: there is no need to check `self.external_trainable_variables`. Your previous code is good, and you can revert it back.

@bonneted

PR #1776, opened 2024-06-17
Separable-PINN in DeepXDE

bonneted:
This is an implementation of the SPINN model: https://jwcho5576.github.io/spinn.github.io/

The code for the network architecture (`snn.py`) is directly adapted from the original paper (https://github.com/stnamjef/SPINN) 

I've achieved really fast convergence with this implementation of SPINN compared to PINN (similar to the paper claim), for both forward and inverse quantification on the linear elastic plate problem.
Forward comparison :
https://github.com/lululxvi/deepxde/assets/53513604/499a961d-748c-458f-be99-56156b516ace

Inverse with PINN :
https://github.com/lululxvi/deepxde/assets/53513604/ab89554d-b82b-406a-8d73-05b3f72a3961

Inverse with SPINN :
https://github.com/lululxvi/deepxde/assets/53513604/07171442-ea03-48b4-87a6-8b5094f6809c

The implementation was more complicated than expected for the following reasons:
due to its architecture, SPINN takes an input of size n and outputs an array of size n**dim (it does the cartesian product of each coordinate) :
`(n,2) --> SPINN --> n**2`

This brings some difficulty with how inputs are handled in `data.pde`.
Indeed, all inputs are concatenated (PDE and BCS points) and throw the net simultaneously.
So if we have `n_PDE` PDE points and `n_BC` BC points we will end up with `(n_PDE+n_BC)**2` points instead of  `n_PDE**2+n_BC**2`

I tried to find a workaround with minimal changes to `model.py`, and came up with the following:
adding a `list_handler` decorator to the `outputs` function in JAX so that it can handle list inputs by applying the function to each input and then concatenates. 

I then modified the `pde.py` file by adding a `is_SPINN` argument, if true, PDE and BC inputs are put together in a list instead of stacked. The `bcs_start` should also be modified as the outputs sizes no longer equal the inputs.

I understand that this brings a lot of changes to `data.pde`, so another possibility is to create a separate `data` subclass dedicated to SPINN so that the `data.pde` class isn't overly complicated.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,7 +46,7 @@ jobs:
         env:
           DDE_BACKEND: tensorflow
         run: |
-          python -m pip install tensorflow-probability
+          python -m pip install tensorflow-probability[tf]
           python -c "import deepxde"
 
       - name: Test build with PyTorch
--- a/deepxde/backend/jax/tensor.py
+++ b/deepxde/backend/jax/tensor.py
@@ -41,8 +41,17 @@ def reshape(tensor, shape):
     return jnp.reshape(tensor, shape)
 
 
-def Variable(initial_value, dtype=None):
-    return jnp.array(initial_value, dtype=dtype)
+class Variable:
+    def __init__(self, initial_value, dtype=None):
+        self._value = jnp.array(initial_value, dtype=dtype)
+
+    @property
+    def value(self):
+        return self._value
+
+    @value.setter
+    def value(self, new_value):
+        self._value = new_value
 
 
 def as_tensor(data, dtype=None):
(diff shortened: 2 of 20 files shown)


lululxvi:
There are too many modifications. We can start with a mathematically equivalent (but slow speed) implementation by repeating the n inputs to n**2. This is similar to DeepONet
https://github.com/lululxvi/deepxde/blob/ad6399b27c3a0702abd24d45f889ca55497476c2/deepxde/nn/tensorflow/deeponet.py#L17C7-L17C15
vs DeepONetCartesianProd
https://github.com/lululxvi/deepxde/blob/ad6399b27c3a0702abd24d45f889ca55497476c2/deepxde/nn/tensorflow/deeponet.py#L153
Then the code change would be very minimal.

bonneted:
OK, I'll try that when I have more time, hopefully soon.

echen5503:
@bonneted still working on this?

bonneted:
@echen5503 I'm back at it, I'll try to make some progress next week

PR #1781, opened 2024-06-20
[Fix] TF Backend Regularization fix.

agniv-the-marker:
# TF Regularization

Check the [docs](https://www.tensorflow.org/api_docs/python/tf/keras/regularizers/L1) and the argument for `tf.keras.regualizers.l1/l2` is of the form `l1`/`l2` instead of just `l` in the current version. Without it this breaks:

```python
net = dde.maps.PFNN(
    network,
    "swish",
    "Glorot normal",
    regularization=["l2", 1e-8]
)
```

Specifically it gives you the error:

```python
TypeError: L1.__init__() got an unexpected keyword argument 'l'
```

As of right now, the following works instead:

```python
net = dde.maps.PFNN(
    network,
    "swish",
    "Glorot normal",
    regularization=["l1+l2", 1e-8, 1e-8]
)
```

But merging would allow for each one to be used on their own.

### DIFF
--- a/deepxde/nn/regularizers.py
+++ b/deepxde/nn/regularizers.py
@@ -5,13 +5,13 @@ def get(identifier):
     # TODO: other backends
     if identifier is None:
         return None
-    name, scales = identifier[0], identifier[1:]
+    name, scales = identifier[0].lower(), identifier[1:]
     return (
         tf.keras.regularizers.L1(l1=scales[0])
-        if name == "l1" or name == "L1"
+        if name == "l1"
         else tf.keras.regularizers.L2(l2=scales[0])
-        if name == "l2" or name == "L2"
+        if name == "l2"
         else tf.keras.regularizers.L1L2(l1=scales[0], l2=scales[1])
-        if name == "l1+l2" or name == "L1L2"
+        if name == "l1+l2" or name == "l1l2"
         else None
     )
--- a/deepxde/nn/regularizers.py
+++ b/deepxde/nn/regularizers.py
@@ -7,9 +7,9 @@ def get(identifier):
         return None
     name, scales = identifier[0], identifier[1:]
     return (
-        tf.keras.regularizers.l1(l=scales[0])
+        tf.keras.regularizers.l1(l1=scales[0])
         if name == "l1"
-        else tf.keras.regularizers.l2(l=scales[0])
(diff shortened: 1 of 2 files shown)


deepxde/nn/regularizers.py
```diff
@@ -7,9 +7,9 @@ def get(identifier):
         return None
     name, scales = identifier[0], identifier[1:]
     return (
-        tf.keras.regularizers.l1(l=scales[0])
+        tf.keras.regularizers.l1(l1=scales[0])
         if name == "l1"
-        else tf.keras.regularizers.l2(l=scales[0])
+        else tf.keras.regularizers.l2(l2=scales[0])
```
lululxvi, line 12: Use `tf.keras.regularizers.L2`

deepxde/nn/regularizers.py
```diff
@@ -7,9 +7,9 @@ def get(identifier):
         return None
     name, scales = identifier[0], identifier[1:]
```
lululxvi, line 8:
```
name, scales = identifier[0].lower(), identifier[1:]
```

deepxde/nn/regularizers.py
```diff
@@ -7,9 +7,9 @@ def get(identifier):
...
     return (
-        tf.keras.regularizers.l1(l=scales[0])
+        tf.keras.regularizers.l1(l1=scales[0])
         if name == "l1"
-        else tf.keras.regularizers.l2(l=scales[0])
+        else tf.keras.regularizers.l2(l2=scales[0])
         if name == "l2"
         else tf.keras.regularizers.l1_l2(l1=scales[0], l2=scales[1])
```
lululxvi, line 14: `tf.keras.regularizers.L1L2`

lululxvi:
Split the regularization and model save into two PRs.

deepxde/nn/regularizers.py
```diff
@@ -7,9 +7,9 @@ def get(identifier):
         return None
     name, scales = identifier[0], identifier[1:]
     return (
-        tf.keras.regularizers.l1(l=scales[0])
+        tf.keras.regularizers.l1(l1=scales[0])
```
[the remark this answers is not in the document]
agniv-the-marker: The old one still works: https://www.tensorflow.org/api_docs/python/tf/keras/regularizers/

The lowercase is just an alias.

PR #1782, opened 2024-06-20
[Fix] Update model saving for TF backend

agniv-the-marker:
# TF save_weights

Currently the saving for backend saves to the wrong filetype, according to the [docs](https://www.tensorflow.org/api_docs/python/tf/keras/Model#save_weights) it must end with a `.weights.h5` suffix, currently ends in `.ckpt` suffix which breaks if you are using the TF backend.

### DIFF
--- a/deepxde/model.py
+++ b/deepxde/model.py
@@ -1028,31 +1028,39 @@ class Model:
                 If `protocol` is "pickle", save using the Python pickle module. Only the
                 protocol "backend" supports ``restore()``.
 
+        Raises:
+            ValueError: If the provided `save_path` is empty.
+
         Returns:
             string: Path where model is saved.
         """
-        # TODO: backend tensorflow
+        if not save_path:
+            raise ValueError("The save_path cannot be empty.")
         save_path = f"{save_path}-{self.train_state.epoch}"
         if protocol == "pickle":
             save_path += ".pkl"
             with open(save_path, "wb") as f:
                 pickle.dump(self.state_dict(), f)
         elif protocol == "backend":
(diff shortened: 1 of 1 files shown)


deepxde/model.py
```diff
@@ -1028,31 +1028,39 @@ def save(self, save_path, protocol="backend", verbose=0):
...
                 protocol "backend" supports ``restore()``.
 
+        Raises:
+            ValueError: If the provided `save_path` is empty.
+
         Returns:
             string: Path where model is saved.
         """
-        # TODO: backend tensorflow
+        if not save_path:
```
lululxvi, line 1037: `save_path` is not optional, so this `if` is rarely to be True, and not necessary.

deepxde/model.py
```diff
@@ -1028,31 +1028,39 @@ def save(self, save_path, protocol="backend", verbose=0):
...
+            raise ValueError("The save_path cannot be empty.")
         save_path = f"{save_path}-{self.train_state.epoch}"
         if protocol == "pickle":
             save_path += ".pkl"
             with open(save_path, "wb") as f:
                 pickle.dump(self.state_dict(), f)
         elif protocol == "backend":
+            # Handle file endings
```
lululxvi, line 1045: Use the original way to handle ending, where different backend codes are independent as much as possible.

deepxde/model.py
agniv-the-marker: Removed

deepxde/model.py
agniv-the-marker: Reverted

PR #1783, opened 2024-06-20
[Update] Update regularizers so it is backwards compatabile

agniv-the-marker:
Regularizers supported in tf v2.2.0:

image

Keeping it like this (lowercase and without explicitly calling the argument name) maintains backwards compatibility and doesn't need to be upgraded to tf v2.3.0 and up.


agniv-the-marker:
Note #1784 renders this PR not useful, although this one has more backwards compatibility with TF 2.2.0 versions.

PR #1811, opened 2024-07-31
Add uniform_boundary_points to Hypercube

vl-dud:
The method generates a specified number of points uniformly distributed along the boundary of an n-dimensional hypercube

### DIFF
--- a/deepxde/geometry/geometry_nd.py
+++ b/deepxde/geometry/geometry_nd.py
@@ -81,6 +81,37 @@ class Hypercube(Geometry):
             )
         return x
 
+    def uniform_boundary_points(self, n):
+        points_per_face = max(1, n // (2 * self.dim))
+        points = []
+        for d in range(self.dim):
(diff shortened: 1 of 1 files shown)


deepxde/geometry/geometry_nd.py
```diff
@@ -81,6 +81,37 @@ def uniform_points(self, n, boundary=True):
...
+                        xi.append(
+                            np.linspace(
+                                self.xmin[i],
+                                self.xmax[i],
+                                num=ni + 1,
+                                endpoint=False,
+                                dtype=config.real(np),
+                            )[1:]
+                        )
+                face_points = np.array(list(itertools.product(*xi)))
```
lululxvi, line 104: Would the points in the corners or edges be computed more than once?

vl-dud: No, they shouldn't

PR #1849, opened 2024-10-02
Tensorflow 1.x backend: branch subnet refactoring for DeepONet

vl-dud:
Made the `build_branch_net` method more readable. At the same time, I added a regularization for the stacked case.

### DIFF
--- a/deepxde/nn/tensorflow_compat_v1/deeponet.py
+++ b/deepxde/nn/tensorflow_compat_v1/deeponet.py
@@ -321,54 +321,48 @@ class DeepONet(NN):
         y_func = self.X_func
         if callable(self.layer_size_func[1]):
             # User-defined network
-            y_func = self.layer_size_func[1](y_func)
-        elif self.stacked:
-            # Stacked fully connected network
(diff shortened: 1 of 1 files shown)


deepxde/nn/tensorflow_compat_v1/deeponet.py
```diff
@@ -321,54 +321,48 @@ def build_branch_net(self):
...
-        elif self.stacked:
-            # Stacked fully connected network
-            stack_size = self.layer_size_func[-1]
-            for i in range(1, len(self.layer_size_func) - 1):
-                y_func = self._stacked_dense(
-                    y_func,
-                    self.layer_size_func[i],
-                    stack_size,
-                    activation=self.activation_branch,
+            return self.layer_size_func[1](y_func)
```
lululxvi, line 324: This `return` is good. But it is better to clearly separate code of stacked and unstacked, as the stacked code is not useful for 99% cases.

vl-dud: Separated

PR #1859, opened 2024-10-21
Backend Tensorflow 1.x: replace tf.layers.dense to tf.keras.layers.Dense

vl-dud:
Switch to the modern TF style, which avoids such warnings: 
```
UserWarning: `tf.layers.dense` is deprecated and will be removed in a future version. Please use `tf.keras.layers.Dense` instead.
```

### DIFF
--- a/deepxde/nn/tensorflow_compat_v1/deeponet.py
+++ b/deepxde/nn/tensorflow_compat_v1/deeponet.py
@@ -422,14 +422,18 @@ class DeepONet(NN):
         regularizer=None,
         trainable=True,
     ):
-        return tf.keras.layers.Dense(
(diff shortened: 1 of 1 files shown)


lululxvi:
Please see below for the regularizaion:
https://github.com/lululxvi/deepxde/blob/ba8e8240ed7ab4ed9d30316409a9f20713027bbc/deepxde/nn/tensorflow_compat_v1/nn.py#L14

https://github.com/lululxvi/deepxde/blob/ba8e8240ed7ab4ed9d30316409a9f20713027bbc/deepxde/nn/tensorflow_compat_v1/fnn.py#L118

vl-dud:
You are right, the loss of regularization was ignored. I have corrected it.

PR #1864, opened 2024-10-27
Modify Tensorflow regularizers

vl-dud:
This pull request introduces several enhancements to TF regularizers, including:
- refactoring
- added the option to call regularizers by name (without the need for scalers)
- support of OrthogonalRegularizer

### DIFF
--- a/deepxde/nn/regularizers.py
+++ b/deepxde/nn/regularizers.py
@@ -1,14 +1,5 @@
 from ..backend import tf
 
-REGULARIZER_DICT = {
-    "l1": tf.keras.regularizers.L1,
-    "l2": tf.keras.regularizers.L2,
-    "l1l2": tf.keras.regularizers.L1L2,
-    "l1+l2": tf.keras.regularizers.L1L2,
-}
-if hasattr(tf.keras.regularizers, "OrthogonalRegularizer"):
-    REGULARIZER_DICT["orthogonal"] = tf.keras.regularizers.OrthogonalRegularizer
-
 
 def get(identifier):
     """Retrieves a TensorFlow regularizer instance based on the given identifier.
@@ -24,35 +15,27 @@ def get(identifier):
     # TODO: other backends
     if identifier is None or not identifier:
         return None
+    if not isinstance(identifier, (list, tuple)):
+        raise ValueError("Identifier must be a list or a tuple.")
 
-    if isinstance(identifier, (list, tuple)):
-        name = identifier[0].lower()
-        factor = identifier[1:]
-    else:
-        raise ValueError("Identifier must be a non-empty list or tuple.")
-
+    name = identifier[0].lower()
+    factor = identifier[1:]
     if not factor:
         raise ValueError("Regularization factor must be provided.")
 
-    regularizer_class = REGULARIZER_DICT.get(name)
-    if not regularizer_class:
-        if name == "orthogonal":
+    if name == "l1":
+        return tf.keras.regularizers.L1(l1=factor[0])
+    if name == "l2":
+        return tf.keras.regularizers.L2(l2=factor[0])
+    if name == "orthogonal":
+        if not hasattr(tf.keras.regularizers, "OrthogonalRegularizer"):
             raise ValueError(
                 "The 'orthogonal' regularizer is not available "
                 "in your version of TensorFlow"
             )
-        raise ValueError(f"Unknown regularizer name: {name}")
-
-    regularizer_kwargs = {}
-    if name == "l1":
-        regularizer_kwargs["l1"] = factor[0]
-    elif name == "l2":
-        regularizer_kwargs["l2"] = factor[0]
-    elif name == "orthogonal":
-        regularizer_kwargs["factor"] = factor[0]
-    elif name in ("l1l2", "l1+l2"):
+        return tf.keras.regularizers.OrthogonalRegularizer(factor=factor[0])
+    if name in ("l1l2", "l1+l2"):
         if len(factor) < 2:
             raise ValueError("L1L2 regularizer requires both L1/L2 penalties.")
-        regularizer_kwargs["l1"] = factor[0]
-        regularizer_kwargs["l2"] = factor[1]
-    return regularizer_class(**regularizer_kwargs)
+        return tf.keras.regularizers.L1L2(l1=factor[0], l2=factor[1])
+    raise ValueError(f"Unknown regularizer name: {name}")
--- a/deepxde/nn/regularizers.py
+++ b/deepxde/nn/regularizers.py
@@ -22,10 +22,10 @@ def get(identifier):
     """
 
     # TODO: other backends
-    if identifier is None:
+    if identifier is None or not identifier:
         return None
 
-    if isinstance(identifier, (list, tuple)) and identifier:
+    if isinstance(identifier, (list, tuple)):
         name = identifier[0].lower()
         factor = identifier[1:]
     else:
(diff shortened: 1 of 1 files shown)


deepxde/nn/regularizers.py
```diff
@@ -1,17 +1,61 @@
...
+
 
 def get(identifier):
+    """Retrieves a TensorFlow regularizer instance based on the given identifier.
+
+    Args:
+        identifier (str or list/tuple): Specifies the type of regularizer and
+            the optional scale values. If a string, it should be one of "l1",
+            "l2", "orthogonal", or "l1l2" ("l1+l2"); default scale values will
+            be used. If a list or tuple, the first element should be one of
```
lululxvi, line 20: It doesn't seem a good idea to support default scale values, as the value is very important and there is no default value agreed in the community.

vl-dud: TF has default scale values for all regularizers. However, I agree with you. Fixed.

deepxde/nn/regularizers.py
```diff
@@ -1,17 +1,58 @@
...
-        tf.keras.regularizers.L1(l1=scales[0])
-        if name == "l1"
-        else tf.keras.regularizers.L2(l2=scales[0])
-        if name == "l2"
-        else tf.keras.regularizers.L1L2(l1=scales[0], l2=scales[1])
-        if name in ("l1+l2", "l1l2")
-        else None
-    )
+
+    if isinstance(identifier, (list, tuple)) and identifier:
```
lululxvi, line 28: Why `and identifier` is needed?

It is better to change line 25 to
```
if identifier is None or not identifier:
```

vl-dud: Fixed

deepxde/nn/regularizers.py
```diff
@@ -1,17 +1,58 @@
...
-        else tf.keras.regularizers.L1L2(l1=scales[0], l2=scales[1])
-        if name in ("l1+l2", "l1l2")
-        else None
-    )
+
+    if isinstance(identifier, (list, tuple)):
+        name = identifier[0].lower()
+        factor = identifier[1:]
+    else:
+        raise ValueError("Identifier must be a non-empty list or tuple.")
```
lululxvi, line 32: This error message doesn't seem useful anymore, after revising line 25.

deepxde/nn/regularizers.py
```diff
@@ -1,17 +1,58 @@
...
+    if not regularizer_class:
+        if name == "orthogonal":
+            raise ValueError(
+                "The 'orthogonal' regularizer is not available "
+                "in your version of TensorFlow"
+            )
+        raise ValueError(f"Unknown regularizer name: {name}")
+
+    regularizer_kwargs = {}
+    if name == "l1":
```
lululxvi, line 47: I prefer the following:
```
if name == "l1":
    return tf.keras.regularizers.L1(l1=factor[0])
if name == "l2":
    return tf.keras.regularizers.L2(l2=factor[0])
...
```
This is easier to read and simpler.

deepxde/nn/regularizers.py
vl-dud: I hope I understood your wish correctly. 
I left the following check:

```
if not isinstance(identifier, (list, tuple)):
    raise ValueError("Identifier must be a list or a tuple.")
```

in case user tries to pass a string like `regularization="l1"`

deepxde/nn/regularizers.py
vl-dud: Fixed

deepxde/nn/regularizers.py
```diff
@@ -2,16 +2,40 @@
...
+    if name == "l2":
+        return tf.keras.regularizers.L2(l2=factor[0])
+    if name == "orthogonal":
+        if not hasattr(tf.keras.regularizers, "OrthogonalRegularizer"):
+            raise ValueError(
+                "The 'orthogonal' regularizer is not available "
+                "in your version of TensorFlow"
+            )
+        return tf.keras.regularizers.OrthogonalRegularizer(factor=factor[0])
+    if name in ("l1l2", "l1+l2"):
```
lululxvi, line 37: swap the order of "l1l2" and "orthogonal"

vl-dud: Fixed

deepxde/nn/regularizers.py
```diff
@@ -2,16 +2,40 @@
 
 
 def get(identifier):
+    """Retrieves a TensorFlow regularizer instance based on the given identifier.
+
+    Args:
+        identifier (list/tuple): Specifies the type of regularizer and
+            regularization factor. The first element should be one of "l1", "l2",
+            "orthogonal", or "l1l2" ("l1+l2"). For "l1", "l2", or "orthogonal",
```
lululxvi, line 10: move l1l2 before orthogonal

vl-dud: Fixed

deepxde/nn/regularizers.py
```diff
@@ -2,16 +2,40 @@
...
+        if len(factor) < 2:
+            raise ValueError("L1L2 regularizer requires both L1/L2 penalties.")
+        return tf.keras.regularizers.L1L2(l1=factor[0], l2=factor[1])
+    if name == "orthogonal":
+        if not hasattr(tf.keras.regularizers, "OrthogonalRegularizer"):
+            raise ValueError(
+                "The 'orthogonal' regularizer is not available "
+                "in your version of TensorFlow"
+            )
+        return tf.keras.regularizers.OrthogonalRegularizer(factor=factor[0])
```
lululxvi, line 40: What is the first version of TF with OrthogonalRegularizer?

deepxde/nn/regularizers.py
```diff
@@ -2,16 +2,40 @@
...
+    if name == "l1":
+        return tf.keras.regularizers.L1(l1=factor[0])
+    if name == "l2":
+        return tf.keras.regularizers.L2(l2=factor[0])
+    if name in ("l1l2", "l1+l2"):
+        if len(factor) < 2:
+            raise ValueError("L1L2 regularizer requires both L1/L2 penalties.")
+        return tf.keras.regularizers.L1L2(l1=factor[0], l2=factor[1])
+    if name == "orthogonal":
+        if not hasattr(tf.keras.regularizers, "OrthogonalRegularizer"):
```
lululxvi, line 35: Delete this. But add the requirement of tf >= 2.9.0 in
- README
- https://github.com/lululxvi/deepxde/blob/master/docs/requirements.txt
- https://github.com/lululxvi/deepxde/blob/master/docs/user/installation.rst
- https://github.com/lululxvi/deepxde/blob/master/deepxde/backend/tensorflow/tensor.py

vl-dud: OrthogonalRegularizer is not available in TF 1.x. So some checking will still be needed.

lululxvi: Do you find OrthogonalRegularizer useful?

lululxvi: Let us not include OrthogonalRegularizer at the moment for simplicity. If needed, can add it later.

vl-dud: OK

deepxde/nn/regularizers.py
```diff
@@ -2,16 +2,33 @@
 
 
 def get(identifier):
+    """Retrieves a TensorFlow regularizer instance based on the given identifier.
+
+    Args:
+        identifier (list/tuple): Specifies the type of regularizer and
+            regularization factor. The first element should be one of "l1", "l2",
+            or "l1l2" ("l1+l2"). For "l1" and "l2",
+            you can provide a single factor value. For "l1l2" ("l1+l2"),
```
lululxvi, line 11: Refactor the comments here.

vl-dud: Check it now, please

PR #1884, opened 2024-11-13
Add l1 and l1+l2 regularizers for PyTorch backend

vl-dud:
L1 regularizer is added. Not quite sure about the line:
```
l1_loss = torch.sum(torch.stack([torch.sum(p.abs()) for p in trainable_variables]))
```
Perhaps regularization should not be applied to `self.external_trainable_variables`? Then it must be like:
```
l1_loss = torch.sum(torch.stack([torch.sum(p.abs()) for p in self.net.parameters()]))
```
However, in current l2 regularization `trainable_variables` (both `self.net.parameters()` and `self.external_trainable_variables`) are used.

### DIFF
--- a/deepxde/model.py
+++ b/deepxde/model.py
@@ -339,24 +339,22 @@ class Model:
                 )
             weight_decay = self.net.regularizer[1]
 
-        if weight_decay == 0:
-            if self.opt_name == "adamw":
-                raise ValueError("AdamW optimizer requires non-zero weight decay")
+        if self.opt_name in ["L-BFGS", "L-BFGS-B"]:
             optimizer_params = (
                 list(self.net.parameters()) + self.external_trainable_variables
             )
         else:
-            if self.opt_name in ["L-BFGS", "L-BFGS-B"]:
-                raise ValueError(
-                    f"{self.opt_name} optimizer doesn't support weight_decay > 0"
-                )
             optimizer_params = [
-                {"params": self.net.parameters(), "weight_decay": weight_decay},
+                {"params": self.net.parameters()},
                 {"params": self.external_trainable_variables, "weight_decay": 0},
             ]
 
         self.opt, self.lr_scheduler = optimizers.get(
-            optimizer_params, self.opt_name, learning_rate=lr, decay=decay
+            optimizer_params,
+            self.opt_name,
+            learning_rate=lr,
+            decay=decay,
+            weight_decay=weight_decay,
         )
 
         def train_step(inputs, targets, auxiliary_vars):
--- a/deepxde/model.py
+++ b/deepxde/model.py
@@ -342,6 +342,7 @@ class Model:
 
         optimizer_params = self.net.parameters()
         if self.external_trainable_variables:
+            # L-BFGS doesn't support per-parameter options.
             if self.opt_name in ["L-BFGS", "L-BFGS-B"]:
                 optimizer_params = (
                     list(optimizer_params) + self.external_trainable_variables
--- a/deepxde/model.py
+++ b/deepxde/model.py
@@ -340,20 +340,22 @@ class Model:
                 )
             weight_decay = self.net.regularizer[1]
 
-        if self.opt_name in ["L-BFGS", "L-BFGS-B"]:
-            optimizer_params = (
-                list(self.net.parameters()) + self.external_trainable_variables
-            )
-            if weight_decay > 0:
-                print(
-                    "Warning: L2 regularization will also be applied to external_trainable_variables. "
-                    "Ensure this is intended behavior."
+        optimizer_params = self.net.parameters()
+        if self.external_trainable_variables:
+            if self.opt_name in ["L-BFGS", "L-BFGS-B"]:
+                optimizer_params = (
+                    list(optimizer_params) + self.external_trainable_variables
                 )
-        else:
-            optimizer_params = [
-                {"params": self.net.parameters()},
-                {"params": self.external_trainable_variables, "weight_decay": 0},
-            ]
+                if weight_decay > 0:
+                    print(
+                        "Warning: L2 regularization will also be applied to external_trainable_variables. "
+                        "Ensure this is intended behavior."
+                    )
+            else:
+                optimizer_params = [
+                    {"params": optimizer_params},
+                    {"params": self.external_trainable_variables, "weight_decay": 0},
+                ]
 
         self.opt, self.lr_scheduler = optimizers.get(
             optimizer_params,
--- a/deepxde/model.py
+++ b/deepxde/model.py
@@ -330,37 +330,32 @@ class Model:
                 False, inputs, targets, auxiliary_vars, self.data.losses_test
             )
 
-        # Another way is using per-parameter options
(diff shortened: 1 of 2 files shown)


lululxvi:
You are right that regularization should not be applied to `self.external_trainable_variables`. Let us fix this first in this PR.

vl-dud:
Fixed. Please note that I replaced
```
raise ValueError("L-BFGS optimizer doesn't support weight_decay > 0")
```
with
```
print(f"Warning: weight decay is ignored for {self.opt_name}")
```

And I removed the exception
```
raise ValueError("AdamW optimizer requires non-zero weight decay")
```
because AdamW will still start with the default `weight_decay`

lululxvi:
Let us fix l2 in this PR. Use a new PR for adding l1 and l1+l2.

vl-dud:
Take a look now please.

deepxde/model.py
```diff
@@ -330,30 +330,33 @@ def outputs_losses_test(inputs, targets, auxiliary_vars):
...
             if self.net.regularizer[0] == "l2":
-                self.opt, self.lr_scheduler = optimizers.get(
-                    trainable_variables,
-                    self.opt_name,
-                    learning_rate=lr,
-                    decay=decay,
-                    weight_decay=self.net.regularizer[1],
-                )
+                if self.opt_name in ["L-BFGS", "L-BFGS-B"]:
+                    print(f"Warning: weight decay is ignored for {self.opt_name}")
```
lululxvi, line 340: raise ValueError, not only warning

deepxde/optimizers/pytorch/optimizers.py
```diff
@@ -33,21 +31,13 @@ def get(params, optimizer, learning_rate=None, decay=None, weight_decay=0):
...
-                params, lr=learning_rate, weight_decay=weight_decay
-            )
+            optim = torch.optim.RMSprop(params, lr=learning_rate)
         elif optimizer == "adam":
-            optim = torch.optim.Adam(
-                params, lr=learning_rate, weight_decay=weight_decay
-            )
+            optim = torch.optim.Adam(params, lr=learning_rate)
         elif optimizer == "adamw":
-            if weight_decay == 0:
```
lululxvi, line 46: We need to keep this error to avoid any unexcepted effect.

deepxde/model.py
vl-dud: Fixed

deepxde/optimizers/pytorch/optimizers.py
vl-dud: Fixed

deepxde/optimizers/pytorch/optimizers.py
```diff
@@ -9,14 +9,12 @@ def is_external_optimizer(optimizer):
     return optimizer in ["L-BFGS", "L-BFGS-B"]
 
 
-def get(params, optimizer, learning_rate=None, decay=None, weight_decay=0):
```
lululxvi, line 12: Keep this `weight_decay=0`. Then use the following in model.py
```
optimizer_params = [
                {"params": self.net.parameters()},
                {"params": self.external_trainable_variables, "weight_decay": 0},
            ]
``

Comment (vl-dud) [reply deepxde/optimizers/pytorch/optimizers.py]: Fixed.

**Comment by lululxvi** [comment]
In line 113, emphasize that `external_trainable_variables` will not apply regularization.

File: deepxde/model.py [hunk 4]
```diff
@@ -330,30 +330,32 @@ def outputs_losses_test(inputs, targets, auxiliary_vars):
...
                     f"{self.net.regularizer[0]} regularization to be implemented for "
-                    "backend pytorch."
+                    "backend pytorch"
                 )
+            weight_decay = self.net.regularizer[1]
+
+        if self.opt_name in ["L-BFGS", "L-BFGS-B"]:
+            optimizer_params = (
+                list(self.net.parameters()) + self.external_trainable_variables
+            )
```
lululxvi, line 345: Add a warning here that external_trainable_variables will also happy regularization if weight decay > 0.

vl-dud: If weight decay > 0, an exception is thrown in *optimizers.py*:
```
elif optimizer in ["L-BFGS", "L-BFGS-B"]:
	if weight_decay > 0:
		raise ValueError("L-BFGS optimizer doesn't support weight_decay > 0")
```
The warning doesn't seem to make much sense, does it?

vl-dud:
> In line 113, emphasize that `external_trainable_variables` will not apply regularization.

Fixed.

deepxde/model.py
```diff
@@ -112,9 +112,11 @@ def compile(
...
             external_trainable_variables: A trainable ``dde.Variable`` object or a list
                 of trainable ``dde.Variable`` objects. The unknown parameters in the
-                physics systems that need to be recovered. If the backend is
-                tensorflow.compat.v1, `external_trainable_variables` is ignored, and all
-                trainable ``dde.Variable`` objects are automatically collected.
+                physics systems that need to be recovered. Regularization will not be
+                applied to these variables. If the backend is tensorflow.compat.v1,
+                `external_trainable_variables` is ignored, and all trainable ``dde.Variable``
+                objects are automatically collected.
+
```
lululxvi, line 119: Delete this empty line.

deepxde/model.py
lululxvi: I still suggest to add this warning, as in the future, L-BFGS may support weight decay > 0.

deepxde/model.py
vl-dud: Fixed.

deepxde/model.py
```diff
@@ -330,30 +331,37 @@ def outputs_losses_test(inputs, targets, auxiliary_vars):
...
+            )
+            if weight_decay > 0:
+                print(
+                    "Warning: L2 regularization will also be applied to external_trainable_variables. "
+                    "Ensure this is intended behavior."
+                )
+        else:
+            optimizer_params = [
+                {"params": self.net.parameters()},
+                {"params": self.external_trainable_variables, "weight_decay": 0},
```
lululxvi, line 355: What if external_trainable_variables = None?

vl-dud: Although the PyTorch optimizer handles empty values correctly, you're right — it’s better to make this point more explicit. Fixed.

deepxde/model.py
```diff
@@ -330,30 +331,39 @@ def outputs_losses_test(inputs, targets, auxiliary_vars):
...
                     f"{self.net.regularizer[0]} regularization to be implemented for "
-                    "backend pytorch."
+                    "backend pytorch"
                 )
+            weight_decay = self.net.regularizer[1]
+
+        optimizer_params = self.net.parameters()
+        if self.external_trainable_variables:
+            if self.opt_name in ["L-BFGS", "L-BFGS-B"]:
+                optimizer_params = (
```
lululxvi, line 346: Add a comment:
```
L-BFGS doesn't support per-parameter options.
```

vl-dud: Fixed

PR #1895, opened 2024-11-28
Backend paddle: add mfnn and modify fnn

lijialin03:
For paddle backend:

1. add mfnn net
2. add regularization for fnn net

PR #1896, opened 2024-11-28
Backend paddle: add optimizers with supporting regularizer

lijialin03:
add sgd/rmsprop/adamw with supporting regularizer for paddle backend

### DIFF
--- a/deepxde/model.py
+++ b/deepxde/model.py
@@ -517,13 +517,12 @@ class Model:
         trainable_variables = (
             list(self.net.parameters()) + self.external_trainable_variables
         )
-        weight_decay = getattr(self.net, "regularizer", None)
         self.opt = optimizers.get(
             trainable_variables,
             self.opt_name,
             learning_rate=lr,
             decay=decay,
-            weight_decay=weight_decay,
+            weight_decay=self.net.regularizer,
         )
 
         def train_step(inputs, targets, auxiliary_vars):
--- a/deepxde/model.py
+++ b/deepxde/model.py
@@ -517,8 +517,13 @@ class Model:
         trainable_variables = (
             list(self.net.parameters()) + self.external_trainable_variables
         )
+        regularizer = getattr(self.net, 'regularizer', None)
+        if regularizer is not None:
+            weight_decay = self.net.regularizer_value if self.opt_name == "adamw" else self.net.regularizer
+        else: 
+            weight_decay = None
         self.opt = optimizers.get(
-            trainable_variables, self.opt_name, learning_rate=lr, decay=decay
+            trainable_variables, self.opt_name, learning_rate=lr, decay=decay, weight_decay=weight_decay,
         )
 
         def train_step(inputs, targets, auxiliary_vars):
--- a/deepxde/model.py
+++ b/deepxde/model.py
@@ -517,13 +517,21 @@ class Model:
         trainable_variables = (
             list(self.net.parameters()) + self.external_trainable_variables
         )
-        regularizer = getattr(self.net, 'regularizer', None)
+        regularizer = getattr(self.net, "regularizer", None)
         if regularizer is not None:
-            weight_decay = self.net.regularizer_value if self.opt_name == "adamw" else self.net.regularizer
-        else: 
+            weight_decay = (
+                self.net.regularizer_value
+                if self.opt_name == "adamw"
+                else self.net.regularizer
+            )
+        else:
             weight_decay = None
         self.opt = optimizers.get(
-            trainable_variables, self.opt_name, learning_rate=lr, decay=decay, weight_decay=weight_decay,
+            trainable_variables,
+            self.opt_name,
+            learning_rate=lr,
+            decay=decay,
+            weight_decay=weight_decay,
         )
 
         def train_step(inputs, targets, auxiliary_vars):
--- a/deepxde/model.py
+++ b/deepxde/model.py
@@ -517,15 +517,7 @@ class Model:
         trainable_variables = (
             list(self.net.parameters()) + self.external_trainable_variables
         )
-        regularizer = getattr(self.net, "regularizer", None)
-        if regularizer is not None:
-            weight_decay = (
-                self.net.regularizer_value
-                if self.opt_name == "adamw"
-                else self.net.regularizer
-            )
-        else:
-            weight_decay = None
+        weight_decay = getattr(self.net, "regularizer", None)
         self.opt = optimizers.get(
             trainable_variables,
             self.opt_name,
--- a/deepxde/optimizers/paddle/optimizers.py
+++ b/deepxde/optimizers/paddle/optimizers.py
@@ -66,9 +66,7 @@ def get(params, optimizer, learning_rate=None, decay=None, weight_decay=None):
             not isinstance(weight_decay, paddle.regularizer.L2Decay)
             or weight_decay._coeff == 0
         ):
-            raise ValueError(
-                "AdamW optimizer requires non-zero L2 regularizer"
-            )
+            raise ValueError("AdamW optimizer requires non-zero L2 regularizer")
         return paddle.optimizer.AdamW(
             learning_rate=learning_rate,
             parameters=params,
--- a/deepxde/optimizers/paddle/optimizers.py
+++ b/deepxde/optimizers/paddle/optimizers.py
@@ -62,13 +62,11 @@ def get(params, optimizer, learning_rate=None, decay=None, weight_decay=None):
             weight_decay=weight_decay,
         )
     if optimizer == "adamw":
-        if isinstance(weight_decay, paddle.regularizer.L2Decay):
-            if weight_decay._coeff == 0:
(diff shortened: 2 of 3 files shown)


deepxde/optimizers/paddle/optimizers.py
```diff
@@ -46,5 +48,17 @@ def get(params, optimizer, learning_rate=None, decay=None):
         learning_rate = _get_lr_scheduler(learning_rate, decay)
 
     if optimizer == "adam":
-        return paddle.optimizer.Adam(learning_rate=learning_rate, parameters=params)
+        return paddle.optimizer.Adam(learning_rate=learning_rate, parameters=params, weight_decay=weight_decay)
```
lululxvi, line 51: Refactor using black

deepxde/optimizers/paddle/optimizers.py
```diff
@@ -46,5 +48,25 @@ def get(params, optimizer, learning_rate=None, decay=None):
         learning_rate = _get_lr_scheduler(learning_rate, decay)
 
     if optimizer == "adam":
-        return paddle.optimizer.Adam(learning_rate=learning_rate, parameters=params)
+        return paddle.optimizer.Adam(
+            learning_rate=learning_rate, parameters=params, weight_decay=weight_decay
+        )
+    elif optimizer == "sgd":
```
lululxvi, line 54:
```
if optimizer == "sgd":
```

deepxde/optimizers/paddle/optimizers.py
```diff
@@ -46,5 +48,25 @@ def get(params, optimizer, learning_rate=None, decay=None):
...
+        return paddle.optimizer.SGD(
+            learning_rate=learning_rate, parameters=params, weight_decay=weight_decay
+        )
+    elif optimizer == "rmsprop":
```
lululxvi, line 58:
```
if optimizer == "rmsprop":
```

deepxde/optimizers/paddle/optimizers.py
```diff
@@ -46,5 +48,25 @@ def get(params, optimizer, learning_rate=None, decay=None):
...
+        return paddle.optimizer.RMSProp(
+            learning_rate=learning_rate,
+            parameters=params,
+            weight_decay=weight_decay,
+        )
+    elif optimizer == "adamw":
```
lululxvi, line 64:
```
if optimizer == "adamw":
```

deepxde/optimizers/paddle/optimizers.py
```diff
@@ -46,5 +48,25 @@ def get(params, optimizer, learning_rate=None, decay=None):
...
+            learning_rate=learning_rate, parameters=params, weight_decay=weight_decay
+        )
+    if optimizer == "rmsprop":
+        return paddle.optimizer.RMSProp(
+            learning_rate=learning_rate,
+            parameters=params,
+            weight_decay=weight_decay,
+        )
+    if optimizer == "adamw":
+        if weight_decay[0] == 0:
```
lululxvi, line 65: Why `weight_decay[0]`?

lijialin03: Because `weight_decay` of `paddle.optimizer.AdamW` should be a value Reference [AdamW](https://www.paddlepaddle.org.cn/documentation/docs/zh/2.6/api/paddle/optimizer/AdamW_cn.html#adamw).
So code to get value of regularizer is added in another PR, see line 31 of [PR](https://github.com/lululxvi/deepxde/pull/1895/files#diff-03cc98ff740e84f0a25d25416b8b9fddd193da2485d496fd39cbed8f47da8724).
Now that format of regularizer is `["name",value1,value2,...]` and `AdamW` only need one value, `weight_decay[0]` here actually means `regularizer[1:][0]`, which is `value1`. I think this is a way to avoid errors as much as possible.

lululxvi: This makes things complicated, as you also need to modify model.py.
How about modifying regularizers.py to handle this.

lijialin03: It is not work because optimizer's type can not be got in regularizers.py. Actually regularization is got with network settings, before setting optimizer.
Or there might be a feasible way to handle this situation which is adding a regularizer type in regularizers.py, such as `value`. Is that ok?

lululxvi: For AdamW, it should be L2 regularization. So, we should check if `isinstance(weight_decay,  paddle.regularizer. L2Decay)` and then use `weight_decay._coeff`

lijialin03: That's right. I have updated this PR and https://github.com/lululxvi/deepxde/pull/1895/commits/375e19c1498e25121c312b1f053a168069d9c01a at the same time.

lijialin03: Sorry, that is not work because of a code-style error 'Access to a protected member _coeff of a client class (protected-access)'

deepxde/model.py
```diff
@@ -517,8 +517,13 @@ def outputs_losses_test(inputs, targets, auxiliary_vars):
         trainable_variables = (
             list(self.net.parameters()) + self.external_trainable_variables
         )
+        weight_decay = getattr(self.net, "regularizer", None)
```
lululxvi, line 520: https://github.com/lululxvi/deepxde/blob/master/deepxde/nn/paddle/nn.py

The NN class should add a `regularizer` attribute, like other backends.

deepxde/model.py
```diff
@@ -517,8 +517,13 @@ def outputs_losses_test(inputs, targets, auxiliary_vars):
...
         self.opt = optimizers.get(
-            trainable_variables, self.opt_name, learning_rate=lr, decay=decay
+            trainable_variables,
+            self.opt_name,
+            learning_rate=lr,
+            decay=decay,
+            weight_decay=weight_decay,
```
lululxvi, line 526:
```
weight_decay=self.net.regularizer,
```

deepxde/optimizers/paddle/optimizers.py
```diff
@@ -46,5 +48,25 @@ def get(params, optimizer, learning_rate=None, decay=None):
...
+            learning_rate=learning_rate, parameters=params, weight_decay=weight_decay
+        )
+    if optimizer == "rmsprop":
+        return paddle.optimizer.RMSProp(
+            learning_rate=learning_rate,
+            parameters=params,
+            weight_decay=weight_decay,
+        )
+    if optimizer == "adamw":
+        if not isinstance(weight_decay, paddle.regularizer.L2Decay) or weight_decay._coeff == 0:
```
lululxvi, line 65: Reformat via black

deepxde/optimizers/paddle/optimizers.py
```diff
@@ -19,12 +19,14 @@ def is_external_optimizer(optimizer):
...
+def get(params, optimizer, learning_rate=None, decay=None, weight_decay=None):
     """Retrieves an Optimizer instance."""
     if isinstance(optimizer, paddle.optimizer.Optimizer):
         return optimizer
 
     if optimizer in ["L-BFGS", "L-BFGS-B"]:
+        if weight_decay is not None:
+            raise ValueError("L-BFGS optimizer doesn't support weight_decay")
         if learning_rate is not None or decay is not None:
             print("Warning: learning rate is ignored for {}".format(optimizer))
```
lululxvi, line 31:
```
if learning_rate is not None or decay is not None:
    ...
if weight_decay is not None:
    ...
```

deepxde/optimizers/paddle/optimizers.py
```diff
@@ -46,5 +48,30 @@ def get(params, optimizer, learning_rate=None, decay=None):
...
+            parameters=params,
+            weight_decay=weight_decay,
+        )
+    if optimizer == "adamw":
+        if (
+            not isinstance(weight_decay, paddle.regularizer.L2Decay)
+            or weight_decay._coeff == 0
+        ):
+            raise ValueError(
+                "AdamW optimizer requires L2 regularizer and non-zero weight decay"
```
lululxvi, line 70: "AdamW optimizer requires non-zero L2 regularizer"

lijialin03:
A code style error occurred, "Access to a protected member _coeff of a client class (protected-access)".
It seems that weight_decay._coeff cannot be called directly.（The strange thing is that this error did not appear in the first few submissions）

deepxde/optimizers/paddle/optimizers.py
```diff
@@ -46,5 +48,30 @@ def get(params, optimizer, learning_rate=None, decay=None):
...
+            learning_rate=learning_rate,
+            parameters=params,
+            weight_decay=weight_decay,
+        )
+    if optimizer == "adamw":
+        if (
+            not isinstance(weight_decay, paddle.regularizer.L2Decay)
+            or weight_decay._coeff == 0
+        ):
+            raise ValueError(
```
lululxvi, line 69:
```
raise ValueError("AdamW optimizer requires non-zero L2 regularizer")
```

PR #1897, opened 2024-11-28
Backend paddle: adding the type conversion of mf data to support examples in directory 'function'

lijialin03:
1. support examples in examples/function/
2. comparison of results(tf vs paddle) in https://github.com/lululxvi/deepxde/pull/1893

PR #1923, opened 2024-12-18
Backend paddle: add mfnn net

lijialin03:
1. add mfnn net

### DIFF
--- a/deepxde/nn/paddle/mfnn.py
+++ b/deepxde/nn/paddle/mfnn.py
@@ -26,7 +26,6 @@ class MfNN(NN):
         self.layer_size_hi = layer_sizes_high_fidelity
 
         self.activation = activations.get(activation)
-        self.activation_tanh = activations.get("tanh")
         self.initializer = initializers.get(kernel_initializer)
         self.initializer_zero = initializers.get("zeros")
         self.trainable_lo = trainable_low_fidelity
@@ -35,7 +34,7 @@ class MfNN(NN):
         self.regularizer = regularizers.get(regularization)
 
         # low fidelity
-        self.linears_lo = self.init_dense(self.layer_size_lo, self.trainable_lo)
+        self.linears_lo = self._init_dense(self.layer_size_lo, self.trainable_lo)
 
         # high fidelity
         # linear part
@@ -52,18 +51,18 @@ class MfNN(NN):
         self.layer_size_hi = [
             self.layer_size_lo[0] + self.layer_size_lo[-1]
         ] + self.layer_size_hi
-        self.linears_hi = self.init_dense(self.layer_size_hi, self.trainable_hi)
+        self.linears_hi = self._init_dense(self.layer_size_hi, self.trainable_hi)
         # linear + nonlinear
         if not self.residue:
-            alpha = self.init_alpha(0.0, self.trainable_hi)
+            alpha = self._init_alpha(0.0, self.trainable_hi)
             self.add_parameter("alpha", alpha)
         else:
-            alpha1 = self.init_alpha(0.0, self.trainable_hi)
-            alpha2 = self.init_alpha(0.0, self.trainable_hi)
+            alpha1 = self._init_alpha(0.0, self.trainable_hi)
+            alpha2 = self._init_alpha(0.0, self.trainable_hi)
             self.add_parameter("alpha1", alpha1)
             self.add_parameter("alpha2", alpha2)
 
-    def init_dense(self, layer_size, trainable):
+    def _init_dense(self, layer_size, trainable):
         linears = paddle.nn.LayerList()
         for i in range(len(layer_size) - 1):
             linear = paddle.nn.Linear(
@@ -78,7 +77,7 @@ class MfNN(NN):
             linears.append(linear)
         return linears
 
-    def init_alpha(self, value, trainable):
+    def _init_alpha(self, value, trainable):
         alpha = paddle.create_parameter(
             shape=[1],
             dtype=config.real(paddle),
@@ -88,7 +87,8 @@ class MfNN(NN):
         return alpha
 
     def forward(self, inputs):
-        x = inputs.astype(config.real(paddle))
+        x = inputs
+        
         # low fidelity
         y = x
         for i, linear in enumerate(self.linears_lo):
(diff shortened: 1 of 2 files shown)


lijialin03:
Excuse me, Is any problem in this PR? Could it be merged now?

lijialin03:
> Excuse me, Is any problem in this PR? Could it be merged now?

remind +1

deepxde/nn/paddle/mfnn.py
```diff
@@ -0,0 +1,120 @@
...
+        residue=False,
+        trainable_low_fidelity=True,
+        trainable_high_fidelity=True,
+    ):
+        super().__init__()
+        self.layer_size_lo = layer_sizes_low_fidelity
+        self.layer_size_hi = layer_sizes_high_fidelity
+
+        self.activation = activations.get(activation)
+        self.activation_tanh = activations.get("tanh")
```
lululxvi, line 29: This is only for paddle, so we can directly use `paddle.tanh`

deepxde/nn/paddle/mfnn.py
```diff
@@ -0,0 +1,120 @@
...
+        if not self.residue:
+            alpha = self.init_alpha(0.0, self.trainable_hi)
+            self.add_parameter("alpha", alpha)
+        else:
+            alpha1 = self.init_alpha(0.0, self.trainable_hi)
+            alpha2 = self.init_alpha(0.0, self.trainable_hi)
+            self.add_parameter("alpha1", alpha1)
+            self.add_parameter("alpha2", alpha2)
+
+    def init_dense(self, layer_size, trainable):
```
lululxvi, line 66: `def _init_dense`

deepxde/nn/paddle/mfnn.py
```diff
@@ -0,0 +1,120 @@
...
+                weight_attr=paddle.ParamAttr(initializer=self.initializer),
+                bias_attr=paddle.ParamAttr(initializer=self.initializer_zero),
+            )
+            if not trainable:
+                for param in linear.parameters():
+                    param.stop_gradient = False
+            linears.append(linear)
+        return linears
+
+    def init_alpha(self, value, trainable):
```
lululxvi, line 81: `def _init_alpha`

deepxde/nn/paddle/mfnn.py
```diff
@@ -0,0 +1,120 @@
...
+        alpha = paddle.create_parameter(
+            shape=[1],
+            dtype=config.real(paddle),
+            default_initializer=paddle.nn.initializer.Constant(value),
+        )
+        alpha.stop_gradient = not trainable
+        return alpha
+
+    def forward(self, inputs):
+        x = inputs.astype(config.real(paddle))
```
lululxvi, line 91: `x = inputs.astype(config.real(paddle))` doesn't seem necessary.

deepxde/nn/paddle/mfnn.py
lijialin03: It is Okay. But the reason for that is I noticed the line 95~98 of tensorflow_compat_v1/mfnn.py
image
Maybe It is better to delate the 'self.activation_tanh' and just use 'activations.get' like the above code?

deepxde/nn/paddle/mfnn.py
lijialin03: This error occurs when running example/function/mf_dataset.py. Because data dtype of this example is float64 without conversion.
```sh
ValueError: (InvalidArgument) The type of data we are trying to retrieve (float32) does not match the type of data (float64) currently contained in the container.
  [Hint: Expected dtype() == phi::CppTypeToDataType<T>::Type(), but received dtype():11 != phi::CppTypeToDataType<T>::Type():10.] (at ../paddle/phi/core/dense_tensor.cc:160)
  [operator < linear > error]
```
This example only be supported with 'tensorflow.compat.v1' now. Dtype could be setted when creating placeholder with code below. 
image
This is the reason this line exists here.

deepxde/nn/paddle/mfnn.py
lululxvi: Just use `paddle.tanh` is good enough.

deepxde/nn/paddle/mfnn.py
lululxvi: Then we should modify the code https://github.com/lululxvi/deepxde/blob/master/deepxde/data/mf.py by adding the type conversion, as https://github.com/lululxvi/deepxde/blob/master/deepxde/data/dataset.py

lijialin03: It is added in https://github.com/lululxvi/deepxde/pull/1897 now and `x = inputs.astype(config.real(paddle))` has been modified in this PR at the same time.

deepxde/nn/paddle/mfnn.py
```diff
@@ -0,0 +1,120 @@
...
+        # low fidelity
+        self.linears_lo = self._init_dense(self.layer_size_lo, self.trainable_lo)
+
+        # high fidelity
+        # linear part
+        self.linears_hi_l = paddle.nn.Linear(
+            in_features=self.layer_size_lo[0] + self.layer_size_lo[-1],
+            out_features=self.layer_size_hi[-1],
+            weight_attr=paddle.ParamAttr(initializer=self.initializer),
+            bias_attr=paddle.ParamAttr(initializer=self.initializer_zero),
```
lululxvi, line 45: `bias_attr=paddle.ParamAttr(initializer=self.initializer_zero),` doesn't seem necessary, as by default, bias will be initialized as 0.
https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/nn/Linear_en.html

deepxde/nn/paddle/mfnn.py
```diff
@@ -0,0 +1,120 @@
...
+        alpha = paddle.create_parameter(
+            shape=[1],
+            dtype=config.real(paddle),
+            default_initializer=paddle.nn.initializer.Constant(value),
+        )
+        alpha.stop_gradient = not trainable
+        return alpha
+
+    def forward(self, inputs):
+        x = inputs
```
lululxvi, line 90: No need to define this x.

PR #1925, opened 2024-12-24
Fix a bug of forward-mode AD when multi-output is needed

Jerry-Jzy:
I find the bug from this case: https://github.com/lululxvi/deepxde/blob/master/examples/operator/stokes_aligned_pideeponet.py

when setting the `num_output=3` and `multi_output_strategy="independent"`, the output shape will be (batch size, # of coordinates, 3).
Then, the slice here is wrong, what we really want is slicing in the last dim.


deepxde/gradients/gradients_forward.py
```diff
@@ -94,7 +94,7 @@ def grad_fn(x):
                 # which is also a tuple. This is useful for further computation, e.g.,
                 # Hessian.
                 self.J[i, j] = (
-                    self.J[j][0][:, i : i + 1],
+                    self.J[j][0][..., i : i + 1],
```
lululxvi, line 97: Add a comment that the code still works even the output dim is > 2.

Jerry-Jzy: Ok

deepxde/gradients/gradients_forward.py
```diff
@@ -92,9 +92,9 @@ def grad_fn(x):
                 # In backend tensorflow/pytorch/jax, a tuple of a tensor/tensor/array
                 # and a callable is returned, so that it is consistent with the argument,
                 # which is also a tuple. This is useful for further computation, e.g.,
-                # Hessian.
+                # Hessian. The code still works even the output dim is > 2.
```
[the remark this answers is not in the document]
Jerry-Jzy: Sure, the comment is updated

deepxde/gradients/gradients_forward.py
```diff
@@ -92,9 +92,11 @@ def grad_fn(x):
                 # In backend tensorflow/pytorch/jax, a tuple of a tensor/tensor/array
                 # and a callable is returned, so that it is consistent with the argument,
                 # which is also a tuple. This is useful for further computation, e.g.,
-                # Hessian.
+                # Hessian. The code is designed for the output shape of (batch size, dim),
```
lululxvi, line 95: The code --> The original code self.J[j][0][:, i : i + 1]

lululxvi:
Here, we compute the dim 
https://github.com/lululxvi/deepxde/blob/b0d239b1419bf407fe1c23468a470b5f93b65c4e/deepxde/gradients/jacobian.py#L31
Would this a problem?

Jerry-Jzy:
> problem

deepxde/gradients/gradients_forward.py
Jerry-Jzy: ok

Jerry-Jzy:
> Here, we compute the dim
> 
> https://github.com/lululxvi/deepxde/blob/b0d239b1419bf407fe1c23468a470b5f93b65c4e/deepxde/gradients/jacobian.py#L31
> 
> 
> Would this a problem?

What you mentioned is indeed a problem,  the dim computed is wrong. This error has nothing to do with this pull request. It is about the last Forward-mode AD pull request. Since we changed the forward-mode AD to loop-free, then the dim computed is wrong. 

I have used another way to compute the dim of `ys`, and fix a bug in `PDEOperatorCartesianProd`

deepxde/gradients/jacobian.py
```diff
@@ -28,7 +28,10 @@ def __init__(self, ys, xs):
             elif config.autodiff == "forward":
                 # For forward-mode AD, a tuple of a tensor and a callable is passed, 
                 # similar to backend jax.
-                self.dim_y = ys[0].shape[1]
+                if bkd.ndim(ys[0]) == 2:
+                    self.dim_y = 1
+                elif bkd.ndim(ys[0]) == 3:
+                    self.dim_y = ys[0].shape[2]
```
lululxvi, line 34: How about using
```
self.dim_y = ys[0].shape[-1]
```

deepxde/data/pde_operator.py
```diff
@@ -306,8 +306,8 @@ def forward_call(trunk_input):
             losses_bc.append(losses_i)
 
         losses_bc = zip(*losses_bc)
-        losses_bc = [bkd.reduce_mean(bkd.stack(loss, 0)) for loss in losses_bc]
-        losses.append(losses_bc)
+        for loss in losses_bc:
+            losses.append(bkd.reduce_mean(bkd.stack(loss, 0)))
```
lululxvi, line 310: How about
```
losses_bc = [bkd.reduce_mean(bkd.stack(loss, 0)) for loss in losses_bc]
losses.extend(losses_bc)
```

Jerry-Jzy: ok

deepxde/gradients/jacobian.py
Jerry-Jzy: you mean using `self.dim_y = ys[0].shape[-1]`when ndim of `ys[0]` is 3?

lululxvi: Yes, for both cases of bkd.ndim(ys[0]) == 2 and bkd.ndim(ys[0]) == 3.

I see there is one problem. When bkd.ndim(ys[0]) == 2, for PINN, it should be self.dim_y = ys[0].shape[1], but for PI-DeepONet with cartisian product, it is self.dim_y = 1. 

Need to think about how to solve the problem.

Jerry-Jzy: See my latest code

deepxde/gradients/jacobian.py
```diff
@@ -28,7 +27,7 @@ def __init__(self, ys, xs):
             elif config.autodiff == "forward":
                 # For forward-mode AD, a tuple of a tensor and a callable is passed, 
                 # similar to backend jax.
-                self.dim_y = ys[0].shape[1]
+                self.dim_y = ys[0].shape[-1]
```
lululxvi, line 30: As you did reshape, then `ys[0]` is always 2D. So, there is no need of this change.

deepxde/data/pde_operator.py
```diff
@@ -263,18 +263,45 @@ def _losses(self, outputs, loss_fn, inputs, model, num_func, aux=None):
             # Use stack instead of as_tensor to keep the gradients.
             losses = [bkd.reduce_mean(bkd.stack(loss, 0)) for loss in losses]
         elif config.autodiff == "forward":  # forward mode AD
+            if bkd.ndim(outputs) == 2:
```
lululxvi, line 266: You can use `model.net.num_outputs` to get num of outputs.

deepxde/gradients/jacobian.py
Jerry-Jzy: sure

deepxde/data/pde_operator.py
Jerry-Jzy: ok

deepxde/gradients/gradients_forward.py
```diff
@@ -92,9 +92,12 @@ def grad_fn(x):
...
                 # and a callable is returned, so that it is consistent with the argument,
                 # which is also a tuple. This is useful for further computation, e.g.,
-                # Hessian.
+                # Hessian. The original code self.J[j][0][:, i : i + 1] is designed for
+                # the output shape of (batch size, dim), but we find that the code
+                # also works for the output shape of (batch size 1, batch size 2, dim)
+                # such as multiple-output DeepONet.
                 self.J[i, j] = (
-                    self.J[j][0][:, i : i + 1],
+                    self.J[j][0][..., i : i + 1],
```
lululxvi, line 100: Also, is this change necessary?

Jerry-Jzy: this change is no longer needed

deepxde/data/pde_operator.py
```diff
@@ -263,18 +263,45 @@ def _losses(self, outputs, loss_fn, inputs, model, num_func, aux=None):
             # Use stack instead of as_tensor to keep the gradients.
             losses = [bkd.reduce_mean(bkd.stack(loss, 0)) for loss in losses]
         elif config.autodiff == "forward":  # forward mode AD
+            if model.net.num_outputs == 1:
+                is_multi_outputs = False
+                shape0, shape1 = outputs.shape[0], outputs.shape[1]
+            elif model.net.num_outputs > 1:
```
lululxvi, line 269: `elif model.net.num_outputs > 1:` --> `else`

deepxde/data/pde_operator.py
```diff
@@ -263,18 +263,45 @@ def _losses(self, outputs, loss_fn, inputs, model, num_func, aux=None):
...
+                    outputs.shape[0],
+                    outputs.shape[1],
+                    outputs.shape[2],
+                )
 
             def forward_call(trunk_input):
-                return aux[0]((inputs[0], trunk_input))
+                output = aux[0]((inputs[0], trunk_input))
+                if is_multi_outputs:
+                    return output.reshape(shape0 * shape1, shape2)
```
lululxvi, line 280: use `bkd.reshape`

https://github.com/lululxvi/deepxde/blob/b313d967e4fd190d2dcdf441d710e9a39256d6d7/deepxde/backend/backend.py#L120

deepxde/data/pde_operator.py
```diff
@@ -263,18 +263,45 @@ def _losses(self, outputs, loss_fn, inputs, model, num_func, aux=None):
...
-                    inputs[1], (outputs, forward_call), model.net.auxiliary_vars
-                )
+                f = self.pde.pde(inputs[1], (outputs, forward_call), auxiliary_vars)
                 if not isinstance(f, (list, tuple)):
                     f = [f]
+
+            if not is_multi_outputs:
+                outputs = outputs.reshape(shape0, shape1)
+                f = [fi.reshape(shape0, shape1) for fi in f]
+            elif is_multi_outputs:
```
lululxvi, line 302: `else`

deepxde/data/pde_operator.py
Jerry-Jzy: ok

deepxde/data/pde_operator.py
Jerry-Jzy: sure

deepxde/data/pde_operator.py
Jerry-Jzy: ok

deepxde/data/pde_operator.py
```diff
@@ -263,18 +263,47 @@ def _losses(self, outputs, loss_fn, inputs, model, num_func, aux=None):
             # Use stack instead of as_tensor to keep the gradients.
             losses = [bkd.reduce_mean(bkd.stack(loss, 0)) for loss in losses]
         elif config.autodiff == "forward":  # forward mode AD
+            if model.net.num_outputs == 1:
+                is_multi_outputs = False
```
lululxvi, line 267: No need to introduce `is_multi_outputs`. Just use `if model.net.num_outputs ...`

Jerry-Jzy: Sure

deepxde/data/pde_operator.py
```diff
@@ -263,18 +263,45 @@ def _losses(self, outputs, loss_fn, inputs, model, num_func, aux=None):
...
             losses = [bkd.reduce_mean(bkd.stack(loss, 0)) for loss in losses]
         elif config.autodiff == "forward":  # forward mode AD
+            if model.net.num_outputs == 1:
+                shape0, shape1 = outputs.shape[0], outputs.shape[1]
+            else:
+                shape0, shape1, shape2 = (
+                    outputs.shape[0],
+                    outputs.shape[1],
+                    outputs.shape[2],
+                )
```
lululxvi, line 273:
```
shape0, shape1 = outputs.shape[0], outputs.shape[1]
shape2 = 1 if model.net.num_outputs == 1 else outputs.shape[2]
```

Jerry-Jzy: I have refactored  the code

Happy New Year!

deepxde/data/pde_operator.py
```diff
@@ -263,18 +263,29 @@ def _losses(self, outputs, loss_fn, inputs, model, num_func, aux=None):
             # Use stack instead of as_tensor to keep the gradients.
             losses = [bkd.reduce_mean(bkd.stack(loss, 0)) for loss in losses]
         elif config.autodiff == "forward":  # forward mode AD
+            shape0, shape1 = outputs.shape[0], outputs.shape[1]
+            shape2 = 1 if model.net.num_outputs == 1 else outputs.shape[2]
```
lululxvi, line 267: `shape2 = model.net.num_outputs`

deepxde/data/pde_operator.py
```diff
@@ -263,18 +263,29 @@ def _losses(self, outputs, loss_fn, inputs, model, num_func, aux=None):
             # Use stack instead of as_tensor to keep the gradients.
             losses = [bkd.reduce_mean(bkd.stack(loss, 0)) for loss in losses]
         elif config.autodiff == "forward":  # forward mode AD
+            shape0, shape1 = outputs.shape[0], outputs.shape[1]
```
lululxvi, line 266: Use bkd.shape
https://github.com/lululxvi/deepxde/blob/b313d967e4fd190d2dcdf441d710e9a39256d6d7/deepxde/backend/backend.py#L73

deepxde/data/pde_operator.py
Jerry-Jzy: Ok

deepxde/data/pde_operator.py
Jerry-Jzy: sure

deepxde/data/pde_operator.py
```diff
@@ -263,18 +263,30 @@ def _losses(self, outputs, loss_fn, inputs, model, num_func, aux=None):
             # Use stack instead of as_tensor to keep the gradients.
             losses = [bkd.reduce_mean(bkd.stack(loss, 0)) for loss in losses]
         elif config.autodiff == "forward":  # forward mode AD
+            outputs_shape = bkd.shape(outputs)
+            shape0, shape1 = outputs_shape[0], outputs_shape[1]
```
lululxvi, line 267:
```
shape0, shape1 = bkd.shape(outputs)[: 2]
```

deepxde/data/pde_operator.py
```diff
@@ -263,18 +263,30 @@ def _losses(self, outputs, loss_fn, inputs, model, num_func, aux=None):
...
+            shape2 = model.net.num_outputs
```
lululxvi, line 268: Actually no need to define shape2

deepxde/data/pde_operator.py
Jerry-Jzy: Ok

deepxde/data/pde_operator.py
Jerry-Jzy: Sure

deepxde/data/pde_operator.py
lululxvi: You didn't change to `batchsize1, batchsize2`

Jerry-Jzy: already change to `batchsize1`, `batchsize2`

deepxde/data/pde_operator.py
```diff
@@ -263,18 +263,41 @@ def _losses(self, outputs, loss_fn, inputs, model, num_func, aux=None):
             # Use stack instead of as_tensor to keep the gradients.
             losses = [bkd.reduce_mean(bkd.stack(loss, 0)) for loss in losses]
         elif config.autodiff == "forward":  # forward mode AD
+            batchsize1, batchsize2 = bkd.shape(outputs)[:2]
```
lululxvi, line 267: batchsize_total = batchsize1 * batchsize2

deepxde/data/pde_operator.py
```diff
@@ -263,18 +263,41 @@ def _losses(self, outputs, loss_fn, inputs, model, num_func, aux=None):
...
+                        (batchsize1 * batchsize2, model.net.num_outputs),
+                    ),
                 )
                 if not isinstance(f, (list, tuple)):
                     f = [f]
+            f = (
+                [bkd.reshape(fi, (batchsize1, batchsize2)) for fi in f]
+                if model.net.num_outputs == 1
+                else [
+                    bkd.reshape(fi, (batchsize1, batchsize2, model.net.num_outputs))
```
lululxvi, line 296: Why reshape fi?

Jerry-Jzy: need to reshape each element of f back to the original shape

deepxde/data/pde_operator.py
Jerry-Jzy: ok

deepxde/data/pde_operator.py
lululxvi: But the shape is not necessary `model.net.num_outputs`. The number of PDEs is not required to be the same as number of outputs/variables.

Jerry-Jzy: Oh, It is my fault

deepxde/data/pde_operator.py
```diff
@@ -263,18 +263,31 @@ def _losses(self, outputs, loss_fn, inputs, model, num_func, aux=None):
             # Use stack instead of as_tensor to keep the gradients.
             losses = [bkd.reduce_mean(bkd.stack(loss, 0)) for loss in losses]
         elif config.autodiff == "forward":  # forward mode AD
+            batchsize1, batchsize2 = bkd.shape(outputs)[:2]
+            batchsize_total = batchsize1 * batchsize2
 
             def forward_call(trunk_input):
-                return aux[0]((inputs[0], trunk_input))
+                output = aux[0]((inputs[0], trunk_input))
+                return bkd.reshape(output, (batchsize_total, model.net.num_outputs))
```
lululxvi, line 271: Add a comment why we need to reshape.

deepxde/data/pde_operator.py
```diff
@@ -263,18 +263,31 @@ def _losses(self, outputs, loss_fn, inputs, model, num_func, aux=None):
             # Use stack instead of as_tensor to keep the gradients.
             losses = [bkd.reduce_mean(bkd.stack(loss, 0)) for loss in losses]
         elif config.autodiff == "forward":  # forward mode AD
+            batchsize1, batchsize2 = bkd.shape(outputs)[:2]
+            batchsize_total = batchsize1 * batchsize2
```
lululxvi, line 268:
```
shape_2d = (batchsize1 * batchsize2, model.net.num_outputs)
```

deepxde/data/pde_operator.py
```diff
@@ -263,18 +263,31 @@ def _losses(self, outputs, loss_fn, inputs, model, num_func, aux=None):
...
 
             def forward_call(trunk_input):
-                return aux[0]((inputs[0], trunk_input))
+                output = aux[0]((inputs[0], trunk_input))
+                return bkd.reshape(output, (batchsize_total, model.net.num_outputs))
 
             f = []
             if self.pde.pde is not None:
                 # Each f has the shape (N1, N2)
                 f = self.pde.pde(
```
lululxvi, line 276: After reshape the output, then the trunk input and output have different shape. Would this be a problem for dde.gradients?

Jerry-Jzy: I don't think so

deepxde/data/pde_operator.py
Jerry-Jzy: Ok

deepxde/data/pde_operator.py
Jerry-Jzy: added

deepxde/data/pde_operator.py
Jerry-Jzy: I have tested backends pytorch and tensorflow. both work fine

lululxvi: tensorflow v1 or v2?

Jerry-Jzy: v2

lululxvi: Could you also test v1?

Jerry-Jzy: I have tested v1. The parameter requirements for v1 are different from those of v2 and pytorch when computing the derivatives in forward-mode. So, code is not applicable to v1

lululxvi: Then the code needs to be updated.

lululxvi:
Instead of doing reshape here, another solution is modifying dde.gradients to support the output shape of (batch 1, batch 2, dim). This might be a better solution.

deepxde/data/pde_operator.py
Jerry-Jzy: I have updated the code.

Jerry-Jzy:
> Instead of doing reshape here, another solution is modifying dde.gradients to support the output shape of (batch 1, batch 2, dim). This might be a better solution.

I think reshape in PDEOperatorCartesianProd is also a good solution. Since the problem is originally from the output structure difference between PI-DeepONet and PINN. After unifying the output structure, dde.gradients is a general-purpose module for gradent computation

lululxvi:
Can you test another problem?
change this line
https://github.com/lululxvi/deepxde/blob/b313d967e4fd190d2dcdf441d710e9a39256d6d7/examples/operator/advection_aligned_pideeponet.py#L33
to 
```
return dy_t + dy_x + x
```

Jerry-Jzy:
> Can you test another problem? change this line
> 
> https://github.com/lululxvi/deepxde/blob/b313d967e4fd190d2dcdf441d710e9a39256d6d7/examples/operator/advection_aligned_pideeponet.py#L33
> 
> 
> to
> ```
> return dy_t + dy_x + x[:, 0:1]
> ```
> 
> The reverse autodiff should run, but I am not sure about the current forward code.

the forward-mode is not suitable for this case. The reason is not reshape. Since the forward-code is loop-free. so the shape gradient is same as the output shape. The reverse mode is doable because loop is needed then the shape of gradient is same as input shape. That's the exact difference between JVP and VJP. If we want this case working. we need to repeat the x `num_func` times.

lululxvi:
That is why I consider to modify dde.gradients to keep 3D shape; in this way, we can use broadcast, so no repeating x is needed.

Jerry-Jzy:
> That is why I consider to modify dde.gradients to keep 3D shape; in this way, we can use broadcast, so no repeating x is needed.

although 3D shape is kept, Still cannot use broadcast. since the (num_func, num_points) can not broadcast with (num_points, 1).

lululxvi:
(num_func, num_points, 1) can broadcast with (num_points, 1)

lululxvi:
> return dy_t + dy_x + x[:, 0:1]
> ```
> 
> The reverse autodiff should run, but I am not sure about the current forward code.

The key question now is that if the current reshape approach can solve this problem without asking users to repeat x. If not, then we need to find out another way.

lululxvi:
Is current code working for single output? If so, let us make the code work for single output, and just raise an error for multi-outputs. Then we will return to multi-output in another PR.

Jerry-Jzy:
> (num_func, num_points, 1) can broadcast with (num_points, 1)

If we uniformly reshape to 3D rather than 2D maybe can solve the problem of this case

Jerry-Jzy:
> > return dy_t + dy_x + x[:, 0:1]
> > ```
> > 
> > The reverse autodiff should run, but I am not sure about the current forward code.
> > ```
> 
> The key question now is that if the current reshape approach can solve this problem without asking users to repeat x. If not, then we need to find out another way.

See my latest code

lululxvi:
Then it goes back to our previous discussion: dde.gradient currently cannot correctly handle 3D output, such as

https://github.com/lululxvi/deepxde/blob/b313d967e4fd190d2dcdf441d710e9a39256d6d7/deepxde/gradients/jacobian.py#L31

Jerry-Jzy:
> Then it goes back to our previous discussion: dde.gradient currently cannot correctly handle 3D output, such as
> 
> https://github.com/lululxvi/deepxde/blob/b313d967e4fd190d2dcdf441d710e9a39256d6d7/deepxde/gradients/jacobian.py#L31

I have changed to:
`self.dim_y = ys[0].shape[-1]`

This works for PI-DeepONet and PINN

deepxde/gradients/jacobian.py
```diff
@@ -28,7 +28,7 @@ def __init__(self, ys, xs):
...
```
lululxvi, line 31: This conflicts with the basic design of the class
https://github.com/lululxvi/deepxde/blob/b313d967e4fd190d2dcdf441d710e9a39256d6d7/deepxde/gradients/gradients.py#L22

If we want dde.gradients works for 3D output, then we need to go through the entire module and make the changes.

Jerry-Jzy: I have updated the code so that dde.gradients works for 3D output at least for forward-mode AD

lululxvi: This should be a separate PR. We need to make all necessary changes, including all the docstring and comments.

Jerry-Jzy: Ok, The rest of the work is doc and comments

lululxvi: Open a new PR for dde.gradients. After we finish that one, we will then return to this PR.

Jerry-Jzy: Ok

lululxvi:
> Is current code working for single output? If so, let us make the code work for single output, and just raise an error for multi-outputs. Then we will return to multi-output in another PR.

How about single output case?

Jerry-Jzy:
> > Is current code working for single output? If so, let us make the code work for single output, and just raise an error for multi-outputs. Then we will return to multi-output in another PR.
> 
> How about single output case?

single output works fine

lululxvi:
Then clean the code only for single output. We merge the single output code first. Or the current code in the main branch is good?

Jerry-Jzy:
> Then clean the code only for single output. We merge the single output code first. Or the current code in the main branch is good?

I have rolled back the code

PR #1962, opened 2025-03-13
Rename TrainState.epoch as TrainState.iteration to better represent its actual meaning.

KangyuWeng:
There are three commits. The first commit renames epoch as iteration. TrainState.epoch can still be used for now, but a warning will be raised each time it is used. The second commit changes all TrainState.epoch to TrainState.iteration in deepxde. The third commit replaces epochs in the document with iterations because epochs in Model.train() is depreciated long ago.

### DIFF
--- a/deepxde/callbacks.py
+++ b/deepxde/callbacks.py
@@ -151,7 +151,7 @@ class ModelCheckpoint(Callback):
                 if self.verbose > 0:
                     print(
                         "Epoch {}: {} improved from {:.2e} to {:.2e}, saving model to {} ...\n".format(
-                            self.model.train_state.epoch,
+                            self.model.train_state.iteration,
                             self.monitor,
                             self.best,
                             current,
@@ -224,7 +224,7 @@ class EarlyStopping(Callback):
             self.best = np.inf if self.monitor_op == np.less else -np.inf
 
     def on_epoch_end(self):
-        if self.model.train_state.epoch < self.start_from_epoch:
+        if self.model.train_state.iteration < self.start_from_epoch:
             return
         current = self.get_monitor_value()
         if self.monitor_op(current - self.min_delta, self.best):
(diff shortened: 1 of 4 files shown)


deepxde/model.py
```diff
@@ -1188,6 +1189,24 @@ def __init__(self):
         self.best_ystd = None
         self.best_metrics = None
 
+    @property
+    def epoch(self):
+        warnings.warn(
+            "TrainState.epoch() is deprecated and will be removed in a future version. Use TrainState.step instead.",
```
lululxvi, line 1195: `TrainState.epoch()` --> `TrainState.epoch`

KangyuWeng: Has been corrected. I revised the commit to avoid adding a new commit only for a small problem.

deepxde/model.py
```diff
@@ -1188,6 +1189,24 @@ def __init__(self):
...
+    def epoch(self):
+        warnings.warn(
+            "TrainState.epoch is deprecated and will be removed in a future version. Use TrainState.iteration instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return self.iteration
+    
+    @epoch.setter
+    def epoch(self, value):
```
lululxvi, line 1202: Remove `setter`. `TrainState` is not supposed by users to set epoch directly.

KangyuWeng: I submitted a new commit to remove setter.

deepxde/callbacks.py
```diff
@@ -151,7 +151,7 @@ def on_epoch_end(self):
                 if self.verbose > 0:
                     print(
                         "Epoch {}: {} improved from {:.2e} to {:.2e}, saving model to {} ...\n".format(
-                            self.model.train_state.epoch,
+                            self.model.train_state.iteration,
```
lululxvi, line 154: Need to modify the comment in line 119

KangyuWeng: There are more functions and comments containing "epoch(s)" in callbacks.py. Do we need to modify them? I am not sure about the difference between     
def on_epoch_begin(self):
        """Called at the beginning of every epoch."""
def on_epoch_end(self):
        """Called at the end of every epoch."""
 and
     def on_batch_begin(self):
        """Called at the beginning of every batch."""

    def on_batch_end(self):
        """Called at the end of every batch."""
        
That's why I did not modify them.

PR #1968, opened 2025-03-20
JAX BACKEND: add regularizer

bonneted:
Add regularizer for JAX BACKEND.

For Jax adding the regularization is straightforward: you apply the regularization function to the nn_params and add it to the loss. see https://github.com/google/flax/discussions/1654

I tried to stick to TensorFlow and paddle architecture: regularizer is a net property init from the regularization parameter. (even though regularizer is not used in the net here)
In that case, the regularizer is simply the function to be applied to the net parameters to compute the loss in model.py


lululxvi:
Why the build failed?

bonneted:
According to github copilot:
"""
The job failed due to an error while building the dm-tree package. The specific issue is that the download of pybind11 from the specified URL failed with the status code 52, indicating "Server returned nothing (no headers, no data)".

Suggested Solution
To resolve this issue, you can try the following steps:

Check the URL: Ensure that the URL for pybind11 is correct and accessible. If the URL is outdated, update it to the latest version.

Update dm-tree: Ensure that you are using the latest version of the dm-tree package, as it may have fixed the issue with the dependency.

Retry the Build: Sometimes, transient network issues can cause such errors. Retrying the build might resolve the issue.

Modify the Workflow File: If the issue persists, modify the workflow to include the installation of pybind11 before building dm-tree.
"""

So this is probably a network issue during the build

deepxde/backend/jax/tensor.py
```diff
@@ -175,3 +175,15 @@ def zeros(shape, dtype):
...
+def l1_regularization(l1):
+    return lambda params: l1 * jnp.sum(jnp.concatenate([jnp.abs(w).flatten() for w in params]))
+
+
+def l2_regularization(l2):
+    return lambda params: l2 * jnp.sum(jnp.concatenate([jnp.square(w).flatten() for w in params]))
+
+
+def l1_l2_regularization(l1, l2):
+    return lambda params: l1_regularization(l1)(params) + l2_regularization(l2)(params)
```
lululxvi, line 189: Add a blank line at the end.

deepxde/model.py
```diff
@@ -443,6 +443,9 @@ def outputs_fn(inputs):
             losses = jax.numpy.asarray(losses)
             if self.loss_weights is not None:
                 losses *= jax.numpy.asarray(self.loss_weights)
+            if self.net.regularizer is not None:
```
lululxvi, line 446: Remove the TODO in line 440

deepxde/nn/jax/fnn.py
```diff
@@ -15,11 +16,17 @@ class FNN(NN):
...
     activation: Any
     kernel_initializer: Any
+    regularization: Any = None
 
     params: Any = None
     _input_transform: Callable = None
     _output_transform: Callable = None
 
+    @property
+    def regularizer(self):
```
lululxvi, line 26: This can be moved to the base class `NN`

deepxde/nn/jax/fnn.py
```diff
@@ -8,13 +8,13 @@
...
 from .. import initializers
 
-
 class FNN(NN):
     """Fully-connected neural network."""
 
     layer_sizes: Any
     activation: Any
     kernel_initializer: Any
+    regularization: Any = None
```
lululxvi, line 17: Is it possible to also move this line to `NN`?

bonneted: If I do so, I get:
TypeError: non-default argument 'layer_sizes' follows default argument

Because layer_sizes then follows regularization argument.
I guess this might be the reason why no default argument is put in `NN` but simply in comments :
```
    # All sub-modules should have the following variables:
    # params: Any = None
    # _input_transform: Optional[Callable] = None
    # _output_transform: Optional[Callable] = None

```

And then put in all nn submodules after the non-default arguments, for example in `fnn`:
```
class FNN(NN):
    """Fully-connected neural network."""

    layer_sizes: Any
    activation: Any
    kernel_initializer: Any
    regularization: Any = None

    params: Any = None
    _input_transform: Callable = None
    _output_transform: Callable = None
```

So what I can do is add regularization in the commented lines of the `nn` module:
```
    # All sub-modules should have the following variables:
    # params: Any = None
    # _input_transform: Optional[Callable] = None
    # _output_transform: Optional[Callable] = None
    # regularization: Any = None
```

deepxde/nn/jax/nn.py
```diff
@@ -1,5 +1,5 @@
 from flax import linen as nn
-
+from .. import regularizers
```
lululxvi, line 2:
```
from flax import linen as nn

from .. import regularizers


class NN(nn.Module):
```

deepxde/nn/jax/nn.py
```diff
@@ -1,14 +1,21 @@
 from flax import linen as nn
 
+from .. import regularizers
```
lululxvi, line 4: Add one more black line here.

deepxde/nn/jax/fnn.py
```diff
@@ -8,13 +8,13 @@
 from .. import activations
 from .. import initializers
 
-
 class FNN(NN):
```
lululxvi, line 11: Keep the blank line

PR #1973, opened 2025-03-30
Add Apple MPS GPU support

kyouma:
There is a Pytorch bug that occurs when an optimizer is created for a network which is on an MPS GPU (https://github.com/pytorch/pytorch/issues/149184), thus a trick to avoid it is needed.

### DIFF
--- a/deepxde/backend/pytorch/tensor.py
+++ b/deepxde/backend/pytorch/tensor.py
@@ -20,6 +20,9 @@ if torch.cuda.is_available():
         torch.set_default_device("cuda")
     else:
         torch.set_default_tensor_type(torch.cuda.FloatTensor)
+elif torch.backends.mps.is_available():
+    torch.set_default_device("mps")
+    torch._dynamo.disable()  # A temporary trick to evade the Pytorch MPS bug (https://github.com/pytorch/pytorch/issues/149184)
 
 
 lib = torch


kyouma:
Excuse me, I do not know how the build process is organized. The error message says:
```
RuntimeError: MPS backend out of memory (MPS allocated: 0 bytes, other allocations: 0 bytes, max allowed: 7.93 GB).
Tried to allocate 256 bytes on shared pool.
Use PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0 to disable upper limit for memory allocations (may cause system failure).
```
Maybe the `PYTORCH_MPS_HIGH_WATERMARK_RATIO` parameter is enough or maybe [this Pytorch forum link](https://discuss.pytorch.org/t/mps-back-end-out-of-memory-on-github-action/189773) ("Switching `runs-on` from `macos-latest` to `macos-13`") contains the way to solve the build issue.

deepxde/backend/pytorch/tensor.py
```diff
@@ -20,6 +20,35 @@
         torch.set_default_device("cuda")
     else:
         torch.set_default_tensor_type(torch.cuda.FloatTensor)
+elif torch.backends.mps.is_available():
```
lululxvi, line 23: What is the minimum torch version required?

PR #1975, opened 2025-04-03
PyTorch: specify weights_only for torch.load()

vl-dud:
It makes sense to specify parameter `weights_only` for `torch.load` explicitly. The default value has changed from False to True in the latest version (2.6).


lululxvi:
Just to confirm, using `weights_only=True`, it stil loads optimizer state, right?

vl-dud:
I checked, `model.opt.state_dict()` is the same when using `weights_only=True` and `weights_only=False`

PR #1997, opened 2025-06-10
add mini-batch for PointSetOperatorBC

enigne:
Add mini-batch option to PointSetOperatorBC, similar approach as in PointSetBC


deepxde/icbc/boundary_conditions.py
```diff
@@ -248,19 +248,37 @@ class PointSetOperatorBC:
...
+        if batch_size is not None:  # batch iterator and state
+            if backend_name not in ["pytorch", "paddle"]:
+                raise RuntimeError(
+                    "batch_size only implemented for pytorch and paddle backend"
+                )
+            self.batch_sampler = data.sampler.BatchSampler(len(self), shuffle=shuffle)
+            self.batch_indices = None
+
     def collocation_points(self, X):
+        if self.batch_size is not None:
```
lululxvi, line 274: Would this code work? `self.batch_size` is not defined.

PR #2023, opened 2025-09-28
Update Poisson_Dirichlet_1d_exactBC.py

tianyeeT:


lululxvi:
> > Does this code work well for PyTorch?
> 
> Yeah, I tested it locally, and it turned out well.

Then update the first line of the file by adding pytorch, similar to other examples.

examples/pinn_forward/Poisson_Dirichlet_1d_exactBC.py
```diff
@@ -1,4 +1,4 @@
-"""Backend supported: tensorflow.compat.v1, tensorflow, jax, paddle"""
+"""Backend supported: tensorflow.compat.v1, tensorflow, jax, pytorch, paddle"""
```
lululxvi, line 1:
```
"""Backend supported: tensorflow.compat.v1, tensorflow, pytorch, jax, paddle"""
```

Issue #1963, opened 2025-03-14
Boundary Conditions for geometries defined by discrete points


lm0bo:
Hello all,

I am working with a geometry where the walls are defined by a set of discrete points rather than a continuous surface representation. I need to correctly apply Dirichlet boundary conditions using the coordinates of these wall points, but I am encountering some challenges in doing so.

Thank you in advance!

kyouma:
Hello. I would try `PointSetBC` instead of `DirichletBC`, or `DirichletBC` with interpolation inside `func` and `on_boundary` parameters.

lm0bo:
Thank you for your response! I have the coordinates of the points on the wall of a complex geometry, and I want to impose a Dirichlet boundary condition with a velocity of 0 at those points.

I have defined the wall as follows:
geom_wall = dde.geometry.PointCloud(wall_coords, boundary_points=wall_coords)

And the boundary condition:
bc_u_wall = dde.icbc.PointSetBC(wall_coords, np.zeros((wall_coords.shape[0], 1)), component=0)

However, the solution obtained from the network using PointSetBC for the boundary conditions is not accurate.

kyouma:
Must your equation be solved inside the 3D geometry (volume) or on the wall (surface)? Probably you should use `dde.geometry.PointCloud(INNER_VOLUME_POINTS, boundary_points=WALL_POINTS)` and `dde.icbc.PointSetBC(WALL_POINTS,...)`?

Also, soft boundary conditions (conditions set by loss functions) cannot provide accurate solution, there will always be some residue. You can control its size using loss weights in `model.compile()`. For choosing the weights see the "_An Expert's Guide to Training Physics-informed Neural Networks_" paper. You could also try to find complex transforms to obtain hard constrains on the boundary.

LordPontiag:
> Hello all,
> 
> I am working with a geometry where the walls are defined by a set of discrete points rather than a continuous surface representation. I need to correctly apply Dirichlet boundary conditions using the coordinates of these wall points, but I am encountering some challenges in doing so.
> 
> Thank you in advance!

Hello dear @lm0bo , as @kyouma mentioned using `PointSetBC ` is a perfect alternative for `DirichletBC`. How ever you can use following code that I wrote and used and it works correctly.

```
PINN_walls =[]
# Define boundary conditions
def boundary_wall(X, on_boundary):
    # Convert X to a NumPy array for consistency
    X = np.array(X)
[...]

PR #2038, opened 2025-11-25
Use KDTree to implement PointSet and PointCloud more efficiently

nynauy:
The brute-force matching algorithm in the `PointCloud` is extremely costly in both time and space complexity ( $O(mnk)$ ). When working with hundreds of thousands of points, my process runs for a long time and is often terminated by the OOM killer. Implementing the matching using SciPy's `KDTree` significantly reduces both memory usage and execution time.

### DIFF
--- a/deepxde/geometry/pointcloud.py
+++ b/deepxde/geometry/pointcloud.py
@@ -34,7 +34,7 @@ class PointCloud(Geometry):
                 self.boundary_normals = np.asarray(
                     boundary_normals, dtype=config.real(np)
                 )
-                self._boundary_normal_query = self._boundary_point_set.values_to_func(self.boundary_normals, default_value=None)
+                self._boundary_normal_query = self._boundary_point_set.values_to_func(self.boundary_normals)
         super().__init__(
             len(self.points[0]),
             (np.amin(all_points, axis=0), np.amax(all_points, axis=0)),
--- a/deepxde/utils/external.py
+++ b/deepxde/utils/external.py
@@ -38,7 +38,7 @@ class PointSet:
         distance = np.asarray(distance, self.points.dtype)
         return isclose(distance, 0)
 
-    def values_to_func(self, values, default_value=0):
+    def values_to_func(self, values, default_value=None):
         """Convert the pairs of points and values to a callable function.
 
         Args:


lululxvi:
Have you tested if the new code has the same results as the old code?

nynauy:
Perhaps we should modify [`utils.external.PointSet`](https://github.com/lululxvi/deepxde/blob/57f754f42fa5e41e3be46872426206adbd3622e0/deepxde/utils/external.py#L14) (which currently appears unused within this library) and just delegate `PointCloud.points` and `PointCloud.boundary_points` to it. This would also facilitate reuse by users, like in #1963.
@lululxvi What do you think?

deepxde/geometry/pointcloud.py
```diff
@@ -19,59 +19,43 @@ class PointCloud(Geometry):
 
     def __init__(self, points, boundary_points=None, boundary_normals=None):
         self.points = np.asarray(points, dtype=config.real(np))
-        self.num_points = len(points)
+        self.num_points = len(self.points)
```
lululxvi: The original code of `len(points)` is ok. No need to change.

nynauy: I made this modification because I encountered some uncommon cases: when `points` is not a `numpy.ndarray` but supports `__array__`, `np.asarray` can accept it, allowing it to be used for constructing a `PointCloud`. However, `points` may not have a `__len__` method with $O(1)$ complexity. Of course, users can call `np.asarray` themselves, then pass the array to `PointCloud.__init__`. But providing convenience is always welcome.

deepxde/utils/external.py
```diff
@@ -32,34 +34,35 @@ def inside(self, x):
...
     def values_to_func(self, values, default_value=0):
         """Convert the pairs of points and values to a callable function.
 
         Args:
             values: A NumPy array of shape (`N`, `dy`). `values[i]` is the `dy`-dim
                 function value of the `i`-th point in this point set.
-            default_value (float): The function value of the points not in this point
-                set.
+            default_value: A scalar or vector with the same shape as `values[i]`, or `None`.
+                If not `None`, The function returns it for the points not in this point set.
```
lululxvi: `The` --> `the`

deepxde/utils/external.py
```diff
@@ -32,34 +34,35 @@ def inside(self, x):
...
+                If `None`, the function always returns the nearest neighbor value.
 
         Returns:
             A callable function. The input of this function should be a NumPy array of
                 shape (?, `dx`).
         """
+        values = np.asarray(values)
```
lululxvi:
```
values = np.asarray(values, dtype=config.real(np))
```

nynauy: I prefer not to do type conversions in `PointSet`, since the old code doesn't either. Besides, unlike `geometry.*` which mainly serves `data.*`, `utils.PointSet` is more general and can easily be used elsewhere.
BTW, the new `PointCloud` does perform type conversions.

deepxde/utils/external.py
```diff
@@ -32,34 +34,35 @@ def inside(self, x):
...
-            # A single point
-            return np.any(np.all(isclose(x, self.points), axis=1))
-        if x.ndim == 2:
-            # A list of points
-            return np.any(
-                np.all(isclose(x[:, np.newaxis, :], self.points), axis=-1),
-                axis=-1,
-            )
+        distance, _ = self._kdtree.query(x, workers=-1)
+        distance = np.asarray(distance, self.points.dtype)
```
lululxvi: Why do we need this line?

PR #2052, opened 2025-12-17
Add gravity pendulum example

vl-dud:
Add a PI-DeepONet example for the gravity pendulum problem:

$\theta''(t) = -\sin(\theta(t)) + u(t), \qquad t \in [0, T]$

$\theta(0) = 0, \qquad \theta'(0) = 0$

- $t$: time  
- $\theta(t)$: pendulum angle
- $u(t)$: external forcing

### DIFF
--- a/docs/demos/operator.rst
+++ b/docs/demos/operator.rst
@@ -29,6 +29,7 @@ PI-DeepONet
 - `Diffusion reaction equation with aligned points <https://github.com/lululxvi/deepxde/tree/master/examples/operator/diff_rec_aligned_pideeponet.py>`_
 - `Diffusion reaction equation with unaligned points <https://github.com/lululxvi/deepxde/tree/master/examples/operator/diff_rec_unaligned_pideeponet.py>`_
 - `Stokes flow with aligned points <https://github.com/lululxvi/deepxde/tree/master/examples/operator/stokes_aligned_pideeponet.py>`_
+- `Gravity pendulum with unaligned points <https://github.com/lululxvi/deepxde/tree/master/examples/operator/gravity_pendulum_unaligned_pideeponet.py>`_
 
 .. toctree::
    :maxdepth: 1
--- a/docs/demos/operator.rst
+++ b/docs/demos/operator.rst
@@ -19,6 +19,7 @@ PI-DeepONet
    :maxdepth: 1
 
    operator/poisson.1d.pideeponet
+   operator/gravity_pendulum_unaligned_pideeponet
 
 - `Antiderivative operator with aligned points <https://github.com/lululxvi/deepxde/tree/master/examples/operator/antiderivative_aligned_pideeponet.py>`_
 - `Antiderivative operator with unaligned points <https://github.com/lululxvi/deepxde/tree/master/examples/operator/antiderivative_unaligned_pideeponet.py>`_
--- a/examples/operator/gravity_pendulum_unaligned_pideeponet.py
+++ b/examples/operator/gravity_pendulum_unaligned_pideeponet.py
@@ -1,0 +1,90 @@
+"""Backend supported: tensorflow.compat.v1, tensorflow, pytorch, paddle"""
+import deepxde as dde
+import matplotlib.pyplot as plt
+import numpy as np
+
+sin = dde.backend.sin
+
+
+def pde(x, y, v):
+    # theta''(t) = -sin(theta(t)) + u(t)
+    dy_tt = dde.grad.hessian(y, x, i=0, j=0)
+    return dy_tt + sin(y) - v
+
+
+def boundary_ic(x, on_boundary):
+    return on_boundary and np.isclose(x[0], 0.0)
+
+
+geom = dde.geometry.TimeDomain(0, 1)
+ic1 = dde.icbc.DirichletBC(geom, lambda x: 0.0, boundary_ic)  # theta(0) = 0
+ic2 = dde.icbc.NeumannBC(geom, lambda x: 0.0, boundary_ic)  # theta'(0) = 0
+pde = dde.data.PDE(geom, pde, [ic1, ic2], num_domain=200, num_boundary=20)
+func_space = dde.data.GRF(length_scale=0.2)
+eval_pts = np.linspace(0, 1, num=50)[:, None]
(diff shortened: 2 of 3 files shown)


lululxvi:
Update https://github.com/lululxvi/deepxde/blob/master/docs/demos/operator.rst so the example shows up at https://deepxde.readthedocs.io/en/latest/demos/operator.html

vl-dud:
> Update https://github.com/lululxvi/deepxde/blob/master/docs/demos/operator.rst so the example shows up at https://deepxde.readthedocs.io/en/latest/demos/operator.html

Done

examples/operator/gravity_pendulum_unaligned_pideeponet.py
```diff
@@ -0,0 +1,90 @@
+"""Backend supported: tensorflow.compat.v1, tensorflow, pytorch, paddle"""
+import deepxde as dde
+import matplotlib.pyplot as plt
+import numpy as np
+
+sin = dde.backend.sin
+
+
+def pde(x, y, v):
+    # theta''(t) = -sin(theta(t)) + u(t)
```
echen5503, line 8: Nitpick: Why not use theta unicode character?

docs/demos/operator.rst
```diff
@@ -29,6 +29,7 @@ PI-DeepONet
 - `Diffusion reaction equation with aligned points <https://github.com/lululxvi/deepxde/tree/master/examples/operator/diff_rec_aligned_pideeponet.py>`_
 - `Diffusion reaction equation with unaligned points <https://github.com/lululxvi/deepxde/tree/master/examples/operator/diff_rec_unaligned_pideeponet.py>`_
 - `Stokes flow with aligned points <https://github.com/lululxvi/deepxde/tree/master/examples/operator/stokes_aligned_pideeponet.py>`_
+- `Gravity pendulum with unaligned points <https://github.com/lululxvi/deepxde/tree/master/examples/operator/gravity_pendulum_unaligned_pideeponet.py>`_
```
echen5503, line 33: It may be a good idea to have some documentation for this code. See https://github.com/lululxvi/deepxde/pull/2056/changes. 

It seems to be uniquely valuable here because operator learning has few examples with documentation.

examples/operator/gravity_pendulum_unaligned_pideeponet.py
```diff
@@ -0,0 +1,90 @@
...
+model = dde.Model(data, net)
+model.compile("adam", lr=0.0005)
+model.train(iterations=50000)
+
+
+def solve_pendulum(u_func, t):
+    """
+    Solve the forced pendulum ODE using a 4th-order Runge–Kutta method
+
+    theta''(t) = -sin(theta(t)) + u(t),
```
echen5503, line 40: PDE expression already duplicated above

examples/operator/gravity_pendulum_unaligned_pideeponet.py
```diff
@@ -0,0 +1,90 @@
+"""Backend supported: tensorflow.compat.v1, tensorflow, pytorch, paddle"""
+import deepxde as dde
+import matplotlib.pyplot as plt
+import numpy as np
+
+sin = dde.backend.sin
```
echen5503: This renamed sin function is only used twice. It may be better to just use dde.backend.sin for clarity.

examples/operator/gravity_pendulum_unaligned_pideeponet.py
```diff
@@ -0,0 +1,90 @@
...
+    # theta''(t) = -sin(theta(t)) + u(t)
+    dy_tt = dde.grad.hessian(y, x, i=0, j=0)
+    return dy_tt + sin(y) - v
+
+
+def boundary_ic(x, on_boundary):
+    return on_boundary and np.isclose(x[0], 0.0)
+
+
+geom = dde.geometry.TimeDomain(0, 1)
```
echen5503, line 17: This part is likely difficult to read for newcomers. Comments / Documentation would probably help.

examples/operator/gravity_pendulum_unaligned_pideeponet.py
```diff
@@ -0,0 +1,90 @@
...
+    # sinusoid, amplitude 0.5, freq 2.0 Hz
+    return 0.5 * np.sin(4.0 * np.pi * t_local)
+
+
+# Compare solver and PI-DeepONet
+t = np.arange(0, 1, 0.01)
+v = u_sin(eval_pts)
+solved_theta = solve_pendulum(u_sin, t)
+predicted_theta = model.predict((np.tile(v.T, (t.size, 1)), t.reshape(-1, 1)))
+
```
echen5503, line 80: Could be helpful to add error metrics here

echen5503 (changes requested):
Hi, I'm relatively new here so please take my comments with a grain of salt, but I found some potential issues with the code. Reviewing to learn more right now :)

examples/operator/gravity_pendulum_unaligned_pideeponet.py
vl-dud: This is inconsistent with the other examples, since unicode isn’t used for Greek letters anywhere else.

examples/operator/gravity_pendulum_unaligned_pideeponet.py
vl-dud: This doesn’t seem like a significant issue to me.

examples/operator/gravity_pendulum_unaligned_pideeponet.py
vl-dud: I’m not sure it makes much sense to use metrics for just one example of u(t).

vl-dud:
@echen5503 Thank you for your careful review. I have updated the parts that seemed most important to address.

examples/operator/gravity_pendulum_unaligned_pideeponet.py
echen5503: Right, it's already plotted. SG

echen5503:
@vl-dud great, this looks good to me. 
@lululxvi I recommend merge. Note that the paddlepaddle build failure is not the fault of this PR, seems like the origin of paddlepaddle might have been moved.

Issue #2073, opened 2026-03-28
Strange behavior of a boundary condition loss on the test sample


kyouma:
Hello. I apologize if there have been a similar topic already, but I have not found it in these issues.

I have been experimenting with a simple time-dependent equation in a rectangular 2D domain, and have found out a strange behavior of BC losses. _Or maybe I am doing something wrong._

<details>
<summary>The image of the PDE system</summary>
<img width="1468" height="1182" alt="Image" src="https://github.com/user-attachments/assets/40674fea-3afa-402b-bbb2-866ed7215b65" />
</details>

Below is the code divided into blocks:

<details>
<summary>Imports and random seed setting</summary>

```
import os
os.environ['DDE_BACKEND'] = 'pytorch'
os.environ['CUDA_VISIBLE_DEVICES'] = '0'


import random
import types

import deepxde as dde
import matplotlib.pyplot as plt
import numpy as np
import torch
import tqdm
import tqdm.auto


random.seed(0)
np.random.seed(0)
dde.config.set_random_seed(0)
```
</details>

<details>
<summary>Domain limits</summary>

```
LEFT = 0
RIGHT = np.pi
BOTTOM = -np.pi
TOP = np.pi
T = 1.0
```
</details>

<details>
<summary>PDE, BC, IC, derivatives and exact solution functions</summary>

```
def pde(xyt, u):
    x = xyt[:, 0:1]
    y = xyt[:, 1:2]
    t = xyt[:, 2:3]

    du_t = dde.grad.jacobian(u, xyt, i=0, j=2)
    du_xx = dde.grad.hessian(u, xyt, i=0, j=0)
    du_yy = dde.grad.hessian(u, xyt, i=1, j=1)

    f = torch.exp(-1.3* t) * (77/250 * torch.sin(0.6 * x) + 3807/400 * torch.cos(2.5 * y))

    return du_t - 1.9 * (du_xx + du_yy) - f


def u_torch(xyt):
    '''Same as u_numpy(), but written with PyTorch.'''
    if not isinstance(xyt, torch.Tensor):
        xyt = torch.tensor(xyt)
    x = xyt[:, 0:1]
    y = xyt[:, 1:2]
    t = xyt[:, 2:3]
    return torch.exp(-1.3 * t) * (0.5 * torch.sin(0.6 * x) + 0.9 * torch.cos(2.5 * y)) + 2.4


def u_numpy(xyt):
    '''Same as u_torch(), but written with NumPy.'''
[...]

PR #2076, opened 2026-03-30
Fix coordinate mismatch in test BC loss after resample_train_points(bc_points=True)

echen5503:
Inspired by @kyouma issue #2073 and his code, but updated to a complete fix with backwards compatibility and tests.
## Bug

When `PDEPointResampler` resampled boundary condition points during training, the **test BC losses became meaningless** — exhibiting a sudden ~1000× spike immediately after the first resample, while train loss and solution metrics stayed correct.

### Root cause

`PDE.losses()` computes the BC residual as:

```python
error = bc.error(self.train_x, inputs, outputs, beg, end)
```

Inside `bc.error()`, `X[beg:end]` supplies the **coordinates** used to evaluate reference values and boundary normals, while `inputs[beg:end]` / `outputs[beg:end]` are the network's **predictions** at those same points. They must describe the same physical locations.

During training this holds: `inputs` is a tensor of `train_x`. The problem is that `Data.losses_test()` (inherited unchanged by `PDE`) delegates to `self.losses()`, which **always** reads `self.train_x` for `X` — even when `inputs` and `outputs` come from `test_x`.

After `resample_train_points(bc_points=True)`:

| | Before resample | After resample |
|---|---|---|
| `self.train_x[beg:end]` | old BC coords | **new** BC coords |
| `self.test_x[beg:end]` | old BC coords | old BC coords (**never reset**) |
| `train_state.X_test[beg:end]` | old BC coords | old BC coords (**set once at training start, never updated**) |

`_test()` evaluates the network on stale `train_state.X_test` (old BC coords) to produce `inputs`/`outputs`, then `losses_test()` calls `bc.error(self.train_x, ...)` where `self.train_x[beg:end]` now holds **new** BC coordinates. For a `NeumannBC` this means the normal derivative is taken at old positions but subtracted from the reference value at new positions — a residual that does not represent the true BC error at either set of points.

### Affected classes

| Class | Affected? |
|---|---|
| `PDE` / `TimePDE` | **Yes** — `losses_test()` inherits from `Data` and calls `losses()` |
| `PDEOperator` | **Yes** — same inheritance, `losses()` uses `self.train_x[1]` |
| `PDEOperatorCartesianProd` | **Yes** — `losses_test()` calls `_losses()` which hardcodes `self.train_x[1]` |
| `IDE` | No — `losses_test()` override returns zero for all BC losses |
| `FPDE` / `TimeFPDE` | No — same reason as IDE |

---

## Fix

Three coordinated changes are required. Any two alone leave a residual mismatch.

### Fix 1 — `PDE.losses()` + new `PDE.losses_test()` (`deepxde/data/pde.py`)

`losses()` gains an optional `X_bc` parameter (defaulting to `self.train_x`). A new `losses_test()` override passes `X_bc=self.test_x`:

```python
def losses(self, ..., X_bc=None):
    if X_bc is None:
        X_bc = self.train_x
    ...
    error = bc.error(X_bc, inputs, outputs, beg, end)

def losses_test(self, ...):
    return self.losses(..., X_bc=self.test_x)
```

### Fix 2 — `PDE.resample_train_points()` (`deepxde/data/pde.py`)

When `bc_points=True`, `test_x` is now reset and immediately regenerated so it stays consistent with the new `train_x_bc`:

```python
if bc_points:
    self.train_x_bc = None
    self.test_x, self.test_y, self.test_aux_vars = None, None, None
...
self.train_next_batch()
if bc_points:
    self.test()
```

Without this, Fix 1 alone computes the BC error at stale (but self-consistent) old test BC coordinates instead of the newly resampled ones.

### Fix 3 — `PDEResampler.on_epoch_end()` (`deepxde/callbacks.py`)

After resampling, the callback pushes the regenerated `test_x` into `train_state.X_test`:

```python
self.model.data.resample_train_points(self.pde_points, self.bc_points)
if self.bc_points:
    self.model.train_state.set_data_test(*self.model.data.test())
```

`train_state.X_test` is what `_test()` feeds into the network. Without this, the network is still evaluated at stale coords even after Fixes 1 and 2 update `data.test_x`.

### Same fixes for `PDEOperator` and `PDEOperatorCartesianProd` (`deepxde/data/pde_operator.py`)

- `PDEOperator.losses()` gains `X_bc` / `aux_var` parameters; new `losses_test()` passes `self.test_x[1]` and `self.test_aux_vars`.
- `PDEOperator.resample_train_points()` resets and regenerates `test_x` when `bc_points=True`.
- `PDEOperatorCartesianProd._losses()` gains `X_bc`; `losses_train()` passes `self.train_x[1]` and `losses_test()` passes `self.test_x[1]`.

### Docstring added to `BC.error()` (`deepxde/icbc/boundary_conditions.py`)

Documents the requirement that `X[beg:end]` and `inputs[beg:end]` must refer to the same physical points.

---

## Backwards compatibility

All existing behaviour is fully preserved for usage that does not involve `bc_points=True` resampling:

- `losses()` defaults `X_bc=None` → `self.train_x`, identical to the previous hardcoded value. Subclass overrides without the new parameter continue to work.
- `losses_train()` is not overridden; it still delegates to `losses()` with `X_bc=None`. Training losses are unaffected.
- `PDEResampler` with `bc_points=False` (the default) skips the new `set_data_test()` call entirely.
- `resample_train_points(bc_points=False)` does not reset or regenerate `test_x`.
- `IDE` and `FPDE` / `TimeFPDE` are untouched.

The only observable behaviour change is the intended one: after `resample_train_points(bc_points=True)`, reported test BC losses now correctly reflect the network's BC residual on the new points.

---

## Tests

`test_bc_resample.py` adds five regression tests:

1. **`test_bc_points_equal_before_resample`** — sanity check that `test_x` and `train_x` start with identical BC slices.
2. **`test_test_x_refreshed_after_resample`** — Fix 2: `data.test_x[beg:end]` equals the new `data.train_x[beg:end]` after resampling.
3. **`test_train_state_X_test_refreshed_by_callback`** — Fix 3: `train_state.X_test[beg:end]` is updated after the callback fires.
4. **`test_losses_test_passes_test_x_to_bc_error`** — Fix 1: `losses_test()` passes `self.test_x` (not `self.train_x`) as `X` to `bc.error()`, verified by intercepting the call with a sentinel coordinate value.
5. **`test_github_issue_timepde_neumann_bc`** — end-to-end regression for the reported case: `TimePDE` on a 2-D+time rectangle with two `NeumannBC`s, one `DirichletBC`, one `NeumannBC`, and an `IC`, with `PDEPointResampler(bc_points=True)`. Asserts that after one resample cycle, `train_x[beg:end] == test_x[beg:end]` for every BC and that `train_state.X_test` is in sync.

### DIFF
--- a/deepxde/callbacks.py
+++ b/deepxde/callbacks.py
@@ -596,6 +596,8 @@ class PDEPointResampler(Callback):
             return
         self.epochs_since_last_resample = 0
         self.model.data.resample_train_points(self.pde_points, self.bc_points)
+        if self.bc_points:
+            self.model.train_state.set_data_test(*self.model.data.test())
 
         if not np.array_equal(self.num_bcs_initial, self.model.data.num_bcs):
             print("Initial value of self.num_bcs:", self.num_bcs_initial)
--- a/deepxde/data/pde.py
+++ b/deepxde/data/pde.py
@@ -135,7 +135,7 @@ class PDE(Data):
         self.train_next_batch()
         self.test()
 
-    def losses(self, targets, outputs, loss_fn, inputs, model, aux=None):
+    def losses(self, targets, outputs, loss_fn, inputs, model, aux=None, X_bc=None):
         if backend_name in ["tensorflow.compat.v1", "paddle"]:
             outputs_pde = outputs
         elif backend_name in ["tensorflow", "pytorch"]:
@@ -181,6 +181,8 @@ class PDE(Data):
                 )
             )
 
+        if X_bc is None:
+            X_bc = self.train_x
         bcs_start = np.cumsum([0] + self.num_bcs)
         bcs_start = list(map(int, bcs_start))
         error_f = [fi[bcs_start[-1] :] for fi in f]
@@ -189,10 +191,14 @@ class PDE(Data):
         ]
         for i, bc in enumerate(self.bcs):
             beg, end = bcs_start[i], bcs_start[i + 1]
-            # The same BC points are used for training and testing.
-            error = bc.error(self.train_x, inputs, outputs, beg, end)
+            error = bc.error(X_bc, inputs, outputs, beg, end)
             losses.append(loss_fn[len(error_f) + i](bkd.zeros_like(error), error))
         return losses
+
+    def losses_test(self, targets, outputs, loss_fn, inputs, model, aux=None):
+        return self.losses(
+            targets, outputs, loss_fn, inputs, model, aux=aux, X_bc=self.test_x
+        )
 
     @run_if_all_none("train_x", "train_y", "train_aux_vars")
     def train_next_batch(self, batch_size=None):
@@ -239,8 +245,11 @@ class PDE(Data):
             self.train_x_all = None
         if bc_points:
             self.train_x_bc = None
+            self.test_x, self.test_y, self.test_aux_vars = None, None, None
         self.train_x, self.train_y, self.train_aux_vars = None, None, None
         self.train_next_batch()
+        if bc_points:
+            self.test()
 
     def add_anchors(self, anchors):
         """Add new points for training PDE losses.
--- a/deepxde/data/pde_operator.py
+++ b/deepxde/data/pde_operator.py
@@ -70,29 +70,44 @@ class PDEOperator(Data):
         self.train_next_batch()
         self.test()
 
-    def losses(self, targets, outputs, loss_fn, inputs, model, aux=None):
+    def losses(self, targets, outputs, loss_fn, inputs, model, aux=None, X_bc=None, aux_var=None):
         f = []
         if self.pde.pde is not None:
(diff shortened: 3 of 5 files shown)


kyouma:
Hello.

Thank you for the PR. Please confirm whether I understand correctly:

1. these changes achieve the same goal as my commit does, but they do it more carefully;
2. they also take into consideration passing correct `aux_var`;
3. the test BC/BC points sampling process remains the same: copy the current training BC/IC points (the "TODO" in `PDE.test_points()`);
4. when the training points are resampled, the test BC/IC points are updated according to this procedure.

echen5503:
Yes, your understanding is correct. The PR addresses the same goal as your commit, but goes through a minimalist approach that maximizes backwards compatibility and minimizes overhead.

kyouma:
Thank you for the explanation.

Issue #2074, opened 2026-03-29
IC/BC caching causes critical errors in inverse problems for IC/BC


kyouma:
Hello.

I have tried to solve an inverse problem for the same equation as in #2073: `2.4` in the IC is changed into `LAMBDA`, and some data points ("measurements") are connected into the BC list. I haven't use the `anchors` argument, because even without anchors the whole `PointSetBC` data point set is connected into the training and testing sets by DeepXDE:
```
class PointSetBC:
    ...
    def collocation_points(self, X):
        if self.batch_size is not None:
            self.batch_indices = self.batch_sampler.get_next(self.batch_size)
            return self.points[self.batch_indices]
        return self.points
```

I have found out the following error while working with the modified DeepXDE ([this commit](https://github.com/kyouma/deepxde/commit/b0ffb52873203945a09760c540ecdce93620c67a) to fix the points mismatch during test loss calculations), but I think that the same applies to the stock DeepXDE (main branch, 30.03.2026).

I found out that the IC loss value on the test sample was very high (for example, O(1) or O(0.1)) even when the training IC loss value was low enough and the metrics with the ground truth solution were also low. I decided to calculate the IC loss myself manually using `data.train_x` and `data.test_x` points, and the manual calculation result was even lower than that of the training IC loss value in the log.

In ICs and BCs the function `func` is cached using `npfunc_range_autocache()`, and when the trainable variables involved in calculation of this `func` changes, the return value doesn't change, and moreover, the variable gradient becomes incorrect. If I disable the cache logic in `wrapper_cache()` and `wrapper_cache_auxiliary()`, my inverse problem converges easily, while the variable flew to `>10` with the caching enabled.

Is this a known behavior? Is there a way to fix this? Maybe by passing a list of trainable variables involved in the function calculation. Or just disabling the caching? I can try to prepare a pull request if I get some guidance about DeepXDE caching and coding style politics.

I also don't know how Tensorflow handles such situations: I see that DeepXDE doesn't force it to cache the function values.

kyouma:
If I disable the caching in a simple non-inverse problem (_2D space X 1D time; 1 equation with 1st and 2nd derivatives, 4+1 conditions; approx. 5000 points per the domain, boundary and initial condition; no `external_trainable_variables`_), for PyTorch on a RTX 2080 Ti the performance decreases by 15%.

[...]

PR #2077, opened 2026-03-30
Make basic BC/IC classes correctly trainable

kyouma:
Hello.

In relation to the issue #2074.

Now it is possible to include trainable variables into BC/IC for inverse problems only by using `OperatorBC` (_at least for the main types of BC/IC_), but it did not manage to find an explicit statement about this in the documentation or FAQ. Now, if one tries to use `DirichletBC`, `NeumannBC`, `IC` etc. that depends on trainable variables, the function values are cached and the trainable variable gradients are incorrect, which leads to terrible convergence behavior.

This PR proposes to require from the basic BC/IC classes to specify whether they depend on trainable variables (this disables function caching for them) or not. It is done independently for each BC/IC to protect the performance.

I have put the new argument at the end of the class signatures in order to preserve backward compatibility for scripts with positional arguments. But the default value is None, which will cause an exception saying that the user need to set it correctly.

The alternative to these or different code changes is to add a warning to the documentation (_at least the BC, IC and inverse problems examples sections_) saying that for inverse problems with trainable variables in BC/IC one must use `OperatorBC`.

### DIFF
--- a/deepxde/icbc/boundary_conditions.py
+++ b/deepxde/icbc/boundary_conditions.py
@@ -48,6 +48,9 @@ class BC(ABC):
             depends_on_trainable_variables = False
         self.depends_on_trainable_variables = depends_on_trainable_variables
 
+        # If learnable geometry is introduced, the boundary normal caching must be disabled
+        # if the "geometry depends on trainable variables" flag of `self.geom` is True,
+        # similar to `self.func` here.
         self.boundary_normal = npfunc_range_autocache(
             utils.return_tensor(self.geom.boundary_normal)
         )
@@ -107,6 +110,8 @@ class RobinBC(BC):
         # with other BC/IC functions with `func`
         # and in case in future caching is added here, too.
         super().__init__(geom, on_boundary, component, depends_on_trainable_variables)
+        # If for some reason caching of `func` is added here in future,
+        # it must be disabled if `self.depends_on_trainable_variables` is True.
         self.func = func
 
     def error(self, X, inputs, outputs, beg, end, aux_var=None):
@@ -169,6 +174,8 @@ class OperatorBC(BC):
         # with other BC/IC functions with `func`
         # and in case in future caching is added here, too.
         super().__init__(geom, on_boundary, 0, depends_on_trainable_variables)
+        # If for some reason caching of `func` is added here in future,
+        # it must be disabled if `self.depends_on_trainable_variables` is True.
         self.func = func
 
     def error(self, X, inputs, outputs, beg, end, aux_var=None):
@@ -283,6 +290,8 @@ class PointSetOperatorBC:
             depends_on_trainable_variables = False
         self.depends_on_trainable_variables = depends_on_trainable_variables
 
+        # If for some reason caching of `func` is added here in future,
+        # it must be disabled if `self.depends_on_trainable_variables` is True
         self.func = func
         self.batch_size = batch_size
 
@@ -354,6 +363,9 @@ class Interface2DBC:
         )
         self.direction = direction
 
+        # If learnable geometry is introduced, the boundary normal caching must be disabled
+        # if the "geometry depends on trainable variables" flag of `self.geom` is True,
+        # similar to `self.func` here.
         self.boundary_normal = npfunc_range_autocache(
             utils.return_tensor(self.geom.boundary_normal)
         )
--- a/deepxde/icbc/boundary_conditions.py
+++ b/deepxde/icbc/boundary_conditions.py
@@ -33,8 +33,7 @@ class BC(ABC):
         geom: A ``deepxde.geometry.Geometry`` instance.
         on_boundary: A function: (x, Geometry.on_boundary(x)) -> True/False.
         component: The output component satisfying this BC.
-        depends_on_trainable_variables: Whether this BC depends on any trainable variable
-            and must have cache disabled, or not.
+        depends_on_trainable_variables: Whether this BC depends on any trainable variable or not.
     """
 
     def __init__(self, geom, on_boundary, component, depends_on_trainable_variables=False):
@@ -151,8 +150,7 @@ class OperatorBC(BC):
             `inputs` and `outputs` are the network input and output tensors,
             respectively; `X` are the NumPy array of the `inputs`.
         on_boundary: (x, Geometry.on_boundary(x)) -> True/False.
-        depends_on_trainable_variables: Whether this BC depends on any trainable variable
-            and must have cache disabled, or not.
+        depends_on_trainable_variables: Whether this BC depends on any trainable variable or not.
 
     Warning:
         If you use `X` in `func`, then do not set ``num_test`` when you define
@@ -264,8 +262,7 @@ class PointSetOperatorBC:
             Note, If you want to use batch size here, you should also set callback
             'dde.callbacks.PDEPointResampler(bc_points=True)' in training.
         shuffle: Randomize the order on each pass through the data when batching.
-        depends_on_trainable_variables: Whether this BC depends on any trainable variable
-            and must have cache disabled, or not.
+        depends_on_trainable_variables: Whether this BC depends on any trainable variable or not.
     """
 
     def __init__(self, points, values, func, batch_size=None, shuffle=True, depends_on_trainable_variables=False):
@@ -330,8 +327,7 @@ class Interface2DBC:
         on_boundary1: First edge func. (x, Geometry.on_boundary(x)) -> True/False.
         on_boundary2: Second edge func. (x, Geometry.on_boundary(x)) -> True/False.
         direction (string): "normal" or "tangent".
-        depends_on_trainable_variables: Whether this BC depends on any trainable variable
-            and must have cache disabled, or not.
+        depends_on_trainable_variables: Whether this BC depends on any trainable variable or not.
     """
 
     def __init__(self, geom, func, on_boundary1, on_boundary2, direction="normal", depends_on_trainable_variables=False):
--- a/deepxde/icbc/boundary_conditions.py
+++ b/deepxde/icbc/boundary_conditions.py
@@ -36,12 +36,16 @@ class BC(ABC):
         depends_on_trainable_variables: Whether this BC depends on any trainable variable or not.
     """
 
-    def __init__(self, geom, on_boundary, component, depends_on_trainable_variables=False):
+    def __init__(self, geom, on_boundary, component, depends_on_trainable_variables=None):
         self.geom = geom
         self.on_boundary = lambda x, on: np.array(
             [on_boundary(x[i], on[i]) for i in range(len(x))]
         )
         self.component = component
+
+        if depends_on_trainable_variables is None:
+            _warn_dependance_on_trainable_variables()
+            depends_on_trainable_variables = False
         self.depends_on_trainable_variables = depends_on_trainable_variables
 
         self.boundary_normal = npfunc_range_autocache(
@@ -69,7 +73,7 @@ class BC(ABC):
 class DirichletBC(BC):
     """Dirichlet boundary conditions: y(x) = func(x)."""
 
-    def __init__(self, geom, func, on_boundary, component=0, depends_on_trainable_variables=False):
+    def __init__(self, geom, func, on_boundary, component=0, depends_on_trainable_variables=None):
         super().__init__(geom, on_boundary, component, depends_on_trainable_variables)
         self.func = npfunc_range_autocache(utils.return_tensor(func), self.depends_on_trainable_variables)
 
@@ -86,7 +90,7 @@ class DirichletBC(BC):
 class NeumannBC(BC):
     """Neumann boundary conditions: dy/dn(x) = func(x)."""
 
-    def __init__(self, geom, func, on_boundary, component=0, depends_on_trainable_variables=False):
+    def __init__(self, geom, func, on_boundary, component=0, depends_on_trainable_variables=None):
         super().__init__(geom, on_boundary, component, depends_on_trainable_variables)
         self.func = npfunc_range_autocache(utils.return_tensor(func), self.depends_on_trainable_variables)
 
@@ -98,7 +102,7 @@ class NeumannBC(BC):
 class RobinBC(BC):
     """Robin boundary conditions: dy/dn(x) = func(x, y)."""
 
-    def __init__(self, geom, func, on_boundary, component=0, depends_on_trainable_variables=False):
+    def __init__(self, geom, func, on_boundary, component=0, depends_on_trainable_variables=None):
         # `depends_on_trainable_variables` is here in order to be consistent
         # with other BC/IC functions with `func`
         # and in case in future caching is added here, too.
@@ -160,7 +164,7 @@ class OperatorBC(BC):
         which cannot be fixed in an easy way for all backends.
     """
 
-    def __init__(self, geom, func, on_boundary, depends_on_trainable_variables=False):
+    def __init__(self, geom, func, on_boundary, depends_on_trainable_variables=None):
         # `depends_on_trainable_variables` is here in order to be consistent
         # with other BC/IC functions with `func`
         # and in case in future caching is added here, too.
@@ -265,7 +269,7 @@ class PointSetOperatorBC:
         depends_on_trainable_variables: Whether this BC depends on any trainable variable or not.
     """
 
-    def __init__(self, points, values, func, batch_size=None, shuffle=True, depends_on_trainable_variables=False):
+    def __init__(self, points, values, func, batch_size=None, shuffle=True, depends_on_trainable_variables=None):
         self.points = np.array(points, dtype=config.real(np))
         if not isinstance(values, numbers.Number) and values.shape[1] != 1:
             raise RuntimeError("PointSetOperatorBC should output 1D values")
@@ -274,6 +278,9 @@ class PointSetOperatorBC:
         # `depends_on_trainable_variables` is here in order to be consistent
         # with other BC/IC functions with `func`
         # and in case in future caching is added here, too.
+        if depends_on_trainable_variables is None:
+            _warn_dependance_on_trainable_variables()
+            depends_on_trainable_variables = False
         self.depends_on_trainable_variables = depends_on_trainable_variables
 
         self.func = func
@@ -330,8 +337,12 @@ class Interface2DBC:
         depends_on_trainable_variables: Whether this BC depends on any trainable variable or not.
     """
 
-    def __init__(self, geom, func, on_boundary1, on_boundary2, direction="normal", depends_on_trainable_variables=False):
+    def __init__(self, geom, func, on_boundary1, on_boundary2, direction="normal", depends_on_trainable_variables=None):
         self.geom = geom
+
+        if depends_on_trainable_variables is None:
+            _warn_dependance_on_trainable_variables()
+            depends_on_trainable_variables = False
         self.depends_on_trainable_variables = depends_on_trainable_variables
 
         self.func = npfunc_range_autocache(utils.return_tensor(func), self.depends_on_trainable_variables)
@@ -441,7 +452,7 @@ def npfunc_range_autocache(func, disable_caching=False):
             cache[key] = func(X[beg:end], aux_var[beg:end])
         return cache[key]
 
-    if backend_name in ["tensorflow.compat.v1", "tensorflow", "jax"] or disable_caching:
+    if (backend_name in ["tensorflow.compat.v1", "tensorflow", "jax"]) or disable_caching:
         if utils.get_num_args(func) == 1:
             return wrapper_nocache
         elif utils.get_num_args(func) == 2:
@@ -451,3 +462,12 @@ def npfunc_range_autocache(func, disable_caching=False):
             return wrapper_cache
         elif utils.get_num_args(func) == 2:
             return wrapper_nocache_auxiliary
+
+
+def _warn_dependance_on_trainable_variables():
+    print(
+        "Warning: The BC/IC instance initialization parameter `depends_on_trainable_variables` "
+        "must be explicitly set to either True or False for all BC and IC objects, or else "
+        "the gradients of the loss function with respect to the external trainable variables "
+        "in inverse problems may become wrong."
+    )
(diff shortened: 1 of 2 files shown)


deepxde/icbc/boundary_conditions.py
```diff
@@ -35,13 +35,19 @@ class BC(ABC):
...
 
-    def __init__(self, geom, on_boundary, component):
+    def __init__(self, geom, on_boundary, component, depends_on_trainable_variables):
         self.geom = geom
         self.on_boundary = lambda x, on: np.array(
             [on_boundary(x[i], on[i]) for i in range(len(x))]
         )
         self.component = component
 
+        if not isinstance(depends_on_trainable_variables, bool):
```
echen5503: I think this approach introduces too much friction for the user, because every code after this change will have to specify external trainable variables, breaking all the current examples. Note that when there are inverse problems, external trainable variables are passed into the model, which has access to the PDE, and hence the BCs. 
```
model.compile(
    "adam", lr=0.001, metrics=["l2 relative error"], external_trainable_variables=C
)
```
 See https://deepxde.readthedocs.io/en/latest/demos/pinn_inverse/diffusion.1d.inverse.html

echen5503 (changes requested):

deepxde/icbc/boundary_conditions.py
kyouma: I may miss something, but I can think of only the following solutions:

1. Use the current changes, but set the default value of this new flag to `False` and remove the boolean check.
    - Old scripts will work as they do now.
    - If someone needs to train a BC/IC, then will have to specify it just as the trainable variables.
2. Pass the list of trainable BC/IC objects to `Model.compile()` together with `external_trainable_variables`.
    - Inside, convert the cache wrapped functions into non-cached functions for these conditions. This will require a mechanism for function wrapper conversion.
    - Or, make a clear cache function to the wrapper and to the `Model` class, and for every training step (_there are multiple training step methods inside the `Model` class_) go through the trainable conditions and selectively clear the cache (_if it's possible at all, because the cache is shared - I may be wrong here_).
    - Clear the whole cache, but it still needs functions for this in the wrapper and the `Model` class, modifying multiple training step methods; and decreases the performance greatly.
3. Pass the list of `True`/`False` flags for BC/IC functions to the `Model.compile()` together with `external_trainable_variables`.
    - Same as the previous one, but uglier.

What do you think?

deepxde/icbc/boundary_conditions.py
```diff
@@ -35,13 +35,19 @@ class BC(ABC):
         component: The output component satisfying this BC.
     """
 
-    def __init__(self, geom, on_boundary, component):
+    def __init__(self, geom, on_boundary, component, depends_on_trainable_variables):
```
echen5503:
```suggestion
    def __init__(self, geom, on_boundary, component, depends_on_trainable_variables=False):
```

deepxde/icbc/boundary_conditions.py
echen5503: I see your point; using my approach the implementation is very difficult, too much considering that this seems to be a temporary fix to the caching. The default value method (#1) seems to be by far the best. The suggested code change is what I assume to be what you mean with #1.

echen5503 (changes requested):

kyouma:
Hello. Please check the changes.

I have updated the default value; removed the boolean check; updated the docstrings to reflect the changes in the documentation.

The next step is to add an example and there emphasize the need to set the `depends_on_trainable_variables` flag. I will work on it.

deepxde/icbc/boundary_conditions.py
```diff
@@ -33,14 +33,16 @@ class BC(ABC):
...
     """
 
-    def __init__(self, geom, on_boundary, component):
+    def __init__(self, geom, on_boundary, component, depends_on_trainable_variables=False):
         self.geom = geom
         self.on_boundary = lambda x, on: np.array(
             [on_boundary(x[i], on[i]) for i in range(len(x))]
         )
         self.component = component
+        self.depends_on_trainable_variables = depends_on_trainable_variables
```
echen5503, line 49: Perhaps it might be a good idea to warn the user of the issue if self.depends_on_trainable is not explicitly given

echen5503 (commented):
It might be a good idea to warn the user if the depends_on_trainable_variables isn't set.

deepxde/icbc/boundary_conditions.py
kyouma: I can see 2 ways to check whether `depends_on_trainable` is set explicitly or not and print the warning:

1. inside the BC/IC classes' `__init__()` - this will bombard with the warnings even the users who do not solve inverse problems, but this does not create any interference between BC/IC classes and, for example, the `Model` class that also can check the BCs properties;
2. inside `Model.compile()` (i.e. examine the elements of `self.data.bcs`) if and only if `external_trainable_variables` is not empty - this will not work for the `tensorflow.compat.v1` backend (this list is ignored by `Model.compile()` and, subsequently, by the users), and this may require a stricter control over future additions of classes connectible to `Model.data` (if they are not children of `Data` class).
    > **Edit**: For `tensorflow.compat.v1` there is no explicit caching in BCs and ICs, so maybe the 2nd option is also viable.

I have made a commit with the 1st option. Following the DeepXDE style, the warning is done with a usual print, so that it will not distract the users unaffected by the bug too much. Besides, I will make an example of an inverse problem with a trainable IC and explicitly state the necessity of setting `depends_on_trainable` to True there. And maybe even add a new entry to the FAQ section.

deepxde/icbc/boundary_conditions.py
```diff
@@ -451,3 +462,12 @@ def wrapper_cache_auxiliary(X, beg, end, aux_var):
             return wrapper_cache
         elif utils.get_num_args(func) == 2:
             return wrapper_nocache_auxiliary
+
+
```
echen5503, line 478: Use @cache here, so it only prints once, and maybe add a docstring that tells future maintainers that it only prints once.

echen5503 (changes requested):

deepxde/icbc/boundary_conditions.py
echen5503: Not sure that @cache is the best solution for this, but we shouldn't spam the user with warnings.

echen5503:
Nice, lru_cache(1) is cleaner than @cache by itself. To make final test, can you try non-converging code on this issue, verifying that it now converges?

kyouma:
I was trying to modify the [diffusion equation inverse problem example](https://github.com/lululxvi/deepxde/blob/master/examples/pinn_inverse/diffusion_1d_inverse.py) and move the unknown constant C from the PDE to the IC:

```
torch.sin(C * x[:, 0:1]) * torch.exp(-x[:, 1:2])
```

and found out that if I put it into `sin()`, then, with caching on, a double `.backward()` (I am using PyTorch) execution error appears:

```
    * * *
  File "C:\Users\Yakov\Documents\GitHub\deepxde\deepxde\model.py", line 394, in closure
    total_loss.backward()

    * * *

  File "C:\Users\Yakov\.pyenv310\lib\site-packages\torch\autograd\graph.py", line 824, in _engine_run_backward
    return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
RuntimeError: Trying to backward through the graph a second time (or directly access saved tensors after they have already been freed). Saved intermediate values of the graph are freed when you call .backward() or autograd.grad(). Specify retain_graph=True if you need to backward through the graph a second time or if you need to access saved tensors after calling backward.
```

And if I place the unknown constant C (_initial guess 1.0)_ into the bias of IC:

```
torch.sin(torch.pi * x[:, 0:1]) * torch.exp(-x[:, 1:2]) + C
```

then it runs even with caching on. And, as expected, does not converge (must become zero):

```
Step      Train loss                                  Test loss                                   Test metric
0 [1.00e+00]
1000      [5.31e-03, 1.87e-02, 4.47e-02, 4.10e-03]    [9.81e-03, 1.87e-02, 4.47e-02, 4.10e-03]    [9.68e-01]
1000 [5.47e-01]
2000      [5.48e-03, 1.63e-02, 1.87e-02, 1.98e-03]    [1.36e-02, 1.63e-02, 1.87e-02, 1.98e-03]    [9.56e-01]
2000 [7.74e-02]
3000      [1.40e-03, 9.87e-03, 7.07e-03, 1.76e-03]    [1.52e-02, 9.87e-03, 7.07e-03, 1.76e-03]    [9.09e-01]
3000 [-3.50e-01]
4000      [1.39e-03, 5.78e-03, 5.10e-03, 2.28e-03]    [2.98e-02, 5.78e-03, 5.10e-03, 2.28e-03]    [8.99e-01]
4000 [-6.98e-01]
5000      [3.71e-03, 3.69e-03, 3.36e-03, 1.79e-03]    [8.48e-02, 3.69e-03, 3.36e-03, 1.79e-03]    [8.82e-01]
5000 [-1.08e+00]
6000      [7.08e-04, 2.72e-03, 1.92e-03, 1.98e-03]    [1.35e-01, 2.72e-03, 1.92e-03, 1.98e-03]    [8.88e-01]
6000 [-1.53e+00]
...
```

When I disable caching for the IC, everything works:

```
Step      Train loss                                  Test loss                                   Test metric
0         [1.53e+01, 5.92e-02, 1.25e+00, 8.94e-02]    [1.64e+01, 5.92e-02, 1.25e+00, 8.94e-02]    [8.63e-01]
0 [1.00e+00]
1000      [4.23e-03, 5.93e-03, 1.73e-02, 8.38e-04]    [6.61e-03, 5.93e-03, 1.73e-02, 8.38e-04]    [5.84e-01]
1000 [6.11e-01]
2000      [7.02e-04, 2.42e-03, 3.20e-03, 2.24e-04]    [1.23e-03, 2.42e-03, 3.20e-03, 2.24e-04]    [3.43e-01]
2000 [3.59e-01]
3000      [4.45e-04, 7.93e-04, 4.27e-04, 3.41e-05]    [6.68e-04, 7.93e-04, 4.27e-04, 3.41e-05]    [1.72e-01]
3000 [1.83e-01]
4000      [1.21e-03, 2.48e-04, 5.16e-05, 1.15e-04]    [1.19e-03, 2.48e-04, 5.16e-05, 1.15e-04]    [8.70e-02]
4000 [9.01e-02]
5000      [7.69e-05, 5.23e-05, 1.97e-05, 1.35e-05]    [2.22e-04, 5.23e-05, 1.97e-05, 1.35e-05]    [4.34e-02]
5000 [4.62e-02]
6000      [6.58e-05, 1.82e-05, 1.14e-05, 6.68e-06]    [1.70e-04, 1.82e-05, 1.14e-05, 6.68e-06]    [2.46e-02]
6000 [2.67e-02]
7000      [5.18e-05, 9.94e-06, 8.40e-06, 4.30e-06]    [1.29e-04, 9.94e-06, 8.40e-06, 4.30e-06]    [1.73e-02]
7000 [1.86e-02]
...
```

And also for C inside `sin()` (the ground truth value is $\pi$):

```
Step      Train loss                                  Test loss                                   Test metric
0         [1.60e+01, 1.78e-02, 2.83e-01, 1.39e-01]    [1.72e+01, 1.78e-02, 2.83e-01, 1.39e-01]    [1.15e+00]
0 [1.00e+00]
1000      [4.98e-03, 6.52e-03, 7.09e-02, 8.99e-04]    [5.79e-03, 6.52e-03, 7.09e-02, 8.99e-04]    [1.19e-01]
1000 [1.49e+00]
2000      [4.84e-03, 3.14e-03, 2.14e-02, 3.07e-05]    [7.04e-03, 3.14e-03, 2.14e-02, 3.07e-05]    [9.88e-02]
2000 [2.28e+00]
3000      [2.21e-04, 5.63e-04, 6.48e-04, 1.02e-05]    [1.98e-03, 5.63e-04, 6.48e-04, 1.02e-05]    [4.36e-02]
3000 [2.90e+00]
4000      [6.63e-05, 1.55e-04, 4.20e-05, 2.23e-05]    [1.11e-03, 1.55e-04, 4.20e-05, 2.23e-05]    [2.07e-02]
4000 [3.04e+00]
5000      [4.39e-05, 8.11e-05, 1.13e-05, 1.48e-05]    [8.35e-04, 8.11e-05, 1.13e-05, 1.48e-05]    [1.51e-02]
5000 [3.07e+00]
6000      [3.38e-05, 5.56e-05, 6.44e-06, 8.92e-06]    [7.05e-04, 5.56e-05, 6.44e-06, 8.92e-06]    [1.28e-02]
6000 [3.08e+00]
7000      [7.29e-05, 4.36e-05, 5.76e-06, 5.87e-06]    [6.75e-04, 4.36e-05, 5.76e-06, 5.87e-06]    [1.16e-02]
7000 [3.09e+00]
...
```

I think that adding `retain_graph=True` to some `.backward()` calls (maybe the `jacobian()` and `hessian()` are the culprits?) is too difficult to control and test across the whole library, and also may need to do the same for other backends. Thus, disabling caching for some BCs and ICs is not only a question of convergence, but also a question of solvability for a lot of inverse problems.

I also have found this discussion #1727, where the author uses the same trick to make IC learnable.

echen5503:
When you say "When I disable caching for the IC," you mean that you passed in the `depends_on_trainable_variable` flag to true right? Could you provide the code for the test?

kyouma:
Yes, I do.

Here it is. I am planning to add this as an example of trainable IC. This is the same problem as in [DeepXDE examples](https://deepxde.readthedocs.io/en/stable/demos/pinn_inverse/diffusion.1d.inverse.html), but I have moved `C` from the PDE to the `sin()` inside the IC (so `C` must converge to $\pi$).

[diffusion_1d_inverse_learnable_ic.py](https://github.com/user-attachments/files/26470659/diffusion_1d_inverse_learnable_ic.py)

deepxde/icbc/boundary_conditions.py
```diff
@@ -421,13 +452,27 @@ def wrapper_cache_auxiliary(X, beg, end, aux_var):
...
             return wrapper_nocache
-        if utils.get_num_args(func) == 2:
+        elif utils.get_num_args(func) == 2:
             return wrapper_nocache_auxiliary
-    if backend_name in ["pytorch", "paddle"]:
+    elif backend_name in ["pytorch", "paddle"]:
         if utils.get_num_args(func) == 1:
             return wrapper_cache
-        if utils.get_num_args(func) == 2:
+        elif utils.get_num_args(func) == 2:
```
echen5503, line 475: What's the point of turning this into elif? Isn't this a common early return design pattern, where you don't need elif?

echen5503 (changes requested):
Looks good, just a few small changes. Does this bug apply to PDE as well?

kyouma:
As far as I understand, it only applies to IC and also BCs that wrap the `func` constructor argument in the caching thing. It causes either wrong gradients or an exception due to multiple `backward()` calls for the loss object stored in the cache.

deepxde/icbc/boundary_conditions.py
kyouma: I thought that because of `...or disable_caching` part there can be a situation when both external conditions are `True`, and if in future `return`s are changed into assignments for some reason, a bug may be overlooked. That is why I decided to mark all these branches as incompatible.

echen5503:
That's my understanding too; PDE classes don't seem to use cache.

deepxde/icbc/boundary_conditions.py
echen5503: The outer elif seems necessary by this reasoning, but what about inner elif?

kyouma: For consistency and explicitness.

echen5503: Ok, I guess it's just a style choice.

deepxde/icbc/boundary_conditions.py
```diff
@@ -33,15 +33,21 @@ class BC(ABC):
...
             [on_boundary(x[i], on[i]) for i in range(len(x))]
         )
         self.component = component
 
+        if depends_on_trainable_variables is None:
+            _warn_dependance_on_trainable_variables()
+            depends_on_trainable_variables = False
+        self.depends_on_trainable_variables = depends_on_trainable_variables
+
         self.boundary_normal = npfunc_range_autocache(
```
echen5503, line 54: Aren't you forgetting to add the flag here?

deepxde/icbc/boundary_conditions.py
```diff
@@ -253,13 +266,23 @@ class PointSetOperatorBC:
...
             raise RuntimeError("PointSetOperatorBC should output 1D values")
         self.values = bkd.as_tensor(values, dtype=config.real(bkd.lib))
+
+        # `depends_on_trainable_variables` is here in order to be consistent
+        # with other BC/IC functions with `func`
+        # and in case in future caching is added here, too.
+        if depends_on_trainable_variables is None:
+            _warn_dependance_on_trainable_variables()
+            depends_on_trainable_variables = False
+        self.depends_on_trainable_variables = depends_on_trainable_variables
```
echen5503, line 291: The rationale behind this is just consistency right? Because this is not used.

echen5503 (commented):
Need a few clarifications

deepxde/icbc/boundary_conditions.py
kyouma: If DeepXDE supports non-constant geometry, then indeed it is safer or even must-do to set the caching flag for boundary normal calculation, too.

Correct me if I am wrong, but as far as I know, DeepXDE does not support such geometry, as even parametrized geometry is reduced to some normalized constant geometry with parameters transferred into the functions.

If there is any concern that in future something may change, it is better to apply the flag here, too. Just like I have changed some lines with `if...elif` statements to make them more bug-proof in case of possible future changes.

deepxde/icbc/boundary_conditions.py
kyouma: The consistency and similar "interfaces" in API docs are beautiful, but I have added it here mainly to make it bug-proof in case of future changes when some type of caching is applied here, too. Apart from that, it is not necessary here at all, it is not even a style preference.

echen5503: Sorry, when I said "style choice" I meant "design/robustness choice"

deepxde/icbc/boundary_conditions.py
echen5503: DeepXDE only supports constant geometry, it seems, and there is no plan to make nonconstant in the near future to my knowledge, current focus is fixing current geometry bugs.

deepxde/icbc/boundary_conditions.py
kyouma: Now there is indeed no need to add these lines here (and in 1 or 2 other places, where BC does not use caching for `func` at all), but if in future someone wanted to change this behavior for some reason, this might be of help. As an alternative, a comment about necessity to add the ability to disable caching if it is applied here in future can be written, but I have decided to just implement a part of this mechanism here.

deepxde/icbc/boundary_conditions.py
kyouma: Then it may be OK to leave the boundary normal calculation as it is now. I can add a comment for each creation of `self.boundary_normal` saying that for support of learnable boundaries the "disable caching" flag must be passed inside it, too.

Or actually, because any changes for learnable geometry support might be done in other places of the library and this comment might be overlooked, I will test how no caching for `self.boundary_normal` affects performance. It it is ~5%, I will add the flag to `self.boundary_normal`, too.

kyouma: Actually, this is a more difficult question. If learnable geometry is introduced it is likely to be done via specifying settings of geometry objects, so one cannot pass the `depends_on_trainable_variables` of a BC object to manage the boundary normal caching. Thus, I have only added comments to warn future contributors.

deepxde/icbc/boundary_conditions.py
echen5503: I think adding a comment for the areas where the flag is added but isn't used is sufficient

deepxde/icbc/boundary_conditions.py
echen5503: Sounds good

deepxde/icbc/boundary_conditions.py
```diff
@@ -33,15 +33,24 @@ class BC(ABC):
...
             [on_boundary(x[i], on[i]) for i in range(len(x))]
         )
         self.component = component
 
+        if depends_on_trainable_variables is None:
+            _warn_dependance_on_trainable_variables()
+            depends_on_trainable_variables = False
+        self.depends_on_trainable_variables = depends_on_trainable_variables
+
+        # If learnable geometry is introduced, the boundary normal caching must be disabled
```
echen5503, line 51: What do you mean by learnable geometry? Finding the shape of the domain for inverse problem?

deepxde/icbc/boundary_conditions.py
```diff
@@ -154,8 +169,13 @@ class OperatorBC(BC):
         which cannot be fixed in an easy way for all backends.
     """
 
-    def __init__(self, geom, func, on_boundary):
-        super().__init__(geom, on_boundary, 0)
+    def __init__(self, geom, func, on_boundary, depends_on_trainable_variables=None):
+        # `depends_on_trainable_variables` is here in order to be consistent
```
echen5503, line 173: I agree with not putting a warning for those that don't use autocache.

kyouma: Should I change the default `depends_on_trainable_variables` argument value to `False` here and in other similar classes?

deepxde/icbc/boundary_conditions.py
kyouma: Yes. Or actually, I should have write "learnable or variable" (for example, a domain that does not depend on any trainable variables, but still changes its shape or size over time). But the introduction of such things into DeepXDE seems to be a very distant prospect.

deepxde/icbc/boundary_conditions.py
echen5503: Sorry for the super late response, I didn't see this in inbox and only saw it now; I don't think it should be set to False herel, it would just be misleading.

echen5503:
Looks good, we should check that the cache is actually useful in most cases though. Normally the `func` that is passed in is super simple. If it leads to marginal performance gains most of the time, then we can hide this from normal user and leave it in for advanced user

PR #2079, opened 2026-03-31
Pytorch backend: add orthogonal initializer

vl-dud:
This PR adds support for the orthogonal kernel initializer for PyTorch backend.


echen5503:
This is the only change that is necessary to add initializers, right? It seems to me that just adding the dictionary will work out-of-the-box.

vl-dud:
Yes, it's enough to simply add an initializer to the dictionary.

PR #2080, opened 2026-03-31
Fix L-BFGS options

vl-dud:
This PR fixes `dde.optimizers.set_LBFGS_options` so that it correctly updates `iter_per_step` and `fun_per_step` used by L-BFGS in PyTorch/Paddle backends.

Before PR:
```
dde.optimizers.set_LBFGS_options(maxiter=123, maxfun=456)
print(dde.optimizers.LBFGS_options)
# {'maxcor': 100, 'ftol': 0, 'gtol': 1e-08, 'maxiter': 123, 'maxfun': 456, 'maxls': 50, 'iter_per_step': 1000, 'fun_per_step': 1250}
```

After PR:
```
dde.optimizers.set_LBFGS_options(maxiter=123, maxfun=456)
print(dde.optimizers.LBFGS_options)
# {'maxcor': 100, 'ftol': 0, 'gtol': 1e-08, 'maxiter': 123, 'maxfun': 456, 'maxls': 50, 'iter_per_step': 123, 'fun_per_step': 456}
```

### DIFF
--- a/deepxde/optimizers/config.py
+++ b/deepxde/optimizers/config.py
@@ -60,6 +60,16 @@ def set_LBFGS_options(
     LBFGS_options["maxfun"] = maxfun if maxfun is not None else int(maxiter * 1.25)
     LBFGS_options["maxls"] = maxls
 
+    # Backend-dependent options
+    if backend_name in ["pytorch", "paddle"]:
+        # number of iterations per optimization call
+        LBFGS_options["iter_per_step"] = min(1000, LBFGS_options["maxiter"])
+        LBFGS_options["fun_per_step"] = (
+            LBFGS_options["maxfun"]
+            * LBFGS_options["iter_per_step"]
+            // LBFGS_options["maxiter"]
+        )
+
 
 def set_NNCG_options(
     lr=1,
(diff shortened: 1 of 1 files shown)


deepxde/optimizers/config.py
```diff
@@ -60,6 +60,16 @@ def set_LBFGS_options(
     LBFGS_options["maxfun"] = maxfun if maxfun is not None else int(maxiter * 1.25)
     LBFGS_options["maxls"] = maxls
 
+    # Backend-dependent options
```
echen5503, line 63: Please verify my understanding: set_hvd_opt_options sets the LFBGS options along with doing the backend-specific operations, but in reality set_LFBGS_options should handle this itself, causing the bug that when you call set_LGBGS_options by itself, it wont update iter_per_step and fun_per_step.

vl-dud: `set_LBFGS_options` is expected to update all L-BFGS-related parameters, including `iter_per_step` and `fun_per_step` used in the Paddle and PyTorch backends. However, the current implementation does not update them.

echen5503: Ok. This looks good.

PR #2090, opened 2026-05-14
fix a bug in L-BFGS config

enigne:
`"iter_per_step"` will be calculated according to user input maxiter, and will be updated accordingly.


vl-dud:
Already addressed in an earlier PR [2080](https://github.com/lululxvi/deepxde/pull/2080)

PR #2092, opened 2026-06-01
Add Union data classes for combining multiple Data objects

vl-dud:
Introduces `UnionRoundRobin` and `UnionMerge`, two strategies for training a single model on multiple data objects simultaneously.

- `UnionRoundRobin` cycles through data objects one at a time, using one data object per training step.
- `UnionMerge` concatenates batches from all data objects into a single forward pass per training step. Note: this strategy is not compatible with `PDE`, `PDEOperator`, or other PDE-based `Data` objects.

Both strategies support per-data loss functions via the `loss` argument when it is defined; otherwise, the loss from `model.compile()` is used.

Example: training an FNN to fit two different functions, *x sin(5x)* and *x cos(2.5x)*, on the same geometry using `UnionMerge`:

```
import deepxde as dde
import numpy as np


def func_1(x):
    return x * np.sin(5 * x)


def func_2(x):
    return x * np.cos(2.5 * x)


num_train = 16
num_test = 100

data_1 = dde.data.Function(geom, func_1, num_train, num_test)
data_2 = dde.data.Function(geom, func_2, num_train, num_test)

data = dde.data.UnionMerge(
    data_objects=[data_1, data_2]
)

activation = "tanh"
initializer = "Glorot uniform"
net = dde.nn.FNN([1] + [20] * 3 + [1], activation, initializer)
model = dde.Model(data, net)
model.compile(
    "adam",
    lr=0.0001,
)
model.train(iterations=10000)

x_plot = np.linspace(-1, 1, 500)[:, None]
y_pred = model.predict(x_plot)
y_true_1 = func_1(x_plot)
y_true_2 = func_2(x_plot)
plt.figure()
plt.plot(x_plot[:, 0], y_true_1[:, 0], "--", label="func_1")
plt.plot(x_plot[:, 0], y_true_2[:, 0], "--", label="func_2")
plt.plot(x_plot[:, 0], y_pred[:, 0], label="prediction", c="black", linewidth=2.0)
plt.xlabel("x")
plt.ylabel("y")
plt.legend()
plt.grid(True)
plt.show()
```

union

### DIFF
--- a/deepxde/data/union.py
+++ b/deepxde/data/union.py
@@ -1,0 +1,185 @@
+import numpy as np
+
+from .data import Data
+from .pde import PDE
+from .pde_operator import PDEOperator, PDEOperatorCartesianProd
+
+
+class Union(Data):
+    """Base class for combining multiple ``Data`` objects.
+
+    Args:
+        data_objects: A list of ``Data`` instances.
+        loss: A shared loss function, or a list of per-data loss functions.
+    """
+
+    def __init__(self, data_objects, loss=None):
+        self.data_objects = list(data_objects)
+        self.loss = loss
+
+        if isinstance(self.loss, (list, tuple)) and len(self.loss) != len(
+            self.data_objects
+        ):
+            raise ValueError("loss list must match number of data objects")
+
+    def _get_loss(self, i, default_loss):
+        if self.loss is None:
+            return default_loss
+        if isinstance(self.loss, (list, tuple)):
+            return self.loss[i]
+        return self.loss
+
+
+class UnionRoundRobin(Union):
+    """Combines multiple ``Data`` objects by cycling through them in round-robin order."""
+
+    def __init__(self, data_objects, loss=None):
+        super().__init__(data_objects, loss)
+        self._train_idx = 0
+        self._test_idx = 0
+
+    def train_next_batch(self, batch_size=None):
+        return self.data_objects[self._train_idx].train_next_batch(batch_size)
+
+    def losses_train(self, targets, outputs, loss_fn, inputs, model, aux=None):
+        i = self._train_idx
+        self._train_idx = (self._train_idx + 1) % len(self.data_objects)
+        return self.data_objects[i].losses_train(
+            targets, outputs, self._get_loss(i, loss_fn), inputs, model, aux=aux
+        )
+
+    def test(self):
+        return self.data_objects[self._test_idx].test()
+
+    def losses_test(self, targets, outputs, loss_fn, inputs, model, aux=None):
+        i = self._test_idx
+        self._test_idx = (self._test_idx + 1) % len(self.data_objects)
+        return self.data_objects[i].losses_test(
+            targets, outputs, self._get_loss(i, loss_fn), inputs, model, aux=aux
+        )
+
+
+class UnionMerge(Union):
+    """Combines multiple ``Data`` objects into one merged forward pass per step.
+
+    Incompatible with PDE-based ``Data`` objects (``PDE``, ``PDEOperator``, etc.);
+    use ``UnionRoundRobin`` for those.
+    """
+
+    def __init__(self, data_objects, loss=None):
+        _incompatible = (PDE, PDEOperator, PDEOperatorCartesianProd)
+        for obj in data_objects:
+            if isinstance(obj, _incompatible):
+                raise TypeError(
+                    f"{type(obj).__name__} is incompatible with UnionMerge. "
+                    "Use UnionRoundRobin instead."
+                )
+
+        super().__init__(data_objects, loss)
+        self._train_slices = None
+        self._test_slices = None
+
+    def train_next_batch(self, batch_size=None):
+        batch_sizes = self._split_batch_size(batch_size)
+        batches = [
+            data.train_next_batch(bs)
+            for data, bs in zip(self.data_objects, batch_sizes)
+        ]
+        *merged, self._train_slices = self._merge_batches(batches)
+        return tuple(merged)
+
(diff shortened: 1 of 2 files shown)


echen5503:
Hmm, this is interesting. Could you provide some more examples to go with this, that have some physical meaning?

vl-dud:
It is more about convenient data grouping than a specific physical meaning.

`Union` can be used in any problem where the data comes from different, possibly conflicting, sources. For example, the code above can be extended with an additional dataset like this:

```
data_3 = dde.data.DataSet(
    fname_train="dataset.train",
    fname_test="dataset.test",
    col_x=(0,),
    col_y=(1,),
    standardize=True,
)
data = dde.data.UnionMerge(
    data_objects=[data_1, data_2, data_3]
)
```

The data sources can also be prioritized using `loss_weights`:
```
model.compile(
    "adam",
    lr=0.0001,
    loss_weights=[1, 1e-1, 1e1]
)
```

echen5503:
Why would you want to combine conflicting datasets? Why should we encourage bad design choices when the user should be applying normalizations and other preprocessing steps so that all data is standardized and these merge classes are not needed?

vl-dud:
This is a typical situation in multi-source learning: we may want to train on both an approximate analytical model and noisy experimental data.
In the example above, how else should we combine `dde.data.Function` and `dde.data.DataSet` in a clean way?
These sources are not always perfectly consistent, but combining them with appropriate loss weights can be meaningful. So I would not consider this inherently a bad design choice.

echen5503:
I see, there is no way to combine Function (dynamic) with Dataset (static). I will take a closer look at this PR now.

deepxde/data/union.py
```diff
@@ -0,0 +1,185 @@
...
+        return self.data_objects[self._train_idx].train_next_batch(batch_size)
+
+    def losses_train(self, targets, outputs, loss_fn, inputs, model, aux=None):
+        i = self._train_idx
+        self._train_idx = (self._train_idx + 1) % len(self.data_objects)
+        return self.data_objects[i].losses_train(
+            targets, outputs, self._get_loss(i, loss_fn), inputs, model, aux=aux
+        )
+
+    def test(self):
```
echen5503, line 51: is it intentional to only offer the user one of the datasets at once, based on the internal train_idx counter?

deepxde/data/union.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def _get_loss(self, i, default_loss):
+        if self.loss is None:
+            return default_loss
+        if isinstance(self.loss, (list, tuple)):
+            return self.loss[i]
+        return self.loss
+
+
+class UnionRoundRobin(Union):
```
echen5503, line 33: Does this work for static computational graphs? Because it would change shape over and over again.

deepxde/data/union.py
```diff
@@ -0,0 +1,185 @@
+import numpy as np
+
+from .data import Data
+from .pde import PDE
+from .pde_operator import PDEOperator, PDEOperatorCartesianProd
+
+
+class Union(Data):
```
echen5503, line 8: This is an abstract base class right? It should not be directly instantiated?

deepxde/data/union.py
```diff
@@ -0,0 +1,185 @@
...
+        i = self._test_idx
+        self._test_idx = (self._test_idx + 1) % len(self.data_objects)
+        return self.data_objects[i].losses_test(
+            targets, outputs, self._get_loss(i, loss_fn), inputs, model, aux=aux
+        )
+
+
+class UnionMerge(Union):
+    """Combines multiple ``Data`` objects into one merged forward pass per step.
+
```
echen5503, line 64: We should notify the user that all data objects need to be the same shape.

echen5503: (In both inputs and outputs)

deepxde/data/union.py
```diff
@@ -0,0 +1,185 @@
...
+        super().__init__(data_objects, loss)
+        self._train_idx = 0
+        self._test_idx = 0
+
+    def train_next_batch(self, batch_size=None):
+        return self.data_objects[self._train_idx].train_next_batch(batch_size)
+
+    def losses_train(self, targets, outputs, loss_fn, inputs, model, aux=None):
+        i = self._train_idx
+        self._train_idx = (self._train_idx + 1) % len(self.data_objects)
```
echen5503, line 46: doesn't LFBGS call losses_train multiple times per train step? Then we would be incrementing multiple times per train step, which could cause dataset 1 loss to taken into account for dataset 0, which will lead to adverse effects. We should increment in `train_next_batch` instead.

deepxde/data/union.py
```diff
@@ -0,0 +1,185 @@
...
+            aux,
+            self._test_slices,
+            train=False,
+        )
+
+    def _losses(self, targets, outputs, loss_fn, inputs, model, aux, slices, train):
+        losses = []
+        for i, (data, sl) in enumerate(zip(self.data_objects, slices)):
+            x_i = self._slice_inputs(inputs, sl)
+            y_i = None if targets is None else targets[sl]
```
echen5503, line 130: The code seems to correctly handle if the input is a tuple, but what about targets and outputs?

echen5503 (changes requested):
In general, can more documentation be added here? The inner workings of some functions are quite opaque.

vl-dud:
@echen5503, thanks for the review. I’ve taken all your feedback into account. Finally, I decided to revisit the implementation completely and drop `UnionRobin` for now. I kept only pytorch among the supported backends and also limited the number of supported `Data` classes.

PR #2101, opened 2026-08-06
Validate values array shape in PointSetBC to prevent silent broadcasting

### LINKED ISSUE #2100: PointSetBC accepts a 1-D `values` array and silently broadcasts the error to (N, N)
I hit this problem in an inverse problem where I fit some 1D-data at some observation points with `PointSetBC`. Training ran without any error or warning, but the data loss does not drop. It took me a while to realize why; I had read the observed data with `np.loadtxt`, which returns shape `(N,)`, and passed that array straight to `PointSetBC`.

Minimal version:

```python
import os
os.environ["DDE_BACKEND"] = "pytorch"
import numpy as np
import deepxde as dde
import torch

dde.config.set_random_seed(0)

N = 20
points = np.linspace(0, 1, N).reshape(-1, 1)
values = np.sin(2 * np.pi * points).ravel()   # (N,), not (N, 1)

bc = dde.icbc.PointSetBC(points, values, component=0)
err = bc.error(None, None, torch.zeros(N, 1), 0, N)
print("error shape:", tuple(err.shape))

data = dde.data.PDE(dde.geometry.Interval(0, 1), lambda x, y: 0 * y,
                    [bc], num_domain=2)
model = dde.Model(data, dde.nn.FNN([1] + [32] * 3 + [1], "tanh", "Glorot uniform"))
model.compile("adam", lr=1e-3)
model.train(iterations=5000, display_every=1000)
print("Var(values) =", values.var())
```

Output:

```
error shape: (20, 20)
Step      Train loss              Test loss
0         [0.00e+00, 5.03e-01]    [0.00e+00, 5.03e-01]
(omitted)
5000      [0.00e+00, 4.75e-01]    [0.00e+00, 4.75e-01]
Var(values) = 0.475
```

If I use `values.reshape(-1, 1)` instead, the same script converges to ~1e-7 as expected.

I think the problem comes from `PointSetBC.error`.

```python
return outputs[beg:end, self.component : self.component + 1] - self.values
```

`outputs[beg:end, c:c+1]` is `(N, 1)` and `self.values` is `(N,)`, so the subtraction broadcasts to `(N, N)`. The loss then becomes the mean over all pairs,

```
mean_ij (u_i - v_j)^2 = Var(v) + mean_i (u_i - mean(v))^2
```

which is minimized when the network output is spatially constant, with floor `Var(values)`. That matches the 0.475 plateau above, and in my real problem the plateau matched `0.5(Var(u_obs) + Var(v_obs))` to four digits, which is how I eventually tracked it down.

The docstring says `values` should be "A scalar or a 2D-array of values", and `DirichletBC` and `PointSetOperatorBC` both do check the shape of what they get, but `PointSetBC` do not.

I think it is helpful if `PointSetBC` does that kind of check too.
(However, I did not read the entire DeepXDE codebase. I would appreciate it if anyone could suggest other good solutions.)

Versions: DeepXDE 1.12.1 (the relevant code looks the same on master), PyTorch 2.4.1+cu121,
Python 3.10.12, NumPy 1.26.4.

AyushG-1210:
Fixes #2100

### Problem
Passing a 1D array of shape `(N,)` (e.g., loaded via `np.loadtxt`) to `PointSetBC` causes implicit array broadcasting during subtraction against network predictions of shape `(N, 1)`. This results in an `(N, N)` error matrix during training, causing the loss to plateau around `Var(values)` without raising any warnings or errors.

### Solution
Added input shape and dimension validation inside `PointSetBC.__init__`:
- Enforces that non-scalar `values` are 2D arrays of shape `(N, expected_cols)`.
- Validates point count alignment (`len(points) == values_arr.shape[0]`).
- Validates column/component alignment (`values_arr.shape[1] == expected_cols`), catching both multi-component list mismatches and single-component 2D column mismatches (e.g., passing `(N, 2)` when `component=0`).
- Enforces the shape requirement already documented in the class docstring ("a scalar or a 2D-array"); existing valid code is unaffected.
- Raises a descriptive `RuntimeError` advising users to use `.reshape(-1, 1)` when 1D data is provided.

### Validation & Compatibility
- Validation logic is pure NumPy/Python inside `__init__` before backend tensor conversion (`bkd.as_tensor`), making the check backend-agnostic.
- Verified locally under PyTorch and TensorFlow (confirmed empirically on both that validation triggers correctly before backend specifics apply):
  - 1D input `(N,)` raises `RuntimeError` on initialization.
  - Column count mismatch `(N, 2)` with a single integer component raises `RuntimeError`.
  - Point count mismatch `(10 points, 8 values)` raises `RuntimeError`.
  - Valid 2D input `(N, 1)` and scalar inputs pass cleanly.

### DIFF
--- a/deepxde/icbc/boundary_conditions.py
+++ b/deepxde/icbc/boundary_conditions.py
@@ -184,6 +184,32 @@ class PointSetBC:
 
     def __init__(self, points, values, component=0, batch_size=None, shuffle=True):
         self.points = np.array(points, dtype=config.real(np))
+
+        if not isinstance(values, numbers.Number):
+            values_arr = np.asarray(values)
+            if values_arr.ndim != 2:
+                raise RuntimeError(
+                    "PointSetBC should receive values of shape (N, 1) for a "
+                    "single component, or (N, len(component)) if component "
+                    "is a list, but got an array of shape {}. A common cause "
+                    "is loading 1D data (e.g. via np.loadtxt), which needs to "
+                    "be reshaped first, e.g. values.reshape(-1, 1).".format(
+                        values_arr.shape
+                    )
+                )
+            if values_arr.shape[0] != len(self.points):
+                raise RuntimeError(
+                    "PointSetBC received {} points but {} values.".format(
+                        len(self.points), values_arr.shape[0]
+                    )
+                )
+            expected_cols = len(component) if isinstance(component, list) else 1
+            if values_arr.shape[1] != expected_cols:
+                raise RuntimeError(
+                    "PointSetBC received {} components but values of shape[1] "
+                    "= {}.".format(expected_cols, values_arr.shape[1])
+                )
+
         self.values = bkd.as_tensor(values, dtype=config.real(bkd.lib))
         self.component = component
         if isinstance(component, list) and backend_name != "pytorch":
--- a/deepxde/icbc/boundary_conditions.py
+++ b/deepxde/icbc/boundary_conditions.py
@@ -187,15 +187,12 @@ class PointSetBC:
 
         if not isinstance(values, numbers.Number):
             values_arr = np.asarray(values)
+            expected_cols = len(component) if isinstance(component, list) else 1
+
             if values_arr.ndim != 2:
                 raise RuntimeError(
-                    "PointSetBC should receive values of shape (N, 1) for a "
-                    "single component, or (N, len(component)) if component "
-                    "is a list, but got an array of shape {}. A common cause "
-                    "is loading 1D data (e.g. via np.loadtxt), which needs to "
-                    "be reshaped first, e.g. values.reshape(-1, 1).".format(
-                        values_arr.shape
-                    )
+                    f"PointSetBC received values of shape {values_arr.shape}, "
+                    f"expected 2D array of shape (N, {expected_cols})."
                 )
             if values_arr.shape[0] != len(self.points):
                 raise RuntimeError(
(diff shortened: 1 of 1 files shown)


deepxde/icbc/boundary_conditions.py
```diff
@@ -184,6 +184,32 @@ class PointSetBC:
...
         self.points = np.array(points, dtype=config.real(np))
+
+        if not isinstance(values, numbers.Number):
+            values_arr = np.asarray(values)
+            if values_arr.ndim != 2:
+                raise RuntimeError(
+                    "PointSetBC should receive values of shape (N, 1) for a "
+                    "single component, or (N, len(component)) if component "
+                    "is a list, but got an array of shape {}. A common cause "
+                    "is loading 1D data (e.g. via np.loadtxt), which needs to "
```
echen5503: There is no need to make such an opinionated comment that might confuse readers. Who knows if 1D data is a common cause or just your cause?

echen5503 (changes requested):
Change itself is reasonable, but please keep the shape-checking short like the others.

deepxde/icbc/boundary_conditions.py
AyushG-1210: Thank you for the review! Updated and simplified the error message.

echen5503 (commented):
Looks better now. In future PRs, we should look to standardize this warning for other classes, which have a similar function signature but no corresponding warning.

AyushG-1210:
I'd be happy to take a look at other classes that produce no warnings, and I'll open a PR to standardize the shape validation.

PR #2102, opened 2026-08-07
Standardize shape validation for PointSetOperatorBC

AyushG-1210:
Follow-up to PR #2101. 

### Problem
`PointSetOperatorBC.__init__` checks `values.shape[1] != 1` directly, without 
first confirming `values` is 2D. A 1D array (e.g. `np.ones(10)`, shape `(10,)`) 
raises an unrelated `IndexError: tuple index out of range` before the intended 
validation ever runs. There was also no check that `len(points) == len(values)`.

### Solution
Applied the same validation pattern from #2101 to `PointSetOperatorBC.__init__`:
- Converts `values` to a NumPy array via `np.asarray` before touching `.shape`.
- Validates non-scalar `values` are 2D with `shape == (N, 1)`.
- Validates point count alignment (`len(points) == values_arr.shape[0]`).
- Raises a descriptive `RuntimeError` in place of the previous `IndexError`.

### Testing
Verified locally across both PyTorch and TensorFlow backends:
- 1D input `(10,)` → `RuntimeError` with shape message.
- Column mismatch `(10, 2)` → `RuntimeError` with shape message.
- Point-count mismatch (10 points, 8 values) → `RuntimeError`.
- Valid input `(10, 1)` → passes cleanly, no error.

### Out of scope
`Interface2DBC.error()` has a related but structurally different gap, its 
shape check (`bkd.ndim(values) == 2 and bkd.shape(values)[1] != 1`) silently 
skips validation when `func` returns a 1D array, which can still hit the same 
class of silent-broadcast bug downstream. Not fixing here since it's a 
different constructor pattern (no static `values` array, goes through `bkd` 
tensor ops rather than raw NumPy). Flagging in case it's worth a separate PR.

### DIFF
--- a/deepxde/icbc/boundary_conditions.py
+++ b/deepxde/icbc/boundary_conditions.py
@@ -257,8 +257,17 @@ class PointSetOperatorBC:
 
     def __init__(self, points, values, func, batch_size=None, shuffle=True):
         self.points = np.array(points, dtype=config.real(np))
-        if not isinstance(values, numbers.Number) and values.shape[1] != 1:
-            raise RuntimeError("PointSetOperatorBC should output 1D values")
+        if not isinstance(values, numbers.Number):
+            values_arr = np.asarray(values)
+            if values_arr.ndim != 2 or values_arr.shape[1] != 1:
+                raise RuntimeError(
+                    f"PointSetOperatorBC received values of shape {values_arr.shape}, "
+                    f"expected 2D array of shape (N, 1)."
+                )
+            if values_arr.shape[0] != len(self.points):
+                raise RuntimeError(
+                    f"PointSetOperatorBC received {len(self.points)} points but {values_arr.shape[0]} values."
+                )
         self.values = bkd.as_tensor(values, dtype=config.real(bkd.lib))
         self.func = func
         self.batch_size = batch_size


echen5503:
sorry for confusion, when I said standardize I meant making some general functions and applying them appropriately to classes which they are applicable. Something like the following:

```
def _check_target_values(values, n_points, n_components, cls_name):
    """Validate fixed target values at construction time.
 
    Scalars are allowed and broadcast intentionally. Arrays must match
    ``(n_points, n_components)`` exactly, since a ``(N,)`` vs ``(N, 1)``
    mismatch would otherwise broadcast to ``(N, N)`` in the loss.
    """
    if isinstance(values, numbers.Number):
        return
    shape = np.asarray(values).shape
    if shape != (n_points, n_components):
        raise ValueError(
            f"{cls_name}: `values` must have shape "
            f"{(n_points, n_components)} (n_points, n_components); got {shape}."
        )
 
 
def _check_func_output(values, cls_name):
    """Validate the output of a user-supplied BC function at run time.
 
    Allows 0-d scalars (e.g. ``lambda x: 0``), which broadcast harmlessly
    against an ``(N, 1)`` slice. Rejects 1-D and wide 2-D outputs, which
    broadcast silently to ``(N, N)``. Row count is already guaranteed by
    the ``[beg:end]`` slicing and is not re-checked.
    """
    nd = bkd.ndim(values)
    if nd == 0:
        return
    if nd != 2 or bkd.shape(values)[1] != 1:
        raise RuntimeError(
            f"{cls_name}: func should return an array of shape N by 1 for each "
            "component. Use the argument 'component' for different output "
            "components."
        )
 
```

AyushG-1210:
That makes sense. We could either refactor this PR to integrate these functions, or merge this fix and address the general functions in a future PR, your call.

echen5503:
Let's refactor this PR.

AyushG-1210:
I've refactored the shape validation into centralized helper functions as suggested and applied them across the board. I also added a dedicated test suite covering 1D-rejection, column mismatch, and point-count alignment for all of these classes across backends. We can look into other shape discrepancies in future PRs, such as in modules like `geometry`.

deepxde/icbc/boundary_conditions.py
```diff
@@ -126,13 +169,12 @@ def collocation_points(self, X):
...
         mid = beg + (end - beg) // 2
         if self.derivative_order == 0:
-            yleft = outputs[beg:mid, self.component : self.component + 1]
-            yright = outputs[mid:end, self.component : self.component + 1]
-        else:
-            dydx = grad.jacobian(outputs, inputs, i=self.component, j=self.component_x)
-            yleft = dydx[beg:mid]
-            yright = dydx[mid:end]
-        return yleft - yright
+            return (
```
echen5503: Why did you change this area? Is it relevant to the PR?

echen5503 (commented):
To me, this seems like it makes validation more elegant and comprehensive. I'd like to maybe hear more opinions on if there are any other good approaches to doing this.

Regarding quality, please give the output of the tests in a controlled environment, and address comments below.

deepxde/icbc/boundary_conditions.py
AyushG-1210: Not relevant to the PR, reverting these changes.

AyushG-1210:
Since your last review: fixed RobinBC.error, which had landed as a broken partial edit. The `values` was referenced but never assigned, so every call raised NameError rather than the intended shape check. It's now three lines: evaluate `func`, validate with `_check_func_output`, then compute the return. Also reverted the unrelated PeriodicBC `yleft`/`yright` change you flagged. Separately, tightened the test suite, every `pytest.raises` now matches on the error message text rather than just the exception type, which is what caught the RobinBC bug in the first place.

AyushG-1210:
Tests:
<details>
<summary>Backend test runner (run_backend_check.ps1)</summary>

```powershell
# Sweeps the shape-validation suite across every installed DeepXDE backend.
#
# Lives in and runs from deepxde/icbc/. Resolves the repo root itself so the
# regression run still covers the project's real test suite.
#
# Usage (venv active):
#   cd deepxde\icbc
#   .\run_backend_check.ps1
#
# Output: backend_matrix_results.txt

$ErrorActionPreference = "Continue"

$ScriptDir = Split-Path -Parent $MyInvocation.MyCommand.Definition
$OutFile   = Join-Path $ScriptDir "backend_matrix_results.txt"
$NewSuite  = Join-Path $ScriptDir "test_icbc_shape_validation.py"

if (Test-Path $OutFile) { Remove-Item $OutFile }

function Log($msg) {
    Write-Host $msg
    Add-Content -Path $OutFile -Value $msg
}

# Resolve repo root so the regression suite can be found from anywhere.
$RepoRoot = (git rev-parse --show-toplevel 2>$null)
if ($LASTEXITCODE -eq 0 -and $RepoRoot) {
    $RepoRoot = $RepoRoot -replace "/", "\"
} else {
    # Fallback: deepxde/icbc -> deepxde -> repo root
    $RepoRoot = (Resolve-Path (Join-Path $ScriptDir "..\..")).Path
}
$RegressionSuite = Join-Path $RepoRoot "tests"

if (-not (Test-Path $NewSuite)) {
    Write-Host "ERROR: cannot find $NewSuite"
    exit 1
}

# ---------------------------------------------------------------------------
# Environment provenance - reviewers need this to trust the run
# ---------------------------------------------------------------------------
Log "=========================================================="
Log " ENVIRONMENT"
Log "=========================================================="
Log "date        : $(Get-Date -Format 'yyyy-MM-dd HH:mm:ss zzz')"
Log "os          : $([System.Environment]::OSVersion.VersionString)"
Log "python      : $(python --version 2>&1)"
Log "pytest      : $(pytest --version 2>&1 | Select-Object -First 1)"
Log "repo root   : $RepoRoot"
Log "git commit  : $(git rev-parse --short HEAD 2>&1)"
Log "git branch  : $(git rev-parse --abbrev-ref HEAD 2>&1)"
Log ""
Log "--- installed package versions ---"
$pkgs = @("numpy", "torch", "tensorflow", "paddlepaddle", "jax", "jaxlib")
foreach ($p in $pkgs) {
    $v = (pip show $p 2>$null | Select-String "^Version:") -replace "Version:\s*", ""
    if ($v) { Log ("{0,-14}: {1}" -f $p, $v.Trim()) }
    else    { Log ("{0,-14}: (not installed)" -f $p) }
}
Log ""

# ---------------------------------------------------------------------------
# Backend matrix
# ---------------------------------------------------------------------------
$backends = @(
    "tensorflow.compat.v1",
    "tensorflow",
    "pytorch",
    "paddle",
    "jax"
)

$summary = @()

foreach ($b in $backends) {
    Log ""
    Log "=========================================================="
    Log " BACKEND: $b"
    Log "=========================================================="

    $env:DDE_BACKEND = $b

    # Confirm the backend actually loaded before trusting any test result.
    # Run from the repo root so `import deepxde` resolves the working tree,
    # not a stale site-packages copy.
    Push-Location $RepoRoot
    $probe = python -c "import deepxde; from deepxde.backend import backend_name; print(backend_name)" 2>&1
    $probeOk = ($LASTEXITCODE -eq 0)
    Pop-Location

    Log "backend probe: $probe"

    if (-not $probeOk) {
        Log "SKIPPED - backend failed to import"
        $summary += [PSCustomObject]@{
            Backend = $b; NewSuite = "import failed"; Regression = "-"
        }
        continue
    }

    # New suite (path is absolute, so cwd does not matter)
    Log ""
    Log "--- new suite: deepxde/icbc/test_icbc_shape_validation.py ---"
    Push-Location $RepoRoot
    $new = pytest $NewSuite -v --tb=short 2>&1
    Pop-Location
    $new | ForEach-Object { Log $_ }
    $newTail = ($new | Select-String -Pattern "passed|failed|error" | Select-Object -Last 1)

    # Regression: checked for a pre-existing suite via `git ls-files` and
    # the CI workflows in .github/workflows (build.yml, release.yml) -
    # this repo has no committed pytest suite to regress against. Stating
    # that explicitly rather than running pytest against an empty/missing
    # path, which would otherwise silently print "no tests ran" and look
    # like a check that passed rather than one that didn't exist.
    Log ""
    Log "--- regression ---"
    Log "No existing test suite found in this repo (checked git ls-files"
    Log "and .github/workflows/{build,release}.yml - neither runs pytest)."
    Log "deepxde/icbc/test_icbc_shape_validation.py is the only automated"
    Log "coverage for this module."
    $regTail = "no suite in repo"

    $summary += [PSCustomObject]@{
        Backend    = $b
        NewSuite   = "$newTail".Trim()
        Regression = "$regTail".Trim()
    }
}

Remove-Item Env:\DDE_BACKEND -ErrorAction SilentlyContinue

# ---------------------------------------------------------------------------
# Matrix summary - this is the block to paste at the top of the PR comment
# ---------------------------------------------------------------------------
Log ""
Log "=========================================================="
Log " SUMMARY MATRIX"
Log "=========================================================="
foreach ($row in $summary) {
    Log ("{0,-22} | new: {1,-30} | regression: {2}" -f $row.Backend, $row.NewSuite, $row.Regression)
}
Log ""
Log "Full transcript written to $OutFile"
```
</details>

<details>
<summary>Output Summary (all 5 backends)</summary>

```text
==========================================================
 SUMMARY MATRIX
==========================================================
tensorflow.compat.v1   | new: ======================= 37 passed, 2 skipped in 21.00s ======================== | regression: no suite in repo
tensorflow             | new: ================= 37 passed, 2 skipped, 2 warnings in 22.80s ================== | regression: no suite in repo
pytorch                | new: ============================= 39 passed in 6.42s ============================== | regression: no suite in repo
paddle                 | new: ================== 37 passed, 2 skipped, 1 warning in 5.70s =================== | regression: no suite in repo
jax                    | new: ======================== 37 passed, 2 skipped in 6.79s ======================== | regression: no suite in repo
```
</details>

